PR #4, opened 2017-10-16
CollapsibleSection component

bernardodiasc:
**_Work in progress, please do not merge_**

- https://github.com/x-team/auto/issues/513
- https://github.com/x-team/auto/issues/514


bernardodiasc:
@x-team/auto I'm getting familiar with `cmz` and `elem` usages and syntax patterns. I got something usable so far but I think the css classes logic bytes can be improved. Please let me know any suggestions. /cc @joshwnj 

How can I receive and use `isTwoColumns` inside `cmz`? Something like this:

```js
const Header = elem.h1(cmz(`
  // some style...
  {isTwoColumns && `
    // different styles for this flag...
  `}
`))

// while isTwoColumns is a prop
```
---

The appearance of the WIP looks like this now:

screen shot 2017-10-16 at 6 15 04 pm

will send more commits next hours that will make it look better and include the collapsible effect.

joshwnj:
@bernardodiasc looking good! The thing to know about `cmz` is that even though it might look magical, it tries to be restrained in how much magic is introduced.

So when you call the `cmz()` function, 2 things happen:

- it returns a unique className
- it inserts a `<style>` tag into the `<head>`.

This means that if we want conditional styles, the 2 natural options are:

1. create all needed classes at module-time, and compose at render-time.
2. use css variables

If you're interested what option 2 looks like, here's an example of how you might do it: https://gist.github.com/joshwnj/2ef1f7856fae06e544593a8b6ec9918b

We actually could use variables here, which would be pretty cool (we don't need to support IE).  But I still think it's worth trying to solve the problem neatly with composition first, and only go to vars if it becomes cleaner.

Here's how I would do it.  First, define the base css classes:

```js
const cx = {
  root: cmz(`
  margin: 0
  padding: 1rem
`),

  header: cmz(`
  margin: 0
  padding: 0.5rem
`),

  content: cmz(`
  border: 1px solid red
`)
}
```

Then define a toggle class that we can compose to do a bit of controlled cascading:

```js
const twoColSection = cmz(`
& {
  display: flex;
}

& > .${cx.header} {
  border: 1px solid blue;
}

& > .${cx.content} {
  border: 1px solid orange;
}
`)
```

Then define the elements:

```js
const Root = elem.section(cx.root)
const Header = elem.h1(cx.header)
const Content = elem.div(cx.content)
```

All of this happens at module-time, which means it is only executed once when the app starts.

Finally we just say how to render this:

```js
  render () {
    const {
      title,
      isTwoColumns,
      children,
    } = this.props

    return Root({ className: isTwoColumns && twoColSection },
      Header(title),
      Content(children)
    )
  }
```

Look ok?

bernardodiasc:
Awesome! All questions (and a bit more) cleared now, thanks @joshwnj

bernardodiasc:
# CMZ troubleshooting

The snippets bellow won't work, was following https://www.npmjs.com/package/cmz instructions :/

```
cmz(`
  font-weight: 800
  font-size: 1.2em
  margin: 0
  padding: 0
`).compose([
  typo.family.base,
  cmz.pseudo('after', cmz(`
    content: '▾'
    position: absolute
    right: 0.5em
    top: 0.5em
  `))
])

// Uncaught TypeError: Cannot read property 'forEach' of undefined
//     at renderCss (index.js:26)
//     at Function.cmz.pseudo (index.js:64)
//     at Object.sheets (CollapsibleSection.js:35)
//     at __webpack_require__ (bootstrap fc410375e28dc3511c96:705)
//     at fn (bootstrap fc410375e28dc3511c96:111)
//     at Object.module.exports.aliceblue (index.js?5ba1:26)
//     at __webpack_require__ (bootstrap fc410375e28dc3511c96:705)
//     at fn (bootstrap fc410375e28dc3511c96:111)
//     at renderStyleguide (index.js:17)
//     at Object.<anonymous> (index.js:90)


cmz(`
  font-weight: 800
  font-size: 1.2em
  margin: 0
  padding: 0
`).compose(typo.family.base)

// and

cmz(`
  font-weight: 800
  font-size: 1.2em
  margin: 0
  padding: 0
`).compose([
  typo.family.base,
])

// Uncaught TypeError: (0 , _cmz2.default)(...).compose is not a function
//     at Object.sheets (CollapsibleSection.js:33)
//     at __webpack_require__ (bootstrap 01c9e4bb4403bc8cf217:705)
//     at fn (bootstrap 01c9e4bb4403bc8cf217:111)
//     at Object.module.exports.aliceblue (index.js?5ba1:26)
//     at __webpack_require__ (bootstrap 01c9e4bb4403bc8cf217:705)
//     at fn (bootstrap 01c9e4bb4403bc8cf217:111)
//     at renderStyleguide (index.js:17)
//     at Object.<anonymous> (index.js:90)
//     at __webpack_require__ (bootstrap 01c9e4bb4403bc8cf217:705)
//     at fn (bootstrap 01c9e4bb4403bc8cf217:111)
```

Using a different approach, that is the same on existing components

```
cmz([ typo.family.base, `
  font-weight: 800
  font-size: 1.2em
  margin: 0
  padding: 0
`])

// ^ this works, but there is no pseudo element yet


cmz([ typo.family.base, `
  & {
    font-weight: 800
    font-size: 1.2em
    margin: 0
    padding: 0
  }

  &::after {
    content: '▾'
    position: absolute
    right: 0.5em
    top: 0.5em
  }
`])

// ^ this won't work, only the first item in array was included (typo.family.base)
// BUT using with `hover` pseudoelement will work - the hover effect only - and will not include the styles for the class
// tricky!


cmz([ typo.family.base, `
  font-weight: 800
  font-size: 1.2em
  margin: 0
  padding: 0
`, `
  &::after {
    content: '▾'
    position: absolute
    right: 0.5em
    top: 0.5em
  }
`])

// ^ this won't work, only the 2 first item in array was included
// The `after` wasn't included :/

```

> Conclusion, I couldn't use `::after` anyway so far... The `:hover` works mostly of cases above, but not all perfect.

bernardodiasc:
Made few different prototypes of the behaviour to cover described specs, two of them was made on codepen instead styleguidist just to get the appearance and behaviour working:

- https://codepen.io/bernardodiasc/pen/RLvPqp This one uses `<details>` tag, it's good and don't require any JS, but some very specific behaviours requires different markup that wouldn't work this way. (won't work)
- https://codepen.io/bernardodiasc/pen/EwrVgP This works, but uses `::after` and I still stuck on `cmz` part.

joshwnj:
@bernardodiasc try this:

```js

cmz([ typo.family.base, `
  & {
    font-weight: 800;
    font-size: 1.2em;
    margin: 0;
    padding: 0;
  }

  &::after {
    content: '▾';
    position: absolute;
    right: 0.5em;
    top: 0.5em;
  }
`])
```

Things that aren't very well described in the current `cmz` README:

- to compose classes, pass an array into `cmz`. Eg. `cmz([ ... ])`
- when writing pseudo-selectors, we need semicolons (I ought to update `cmz` to either add them for you, or at least give a warning)

bernardodiasc:
I think the warning could be good to have, but for this specific case I see no problem in using semicolons. :)

bernardodiasc:
Just a note about `cmz`, this technique:

```
const header = cmz([ typo.family.base, `
  & {
    font-weight: 800;
    font-size: 1.2em;
    margin: 0;
    padding: 0;
  }

  &::after {
    content: '▾';
    position: absolute;
    right: 0.5em;
    top: 0.5em;
  }
`])
```
And

```
const twoColSection = cmz(`
& {
  display: flex;
}

& > .${cx.header} {
  border: 1px solid blue;
}

& > .${cx.content} {
  border: 1px solid orange;
}
`)
```

Will not work as expected (see image below). This is not a bug, I can see it's the expected behaviour, just a note.

screen shot 2017-10-19 at 9 57 14 pm

joshwnj:
@bernardodiasc ah right, I see - the problem is that a `cmz` classname is not guaranteed to be a single classname. It might be a composition of multiple classnames, which means my solution above would not work.

Try this instead, which is the same idea but inverted:

```js
const twoColSection = 'twoColSection'

const Section = elem.section(cmz(`
& {
  margin: 0;
  padding: 1rem;
}

&.${twoColSection} {
  display: flex;
}
`))

const Header = elem.h1(cmz(`
& {
  margin: 0;
  padding: 0.5rem;
}

.${twoColSection} > & {
  border: 1px solid blue;
}
`))

const Content = elem.div(cmz(`
& {
  border: 1px solid red;
}

.${twoColSection} > & {
  border: 1px solid orange;
}
`))
```

bernardodiasc:
Quick note, found strange behaviour on cmz:

```
  elem.header(cmz(`
    & {
      cursor: pointer;
    }

    .twoColSection & {
      width: 200px;
    }
  `, typo.family.heading))
```

returns:
```
<header class="
    &amp; {
      cursor: pointer;
    }

    .twoColSection &amp; {
      width: 200px;
    }
   AutoUI_typo-25 AutoUI_typo-14">Panel title</header>
```

found my mistake, correct pattern for cmz composition should be like this:
```
  elem.header(cmz([`
      & {
        cursor: pointer;
      }

      .twoColSection & {
        width: 200px;
      }
    `, typo.family.heading]))
```

bernardodiasc:
## Current state

screen recording 2017-10-25 at 10 51 pm

bernardodiasc:
stuck with

```
Error: src/components/CollapsibleSection.js:4
  4: import { compose, withState, onlyUpdateForKeys } from 'recompose'
                                                           ^^^^^^^^^^^ recompose. Required module not found
```

recompose was added on package

```
  "dependencies": {
    "cmz": "^3.7.0",
    "color": "^2.0.0",
    "recompose": "^0.26.0",
    "webfontloader": "^1.6.28"
  }
```

🤔

joshwnj:
@bernardodiasc this is `flow` complaining that it can't find the type definition for that package. Follow the instructions here: https://github.com/x-team/auto/wiki/Flow-Definitions#add-definitions-for-common-libraries

joshwnj:
@bernardodiasc this is looking good! Please check the above 4 commits I pushed with some minor cleanups

bernardodiasc:
Nice updates @joshwnj, this PR is ready for review. :)

bernardodiasc:
I'm checking how to integrate auto-ui components on auto app. will continue on https://github.com/x-team/auto/issues/514 and https://github.com/x-team/auto/issues/513

joshwnj:
@bernardodiasc good work bringing `elem` and `cmz` back together, that cleans it up nicely.

Please also check https://github.com/x-team/auto-ui/pull/4/commits/691e996316384e2ab05dada169f2b1c9ebb1e822

This is important when we want to export css from one package to another, because by using symbols instead of hardcoded strings we avoid name collisions.

bernardodiasc:
hey, got it! getting my mind to this patterns :)

please, wait merge, I'm going to send one more commit, I put the click area in the entire section when collapsed. just checking details.

bernardodiasc:
there some cases that having the `cmz` separated from `elem`, I'm trying something like `section:hover h1 { color: red; }`, for instance. for that I would need to split `const Root = elem.section(cmz(` declaration again. is that correct?

bernardodiasc:
or something like `section h1:hover { background: silver; }`

bernardodiasc:
@joshwnj alright, found solution, if that last commit pass, then it's ready. the changes are subtle, but it improves experience. now when pass mouse hover section and title, on collapsed or not, will behave differently. this also good to stress `cmz` and `elem` a little bit more :)

joshwnj:
cool :) I've got 1 more change coming then we'll merge :)

joshwnj:
@bernardodiasc please check those last 3 commits to confirm.

The principle I find helpful here is to name the css atoms by what they mean on their own, rather than how the element is using them.

Eg. rather than saying _"the `.collapsed` class means the element should have an arrow"_ we can just make a `.arrow` class.

Following this principle we should end up with a better separation-of-concerns, and make it easier to modify in future (eg. if we decide we don't want an arrow later, we can just remove 1 line where it is composed).

joshwnj:
I also upgrade `cmz` to v3.8.0, which makes some simplifications as well as handling semicolon insertions properly, so the css source is cleaner.

joshwnj:
@bernardodiasc looks good!

PR #5, opened 2017-10-25
React.DOM -> react-dom package

kasprownik:
Hey @joshwnj, I'm upgrading react to v16 in `x-team/auto` and this is currently a blocker


joshwnj:
@kasprownik we should also upgrade `auto-ui` to use React v16 in that case, right?

PR #27, opened 2017-11-08
Result list item component

falcoagustin:
Created component that renders a collection of applicants, referring to [this issue](https://github.com/x-team/auto/issues/525) from `auto` repo. It consist on rendering the basic information of the applicants, including the tags that they have proven worth of and portfolio if there any.

Used as base @kasprownik [commit](https://github.com/x-team/auto/commit/5689a1dea96af2c2d7b87ee63e1a2aaeff291c36).

_**Notes:**_ We were in the need to add `crypto-js` in order to fetch the applicant's profile picture.

Preview: 
screen shot 2017-11-09 at 11 35 22 am


joshwnj:
Looking good @falcoagustin . But before we merge this, we can greatly simplify things.

## Remove some elements

- [ ] There are some things in this component that we don't need:

- "View in Base"
- Tags

We really just want avatar, name & email.

image

- [ ] Also, let's put the avatar and name in a single row, like the wireframes show.

- [ ] We don't want a "No applicants" state for this either. That is a state of the list, not of the list item.

## Naming

If we call it `ResultListItem` that implies a lot of contextual knowledge. What if we named it in a way that it doesn't know whether it's in a list or not? 

Let's consider what this component actually is. It's basically some identifying information about an applicant. Avatar, name and email.

- [ ] We could call it `ApplicantBadge`, and then say our result list is a list of badges. You click a badge to view that applicant's profile. (Feel free to suggest a different name too! 😄 )

Let's start with those simplifications, and review again from there.

bernardodiasc:
Nice review @joshwnj, I think it's better approach indeed. The PR looks good, I have 2 suggestions:

- `cropBody()` is one good example of code piece that can be unit tested. I believe I don't need to explain about this here :D but anyway, even being such small function, is exactly the case of pure function, that should return same value from same inputs, so good case to isolate (maybe or maybe not) and test different possibilities for inputs. My experience with this approach was very good, the benefits are many - mostly if tests are part of the development workflow, like a debug tool, but not going to list everything about here :)
- The first snapshot takes this `<ResultListItem applicant={null} />`, this is good approach, great to have. My suggestion is go further and let's use a default case like this for all UI components, that is, for snapshot testing, it's great to have `<ComponentName />` that takes no props, and call this test something like "does component explodes?". This is called "minimal render" test, basically helps on development, it ensure that the component don't break the app if occur bad usage of. You don't want the app to freeze because UI issues, much less to display unexpected values. About unexpected values, UI components should take care of all possibilities for it's render, imagine you see "undefined" in some of the number fields in you bank app, this would put your trust in the app on the ground, because now you can't be sure that the algorithms behind are good or even correct, imagine if the numbers are being calculated wrongly and you depende on that? It's pretty serious stuff, let that to be the Application implementation problem and release yourself of the search for bugs on UI code 😉

falcoagustin:
Amazing, will upload the changes right away, thanks for the suggestions @joshwnj and @bernardodiasc !

joshwnj:
@falcoagustin looking much better! 

One more simplification to make. As a general rule, always ask yourself whether you really need to pass an object as a prop, rather than just the values. Passing an object adds an extra layer of wrapping that is often unnecessary and unhelpful.

Eg. why

```js
<ApplicantsBadge applicant={
  {firstName: 'Applicant',
   lastName: 'X-Team',
   email: 'applicant@x-team.com',
  }
} />
```

when we could just

```js
<ApplicantsBadge 
  firstName='Applicant'
  lastName='X-Team'
  email='applicant@x-team.com'
/>
```

or, probably a more common (although lazier :P) example of how it will be used:

```js
<ApplicantsBadge {...applicant} />
```

src/components/ApplicantsBadge.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type Props = {
+  applicant: Applicant
+}
+
+class ApplicantsBadge extends PureComponent<Props> {
+  getApplicantsFieldsValues: Function
+
+  getApplicantsFieldsValues (applicant: Applicant) {
```
joshwnj: I don't think we need this function. We can assume that we will have all the required values, or none of them.

falcoagustin:
I completely Agree with you, it added unnecessary complexity to the design, thanks for noticing. I already fixed it, let me know if there are any more changes needed and I'll make them happen 😄

joshwnj:
@falcoagustin looks good. I've renamed `ApplicantsBadge` -> `ApplicantBadge`, as the component only relates to a single applicant.

Issue #32, opened 2017-11-13
Create applicants list filter component


falcoagustin:
Create `ApplicantListFilter` component. 

This component will have 3 checkboxes which will express the filtering of an applicants lists from their current status as it shows in [this](https://www.figma.com/file/KttDP3eJQaS1R79I2oByNtna/Auto-Lists) wireframe.

![applicantsstatusfilter](https://user-images.githubusercontent.com/15353019/32729384-4bafb68a-c862-11e7-8786-3afe40af2f27.png)

PR #33, opened 2017-11-13
32 applicant list filter component

falcoagustin:
Responds to #32, and its bind to [this](https://github.com/x-team/auto/issues/582) issue.

This component is used in `auto` as a filter for filtering applicants rendered with one of this three status: pending|accepted|excluded.

Receives 4 parameters, the amounts to be shown and a function to be executed once we check one of the boxes. If you want to execute different functions from the incoming checkbox, the way is to pass a function that receives an input parameter, this way:

```js  
handleCheck ({ target }: { target: HTMLInputElement }) {
  // target contains in 'name' property the name of the input coming from the executed change.
  (...)
}
```

**For the reviewer:** I took this PR in advantage to also change some styles from `ApplicantBadge.js` component.


src/components/ApplicantListFilter.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+class ApplicantListFilter extends PureComponent<Props> {
+  render () {
+    const { handleCheck, pending, accepted, excluded } = this.props
+    return (
+      <div>
+        <form className={optionContainer}>
+          <div className={option}>
+            <input type='checkbox' name='pending' onChange={handleCheck} />
```
bernardodiasc: Label tags are used for controlled inputs, in this case, if you are using the input outside the label, then you need to include `htmlFor` attribute that points to the `id` of the input.

Although that way is fine, I think that in many cases wrapping the input with the label is preferable, imho. That's because of the clicking area. On user experience perspective, having more area available for the click, not being strict to the input and the label text, can result better experience.

bernardodiasc:
👌 🚀

joshwnj:
@falcoagustin looks good. I made one change here: https://github.com/x-team/auto-ui/pull/33/commits/9c229860a0723f5944a27f2718aa5d95da7bc3af

Since the 3 items have almost identical code, I've replaced the duplication with a map

falcoagustin:
@joshwnj Yeah I thought about that but as it were only three items I thought it wasn't needed and it was clearer to explicitly add the 3 types instead of a map, either way though was fine for me 🙂

joshwnj:
@falcoagustin no worries :) 

For future reference, here's the reason I would argue that a `map` is clearer than longhand: when the 3 sections are written out explicitly, you have to look very closely to understand what the differences are between the 3.  There might be one small sneaky difference that you miss at first glance.  But if we use a `map` you can instantly see which parts are the same and which are not.

PR #36, opened 2017-11-13
HorizontalRuler component

bernardodiasc:
https://github.com/x-team/auto-ui/issues/21

screen shot 2017-11-13 at 3 24 15 pm


src/components/HorizontalRuler.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import { PureComponent } from 'react'
```
nicksp: You forgot to import `React`

bernardodiasc: I think the `React` is only required when using JSX syntax. Are you sure we need this @nicksp?

joshwnj: That's correct.

joshwnj:
👍

PR #37, opened 2017-11-14
Introduce Milestones component

### LINKED ISSUE #11: Milestones component

---

Create new component:

screen shot 2017-11-07 at 8 42 40 am

### Dependencies

- SvgIcon component https://github.com/x-team/auto-ui/issues/15 

### Props

```
level: number,
levels: [
  {
    label: string,
    icon: string,
    handleClick: function,
   }
],
```

### Behaviour

- `level` defaults to 1
- level dots are clickable and trigger `handleClick`
- (optional) `handleClick` are triggered after animation transition 
- `levels[i].label` prop is optional, defaults to the string `Level ${i}`

nicksp:
Closes #11 

### In action

screen recording 2017-11-14 at 03 14 pm

### Screen

localhost_6060_


nicksp:
Thanks @bernardodiasc Was fun to layout this component :) Added missing build file. Should be up on CI server soon.

src/components/Milestones.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  render () {
+    const { levels } = this.props
+
+    if (!levels || !levels.length) return null
+
+    return (
+      <ul className={styles.milestones}>
+        {levels.map((level, index) => this.renderMilestone(level, ++index))}
```
joshwnj: Any reason to use `++index` instead of `index + 1` ?

I think the latter is clearer, since we don't have any intention to change the value of `index`

src/components/Milestones.js
```diff
@@ -0,0 +1,209 @@
...
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translateX(-50%) translateY(-50%);
+      width: 36%;
+      height: auto;
+    }
+  `),
+
+  label: cmz(
```
joshwnj: This can be shortened to:

```js
  label: cmz(
    typo.family.base, `
    display: block
    etc...
  `)
```

joshwnj:
Just a couple of small things then we're good to merge.

src/components/Milestones.js
nicksp: Right, thanks for noticing :)

src/components/Milestones.js
nicksp: No specific reason, just kinda a habit ;-) Fixed that.

nicksp:
@joshwnj Addressed your comments.

joshwnj:
@nicksp FYI I applied this patch to make flowtype pass: https://github.com/x-team/auto-ui/commit/df213a3017be7c5bbdf28f35ce4bb685cc00abd3

nicksp:
@joshwnj Thanks for that. Read your rationals regarding that in another PR 👍 Will keep that in mind next time flow fails with a similar issue.

PR #38, opened 2017-11-15
26 upgrade page title

richa:


richa:
Fixes https://github.com/x-team/auto-ui/issues/26

joshwnj:
Thanks @richa . 

I notice that the red divider is much thicker in some places than others:

image


Also I wonder if we even need a `PageTitle` component? Would it be better to remove it and upgrade `auto` to use `<Text>` instead?

richa:
@joshwnj I see the red divider is thicker when it is used for `h1` and is thinner when used for `h2`. Shall I make it consistent (a thinner one) for all headings?

joshwnj:
@richa yes, it should always be the same thickness. Same as the old `PageTitle` was.

bernardodiasc:
I think the best approach would be completely replace `PageTitle` by `Text`, as @josh pointed out. :)

richa:
@joshwnj Replaced PageTitle with Text. PR is Ready for review

joshwnj:
Something's still not right here. There shouldn't be any divider at all unless the `hasDivider` prop is set:



Also please check those flowtype errors

joshwnj:
Looks good, thanks

PR #41, opened 2017-11-16
Add Button component

### LINKED ISSUE #14: Button component

---

Migrate https://github.com/x-team/auto/blob/master/src/client/components/Button.js and upgrade styles.

screen shot 2017-11-07 at 9 53 08 am

nicksp:
Closes #14 

Took a bit longer than planned initially, but mainly because I had to revise button styles and proposed props, and mainly re-structure it in a way that allows their combined usage, by applying new styles on top of base ones. As a result, we have a highly customisable component that allows to build it whatever we wish by combining various props :)

### Spec page

localhost_6060_

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,7 +1,6 @@
 // @flow
 
 import React, { PureComponent } from 'react'
-import cx from 'classnames'
 
 import type { CmzAtom } from 'cmz'
 import type { Element } from 'react'
@@ -17,7 +16,6 @@ type Props = {
   className: string | CmzAtom,
   size: Size,
   color: Color,
-  rounded: ?boolean,
   outlined: ?boolean,
   disabled: ?boolean,
   component: string,
@@ -122,10 +120,6 @@ const sizeStyles = {
 
 // Button variations
 const extraStyles = {
-  rounded: cmz(`
-    border-radius: 4px;
-  `),
-
   disabled: cmz(`
     &, &:hover {
       background: ${theme.lightGray};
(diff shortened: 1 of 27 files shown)


joshwnj:
@nicksp is there a need for the `rounded` prop? It looks like all of the buttons in the new design have square corners.

nicksp:
@joshwnj I was judging by new x-team.com website where I saw sever of rounded buttons. This prop is off by default and still might be a good addition to the component's possibilities.

joshwnj:
Ok - let's leave it out for now, so that we're not adding complexity with no value in return.

src/components/Button.js
```diff
@@ -0,0 +1,195 @@
...
+      disabled,
+      rounded,
+      component: CustomComponent,
+      children,
+      ...rest
+    } = this.props
+
+    const colorClassName = colorStyles[color] || ''
+    const sizeClassName = sizeStyles[size] || ''
+    const extraClassName = cx({
```
joshwnj: If we write it as:

```js
    const extraClassName = [
      rounded && extraStyles.rounded,
      outlined && extraStyles.outlined,
      outlined && 'outlined',
      disabled && extraStyles.disabled
    ].filter(Boolean).join(' ')
```

we don't need to add an extra package dependency.

nicksp: sure, will adjust the code. I just thought that having `classnames` in place might be helpful for future work to simplify class creation logic ;) Also, getting rid of `rounded` prop.

joshwnj: I'm fine with adding it, but only if we need it :) Once it's added, it's much more difficult to remove :)

joshwnj:
Looks good 👍

PR #44, opened 2017-11-16
Introduce VideoPlayer UI component

### LINKED ISSUE #18: VideoPlayer component

---

Migrate https://github.com/x-team/auto/blob/master/src/client/components/VideoPlayer.js and upgrade styles.

screen shot 2017-11-07 at 8 13 55 am

nicksp:
Closes #18 

Was a fun one :-) Actually, I made it work with two types of videos:
* embedded videos, like on our `CEOMessage` page, using `embedded` boolean prop
    Can be customised with `width`, `height` and whether to show controls (`showControls`)
* regular video files links, default state
   Can be customised with `width`, `height`, whether to show controls (`showControls`), and a `poster` to preload in place of video

### Spec page

localhost_6060

Now we can substitute all the usage of video controls in `Auto` with this component.

### DIFF
--- a/.snapguidist/__snapshots__/VideoPlayer-1.snap
+++ b/.snapguidist/__snapshots__/VideoPlayer-1.snap
@@ -1,0 +1,10 @@
+exports[`VideoPlayer-1 1`] = `
+<video
+  className="AutoUI_VideoPlayer-20"
+  controls={true}
+  poster="">
+  <source
+    src="https://files.slack.com/files-pri/T0257R0RP-F7SGYKZ5F/refactor.mp4" />
+  Video cannot be played in this browser.
+</video>
+`;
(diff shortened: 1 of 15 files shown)


joshwnj:
Looks good 👍

joshwnj:
@nicksp one way we can easily condense some of this code is by using the `elem` utility, which binds a classname to a react element: https://gist.github.com/joshwnj/8306d64927386ce87610a4361a9f007f

Any thoughts on pros / cons for these 2 approaches?  It would be good to standardize at some point.

nicksp:
@joshwnj I saw this is used in some of the components already and is similar in some way to [styled components](https://www.styled-components.com/) I'm just not used to this approach and assume this as a "magic", even though I understand what's happening under the hood. I'm more of a guy sticking with good well known stuff... but might move to newest approaches soon.

joshwnj:
Yeah, if you take a look at `elem.js` you'll see there's very little magic at all. It just calls `React.createElement`.  So in one sense, there's even less magic than using `jsx` :)

nicksp:
Agree. As I said, I understand how it works inside, it's just kinda new t me, a new way of thinking. But I'll come to it sooner or later :)

PR #46, opened 2017-11-17
Make use of flexbox and code cleanup

### LINKED ISSUE #43: Update RoadmapHero component to use flexbox
We'll have a much easier time using flexbox for layout, especially across breakpoints. 

If we can't fit on a single row let's horizontally center: 
image

nicksp:
Closes #43 

* make use of flexbox
* code cleanup

### Horizontal layout

horizontal-layout

### Vertical layout

vertical-layout


joshwnj:
👍

PR #48, opened 2017-11-20
10 - Create RoadmapTimelineElement

### LINKED ISSUE #10: RoadmapTimelineElement Component
Represents the single step from the timeline.

done: **false**
screenshot from 2017-11-06 14-01-34

done: **true**
screenshot from 2017-11-06 14-01-00

## Props
```
isDone: boolean
```

**Done**

Recognizes for the element state

galileo:
- [x] - Dot
- - [x] - pending
- - [x] - done
- [x] - Horizontal line
- - [x] - pending
- - [x] - done 

Fixes #10

### DIFF
--- a/src/components/RoadmapTimelineElement.js
+++ b/src/components/RoadmapTimelineElement.js
@@ -7,49 +7,54 @@ import theme from '../styles/theme'
 const cmz = require('cmz')
 
 type Props = {
-  isDone?: Boolean
+  isDone?: boolean
 }
 
 const styles = {
   timeline: cmz(`
-    & {
       position: relative;
       display: inline-block;
-      width: 0.25em;
-      height: 0.25em;
-      background-color: ${theme.white};
+      width: .25em;
+      height: .24em;
       border: 2px solid ${theme.offwhite};
+      background-color: ${theme.white};
       border-radius: 50%;
-      padding: .1em;
+      padding: .05em;
       max-width: 100%;
-      transition: all .25s ease-out;
-      z-index: 10;
-      cursor: pointer;
-    }
   `),
   checkmark: cmz(`   
-    &:after{
       content: '';
+      margin: -1px 0 0 0;
       display: block;
-   
       width: 0.1em;
       height: 0.2em;
-   
       border: solid ${theme.white};
       border-width: 0 1px 1px 0;
-   
       transform: rotate(45deg);
-    }
-  `)
+  `),
+  verticalLine: cmz(`
+    width: 2px;
+    margin: 0.4em 0 0 1px;
+    height: 3em;
+    background-color: ${theme.offwhite}
+    border: 0;
+`)
 }
 
 styles.active = cmz(
   styles.timeline,
   `
-    & {
-      background-color: ${theme.red};
-      border: 2px solid ${theme.red};    
-    }
+    background-color: ${theme.redHighlight};
+    border-color: ${theme.redHighlight};
+  `
+)
+
+styles.veritcalLineActivated = cmz(
+  styles.verticalLine,
+  `
+    margin: 0.2em 0 0 1px;
+    background-color: ${theme.red}
+    border-color: ${theme.red}
   `
 )
 
@@ -59,15 +64,17 @@ class RoadmapTimelineElement extends PureComponent<Props> {
   }
 
   render () {
-    const {isDone} = this.props
+    const { isDone } = this.props
 
-    const element = !isDone ? styles.timeline : styles.active
+    const wrapperClassName = isDone ? styles.active : styles.timeline
+    const verticalLineClassName = isDone ? styles.veritcalLineActivated : styles.verticalLine
+
+    const checkmarkClassName = isDone && styles.checkmark
 
     return (
-      <div>
-        <div className={element}>
-          <div className={isDone ? styles.checkmark : ''} />
-        </div>
+      <div className={wrapperClassName}>
+        <div className={checkmarkClassName}/>
+        <hr className={verticalLineClassName}/>
       </div>
     )
   }
--- a/src/components/RoadmapTimelineElement.js
+++ b/src/components/RoadmapTimelineElement.js
@@ -100,20 +100,22 @@ class RoadmapTimelineElement extends PureComponent<Props> {
   render () {
     const { isDone } = this.props
 
-    const activeTimelineClassName = isDone ? animatedStyles.timelineActive : ''
-    const checkmarkBoxClassName = animatedStyles.checkmarkBox
-    const checkmarkClassName = isDone ? animatedStyles.checkmark : ''
-    const circleClassName = `${animatedStyles.checkmarkCircle} ${isDone ? animatedStyles.checkmarkCircleActive : ''}`
-    const checkmarkCheckClassName = isDone ? animatedStyles.checkmarkCheck : ''
+    const cx = {
+      active: isDone ? animatedStyles.timelineActive : '',
+      box: animatedStyles.checkmarkBox,
+      checkmark: isDone ? animatedStyles.checkmark : '',
+      checkmarkCheck: isDone ? animatedStyles.checkmarkCheck : '',
+      circle: `${animatedStyles.checkmarkCircle} ${isDone ? animatedStyles.checkmarkCircleActive : ''}`
+    }
 
     return (
-      <div className={`${animatedStyles.timeline} ${activeTimelineClassName}`}>
-        <div className={checkmarkBoxClassName}>
-          <svg className={checkmarkClassName}
+      <div className={`${animatedStyles.timeline} ${cx.active}`}>
+        <div className={cx.box}>
+          <svg className={cx.checkmark}
             xmlns='http://www.w3.org/2000/svg'
             viewBox='0 0 52 52'>
-            <circle className={circleClassName} cx='26' cy='26' r='25' fill='none' />
-            <path className={checkmarkCheckClassName} fill='none' d='M14.1 27.2l7.1 7.2 16.7-16.8' />
+            <circle className={cx.circle} cx='26' cy='26' r='25' fill='none' />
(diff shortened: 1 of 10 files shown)


galileo:
I am just starting the UI work. I know that there is a lot things to improve but I want to learn this thing as fast as possible so may I trouble you to watch this in this state as it is?

My inspiration was `milestone` component. 
I asked @nicksp how to implement the `checkmark` and I received information that I can `render checkmark with css purely` so I searched internet for solution, and you can find it here. 
**Q.1.** How can I render it better?
**Q.2.** My cirlce is not perfect circle, any direct thoughts how can I solve it?
**Q.3.** Any, idea what would be the best approach to create this verical bar?

Any additional advices and resources will be appreciated.

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,76 @@
...
+
+  render () {
+    const {isDone} = this.props
+
+    const element = !isDone ? styles.timeline : styles.active
+
+    return (
+      <div>
+        <div className={element}>
+          <div className={isDone ? styles.checkmark : ''} />
```
falcoagustin: Can be reduced to `isDone && styles.checkmark` in order not to use ternary operator :)

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,76 @@
+// @flow
+
+import type { Element } from 'react'
+import React, { PureComponent } from 'react'
+import theme from '../styles/theme'
+
+const cmz = require('cmz')
+
+type Props = {
+  isDone?: Boolean
```
nicksp: `boolean` type should be defined with lowercase `b` https://flow.org/en/docs/types/primitives/#toc-booleans

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const styles = {
+  timeline: cmz(`
+    & {
```
nicksp: If this is a single element styles, code might be shortened by excluding `& {}`

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,76 @@
...
+   
+      transform: rotate(45deg);
+    }
+  `)
+}
+
+styles.active = cmz(
+  styles.timeline,
+  `
+    & {
```
nicksp: no need for extra `& {}`

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,76 @@
...
+  `
+)
+
+class RoadmapTimelineElement extends PureComponent<Props> {
+  static defaultProps = {
+    isDone: false
+  }
+
+  render () {
+    const {isDone} = this.props
```
nicksp: Put spaces inside curly braces for readability and for consistency with the rest of code.

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,76 @@
...
+
+    const element = !isDone ? styles.timeline : styles.active
```
nicksp: give a variable more meaningful name. This is not an element technically, it's a class name. Smth like `elementClassName` would work better.

nicksp:
@galileo here is a good example of how you can create the component https://codyhouse.co/gem/vertical-timeline/ There you'll also find how to create a vertical line using pseudo element and absolute positioning.

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,76 @@
...
+
+    return (
+      <div>
```
bernardodiasc: this extra `<div>` is probably unnecessary :)

src/components/RoadmapTimelineElement.md
```diff
@@ -0,0 +1,15 @@
...
+
+Active state:
+
+```
+<RoadmapTimelineElement
+    isDone
+/>
+```
+
+Missing props (Watch basic state):
```
bernardodiasc: I think the default usage of this component is already without any props, so this is not really needed. :)

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,83 @@
...
+type Props = {
+  isDone?: boolean
+}
+
+const styles = {
+  timeline: cmz(`
+      position: relative;
+      display: inline-block;
+      width: .25em;
+      height: .24em;
```
nicksp: to make it a perfect circle, you have to make width equal to height.

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,83 @@
...
+      border: 2px solid ${theme.offwhite};
+      background-color: ${theme.white};
+      border-radius: 50%;
+      padding: .05em;
```
nicksp: Give it a better spacing around, smth like `.2em` looks good.

src/components/RoadmapTimelineElement.md
```diff
@@ -12,4 +12,18 @@ Active state:
 />
 ```
 
-Missing props (Watch basic state):
+Composition:
+
+```
+<div>
```
bernardodiasc: This could be simplified to 
```
<div>
    <RoadmapTimelineElement isDone />
    <RoadmapTimelineElement />
</div>
```

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,83 @@
...
+      max-width: 100%;
+  `),
+  checkmark: cmz(`   
+      content: '';
+      margin: -1px 0 0 0;
+      display: block;
+      width: 0.1em;
```
nicksp: Make width equal to height, and the checkmark should be centered ;-)

src/components/RoadmapTimelineElement.md
bernardodiasc: But I guess the composition usage will depend more on how https://github.com/x-team/auto-ui/issues/19 will be made.

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,83 @@
...
+      display: inline-block;
+      width: .25em;
+      height: .24em;
+      border: 2px solid ${theme.offwhite};
+      background-color: ${theme.white};
+      border-radius: 50%;
+      padding: .05em;
+      max-width: 100%;
+  `),
+  checkmark: cmz(`
```
nicksp: In order to make checkmark render a little bit better, I found the following styles to make a better job. Just check it and apply if you like it more:

```
checkmark: cmz(`
  &:after {
    content: '';
  display: block;
  width: 0.1em;
  height: 0.3em;
  border: solid #fff;
  border-width: 0 2px 2px 0;
  transform: rotate(45deg);
  }

  & {
    display: block;
  transform: translateY(-50%);
  position: relative;
  top: 50%;
  }
`)
```

src/components/RoadmapTimelineElement.js
```diff
@@ -0,0 +1,83 @@
...
+  static defaultProps = {
+    isDone: false
+  }
+
+  render () {
+    const { isDone } = this.props
+
+    const wrapperClassName = isDone ? styles.active : styles.timeline
+    const verticalLineClassName = isDone ? styles.veritcalLineActivated : styles.verticalLine
+
```
nicksp: This empty line can be removed

src/components/RoadmapTimelineElement.md
galileo: I would try to make it but if I am trying to do it those elements are looking like this.

Is there a possiblity to add make it work in my element?

screenshot from 2017-11-20 20-37-46

src/components/RoadmapTimelineElement.js
galileo: Are you sure that this is still valid regarding your additional comments below?

src/components/RoadmapTimelineElement.js
galileo: Thank you!

src/components/RoadmapTimelineElement.js
galileo: Removed

src/components/RoadmapTimelineElement.js
nicksp: Nope, the styles below are valid ones.

src/components/RoadmapTimelineElement.md
bernardodiasc: I think you don't really need to worry with composition in this component. This is a small piece that will be used by another component responsible for the composition. The problem you pointed is related to layout, which is concern of the context of the component, not of the component itself :)

galileo:
I improved this solution with svg graphics, which is behaving now much better (in my opinion).
I also used some animations for our timeline, and fixed all errors. 

The composition is looking now much better.
screenshot from 2017-11-21 13-54-08


@nicksp 
As you are working on the composition component would you require some actions to control the component state?

nicksp:
@galileo Not sure yet, if I'd require anything above `isDone`. That should be sufficient for now I guess. But things will get clear once I start implementing the big component.

nicksp:
@galileo With some small changes that I committed, the component now looks good and ready to be shipped :) Mainly condensed some code by removing unnecessary chars, made the animation a bit faster for end users to perceive the component earlier kinda...

@joshwnj once you agree with changes and final look, let's merge it, so the "big" Roadmap component has everything it depends on in place.

joshwnj:
I just made one small cleanup too. Looks good, thanks @galileo & @nicksp

bernardodiasc:
Hey folks, this looks good. But maybe (just maybe), eventually, you'll notice that the **composition is a concern of the parent component** that will use this one. This is small piece of UI that usually don't know it's context. Whenever you feel what I'm talking about here when working on https://github.com/x-team/auto-ui/issues/19, you'll get the idea. But well, I never know for sure, the great minds here can solve the problem differently from what I'm thinking. And I'm enjoying the awesomeness of these contributions. 😄 👍

PR #50, opened 2017-11-21
Changes in RoadmapHero to support similar Next steps component

richa:
need to revert RoadmapHero changes to accept imgUrl as props


bernardodiasc:
Hey, I have suggestions for this :)

I'm not sure if `NextStepsHero` is really needed. As far as I can see, this component is just a wrapper on the `RoadmapHero`. The approach for UI components that I believe it works well, **makes the data providing an Application concern**, and since `RoadmapHero` has now the `imgUrl` being provided as data, it would be much better, imho, to have a generic `Hero` component and let the integration in the app to provide the different content for each case.

Also, for this reason I initially suggested the `Graphic` component (https://github.com/x-team/auto-ui/issues/13), that would be an "image assets catalogue" that could simply return the desired images  (notice -> image assets, not image contents) by using an ID, something very similar to `SvgIcon` (https://github.com/x-team/auto-ui/blob/master/src/components/SvgIcon.js).

That is how I would approach on this situation, but of course, it's really just a suggestion, after all, the approach on this PR is also fine and has nothing wrong with it. :)

/cc @joshwnj

nicksp:
Regarding introducing new component `NextStepsHero`. I strongly believe that there is no need to introduce new component and thus making us maintain more things. All it does is just specifying an image. No need to complicate UI components by wiring new props. It's better to utilize composition, and provide an image for `RoadmapHero` as a child. This way the current component will remain intact and allow flexible way of specifying image by utilizing another existing UI svgicon component. Composition is great :)

richa:
Hi @bernardodiasc and @nicksp . I completely agree that a new NextSteps component is not needed. Since the RoadmapHero component didn't show the image on an `auto` component, I tried to temporarilty hardcode the path of the image. But the component didn't load the image when the image path is set using imgUrl props. That's why I tried creating another temporary component `NextSteps` and check if the image is loaded using this new component. But that didn't work too :)

That's why I have this PR in WIP. Thanks for reviewing. I will share the solution that works.

PR #53, opened 2017-11-22
Compose Roadmap component

### LINKED ISSUE #19: Roadmap component

---

Create new component.

slice 1

### Dependencies

- RoadmapLevel component https://github.com/x-team/auto-ui/issues/8
- RoadmapTimelineElement component https://github.com/x-team/auto-ui/issues/10
- Copy component https://github.com/x-team/auto-ui/issues/12
- SvgIcon component https://github.com/x-team/auto-ui/issues/15
- Button component https://github.com/x-team/auto-ui/issues/14

### Props

```
children: node|array,
```

### Behaviour

This component is the layout styles wrapper for Roadmap, the `children` prop should be the render itself, receiving RoadmapLevel, RoadmapTimelineElement, etc.

Example usage:

```
<Roadmap>
  <RoadmapLevel icon="icon-id" heading="lorem ipsum 1" level=1 body={body1Content} isActive />
  <RoadmapLevel icon="icon-id" heading="lorem ipsum 2" level=2 body={body2Content} />
  <RoadmapLevel icon="icon-id" heading="lorem ipsum 3" level=3 body={body3Content} />
  <RoadmapLevel icon="icon-id" heading="lorem ipsum 4" level=4 body={body4Content} />
  <RoadmapLevel icon="icon-id" heading="lorem ipsum 5" level=5 body={body5Content} />
  <RoadmapLevel icon="icon-id" heading="lorem ipsum 6" level=6 body={body6Content} />
  <RoadmapLevel icon="icon-id" heading="lorem ipsum 7" level=7 body={body7Content} />
</Roadmap>
```

nicksp:
This is the component so far. Somewhat done-ish, but lacks some of the stylings

localhost_6060_

As you can see, it's a bit tricky to stylize it in a way required. First of all, `RoadmapLevel` styles are not in sync with the picture, there is a lot of spacing presented. Also, Each level block might have content of different height, and vertical timeline element should go in line with it.... And each block should be centered relative to the sibling column's element. And each timeline element should have a specific height to align with adjacent elements the way, that the first timeline element aligned with of the top of prev level, next one - with the top of next element, and each connected to each other.... Probably we'll need to find some way to make timeline element to adapt it's height dynamically based on context 🤔 

/cc @joshwnj @bernardodiasc If you guys have any suggestions/improvements I'd appreciate your help.
**P.S.** CSS gurus're required here :-)

Fixes #19


joshwnj:
Looking good so far @nicksp !

I think it probably makes sense to use some javascript for the styling / colouring of the vertical line. Especially as the requirements state the red colour should spread as you scroll down: https://github.com/x-team/auto/issues/573

Also, remember that the Final Level is different from the others:

image

bernardodiasc:
I would try something like this:

screenshot-github com 2017-11-22 20-36-30-094

- Don't use custom markup outside the component, instead make a loop/clone of each children wrapping with the sections that will provide layout styles
- Make sure that layout styles are in Roadmap component, not on Timeline or RoadmapLevel, the non-layout components should only knows about themselves
- Overlap sections above the previous ones
- Don't use float for layout, that was hacky stuff we used in the past, we have flexbox and grids already 😄 
- Use and abuse flexbox, make Timeline element to fill the height of the section

nicksp:
When you scroll down, the element whose bottom is above viewport’s bottom line (and when it first gets visible to the user), gets highlighted and marked as `done`. And it makes sense, as we scroll down and can’t see the level’s bottom yet, we assume it to not be accomplished yet.

When you scroll the opposite side (scroll up), we reverse the logic and highlight only those levels, whose top is above viewport’s bottom line, which makes sense, as while scrolling up and seeing any part of the level element, we assume it to be `done`.

Hope my logic is kinda clear and makes sense to you

joshwnj:
@nicksp I've had a crack at making the flexbox layout work, see `Roadmap2.js`: https://github.com/x-team/auto-ui/pull/53/commits/4962e7f4e76a163414f12ed371ad0c71bbbd29f2

image

nicksp:
Hey @joshwnj I derived markup and styles from your example and integrated it into my solution. It looks really good now and all the dynamic behavior is in place.

Let me know if you like it and agree with the [way I put highlighting logic](https://github.com/x-team/auto-ui/pull/53#issuecomment-347375050).

As a wrapup: This took a bit of wrapping my head around, but now I have much better understanding of how alter DOM based on scroll position, see differences between document and viewport units and stuff related to that using pure DOM API :)

localhost_6060_

joshwnj:
@nicksp looks good! I think we'll still need to tweak some of the transitions as we go (eg. I noticed some flickering when at the bottom and scrolling up) but this is a great starting point.

nicksp:
@joshwnj I'm happy it went that far. Sure, we can tweak it as we go, maybe we'd even want to change the scroll logic a bit, but by merging that, we don't stop developing of redesign feature, as one of the pages relies on this component.

PR #55, opened 2017-11-27
Prevent firing default onClick event when custom function is provided

RafalWilinski:
Related to: https://github.com/x-team/auto/pull/658#issuecomment-347052788

"milestone nav shouldn't be clickable"

Now, `handleClick` function overrides default action instead of extending it.

### DIFF
--- a/src/components/Milestones.js
+++ b/src/components/Milestones.js
@@ -175,8 +175,8 @@ class Milestones extends PureComponent<Props, State> {
   }
 
   handleChangeMilestone = (level: Level, newLevel: number) => () => {
-    this.setState({ currentLevel: newLevel })
     if (typeof level.handleClick === 'function') level.handleClick()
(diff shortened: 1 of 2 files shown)


src/components/Milestones.js
```diff
@@ -175,8 +175,8 @@ class Milestones extends PureComponent<Props, State> {
   }
 
   handleChangeMilestone = (level: Level, newLevel: number) => () => {
-    this.setState({ currentLevel: newLevel })
     if (typeof level.handleClick === 'function') level.handleClick()
+    else this.setState({ currentLevel: newLevel })
```
nicksp: What it says and does is executing custom callback if it was provided. If it's not (as in our case in parent ticket), we change milestone manually. But that's exactly what we want to prevent. So I can see no point in having and approving this PR, thus suggesting to close it for now and find a solution to the problem first, and only then fix it.

PR #56, opened 2017-11-27
Handle missing name-related props

nicksp:
Fixes https://github.com/x-team/auto/issues/654

**Type of Release:** Fix

* Provide defaults for missing props
* Only render name row if any of first name / last name is provided
* Extend spec page with more cases to prove it works in different scenarios:
    * missing `firstName`
    * missing `lastName`
   * missed both props
* Handle avatar's alt text properly

### Spec page

localhost_6060_


joshwnj:
lgtm

PR #58, opened 2017-11-28
Allow changing milestone with a prop only

### LINKED ISSUE #57: Alter 'Milestones' component
This complements https://github.com/x-team/auto/pull/658 PR based on the [comment](https://github.com/x-team/auto/pull/658#issuecomment-347198553)

nicksp:
Closes #57 

I assume this tiny change should be sufficient for what we look for.


joshwnj:
@nicksp looks good, but let's get rid of `this.state` entirely.

nicksp:
@joshwnj Right, thanks for noticing that. We don't use state anymore, so no point to keep it. Fixed.

PR #59, opened 2017-11-30
Fix <HeaderBar /> shadow

RafalWilinski:
Related to: https://github.com/x-team/auto/pull/698

Before:
before

After:
after


nicksp:
Copied `box-shadow` value from the Sketch, tweaked component a little bit and committed a build file. @RafalWilinski Merging this, please take a note, that each PR to `auto-ui` (once everything is completed) requires a build file with running `npm run build`.

PR #60, opened 2017-12-01
Tweak ApplicantBadge component

nicksp:
This complements https://github.com/x-team/auto/pull/697 by making `ApplicantBadge` more flexible

* Add active state to highlight the whole text content area
* Remove the padding from the lefthand side
* Give a top and right padding to the text content

### For reviewers

Should be merged first, before the https://github.com/x-team/auto/pull/697 PR.

### DIFF
--- a/.snapguidist/__snapshots__/ApplicantBadge-1.snap
+++ b/.snapguidist/__snapshots__/ApplicantBadge-1.snap
@@ -1,11 +1,11 @@
 exports[`ApplicantBadge-1 1`] = `
 <div
-  className="AutoUI_ApplicantBadge-6">
+  className="AutoUI_ApplicantBadge-8 false">
   <img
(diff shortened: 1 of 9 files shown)


joshwnj:
@nicksp looks good, I've just made a couple of cleanups:

- un-namespaced the style constants
- `activeApplicantClassName` only needs to be applied once on the container, not twice
- used ternary instead of boolean for conditional class application, to avoid the string `false` being rendered to the DOM

nicksp:
@joshwnj the second `isActive` class name was required, by removing it you got back to the state where we had a bug on long email addresses:

image

Fixed that on `master` branch

joshwnj:
@nicksp hmm but `.isActive` is only changing the background colour. Isn't this an overflow problem?

PR #61, opened 2017-12-01
Tweak footer, fix for situation when more then 4 brands, alignment fix

justtry:
https://github.com/x-team/auto/issues/672

### DIFF
--- a/lib/auto-ui.js
+++ b/lib/auto-ui.js
@@ -6796,11 +6796,11 @@ var cmz = __webpack_require__(1);
 
 var cx = {
   footer: cmz.named('AutoUI_Footer_Footer-19', /*cmz|*/'\n    padding: 6.5rem 0;\n  ' /*|cmz*/),
-  columnContainer: cmz.named('AutoUI_Footer_Footer-22', /*cmz|*/'\n    display: flex;\n  ' /*|cmz*/),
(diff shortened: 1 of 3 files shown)


nicksp:
On mid-size screens to small screens i can see no space between brand images. Compare sketch file:

image

vs current view:

image

In current deployed master specs it looks fine though https://x-team.github.io/auto-ui/#!/FooterBrands

@justtry Also, don't forget to update specs when making PR RFR and run `npm run build` command and commit changes.

image

justtry:
@nicksp thx, run build, margin - added

nicksp:
@justtry Yay, now it looks almost great :) Thanks for the fix. As the last thing before we merge this finally, do you think we can do smth to account for this case on small screens when heading overlaps with the x-logo?

image

justtry:
@nicksp thx, we have a lot of columns and brands, not like on this styleguide, so I think it's not about overlapping but more about having responsive version of site (roadmap and all other elements too)
Thought that responsive is not part of this tickets, but totally can do in terms of this ticket.

nicksp:
@justtry Agree, let's tackle responsive separately once the redesign is completed. Will require applying responsive styles per UI component, by going through all the existing ones. Merging this PR 🎉

justtry:
@nicksp yeah, PR from fork repo

PR #67, opened 2017-12-07
Theme variables

### LINKED ISSUE #25: Provide further coverage of style values using variables
Provide further coverage of style values using variables on https://github.com/x-team/auto-ui/tree/master/src/styles

bernardodiasc:
Closes #25 

screen shot 2017-12-06 at 10 02 48 pm


nicksp:
Soo hard to spot a difference between some of the tones 😆 I.e. `baseRed`, `typoSubheading` and `lineRed` look almost identical to me... will probably need to calibrate the monitor 😃

bernardodiasc:
@nicksp you don't need to calibrate the monitor, they are the same red (called `radicalRed`) but the variables are for different purposes. This is in progress, not final, I did this way to do progress without breaking existing styles and without losing track of where variables are being used (if I had used same red variable name everywhere I was go crazy with batch search&replace).

This is the part that matters https://github.com/x-team/auto-ui/blob/f8f12b947d71d63c2cdcbf1478fcc709ac1377e3/src/styles/theme.js#L17-L67

nicksp:
👍 makes sense @bernardodiasc. While you're doing that, I'm working on unifying typography. Hope to get it done soon and along with what you're doing, clean up some code and make component consistent.

bernardodiasc:
IMHO this "purposed variable names" are cool. I'm used to use this way because it gives context meaning and free the variable from the color value, for instance, the red used for typography may change while other reds remains - it's a possibility but as a speculation, don't often happen really.

In most cases the base red will not change and many developers don't like to do this way because it's ends being unnecessary abstraction. I still think it's better, think about, for the typography you don't need to worry about general colours in the code base, you'll just need to look at `theme.typoHeading`, `theme.typoSubheading`, `theme.typoParagraph` and `theme.typeHighlight`, they are all the colours present in the Sketch file for copy elements. There are also `theme.formText` for the text color in input/textareas, the name is self explanatory...

Well, it's easy to reason about. But anyway, let me know if just the `radicalRed` for every red color in the codebase would be better approach, this is easy to change (but very hard to revert!).

bernardodiasc:
I'm done with this 😄  Please review.

The 4 silver lines have comments with information of where it's being used because I couldn't find better meaningful names. Feel free to recommend better solution.

```
   // roadmap timeline circle 
   // rodamap level border 
   lineSilver1: palette.mercury, 
   // milestones circle and line 
   lineSilver2: palette.porcelain, 
   // roadmap timeline line 
   lineSilver3: palette.athensCray, 
   // horizontal ruler 
   // collapsible section divisors 
   lineSilver4: palette.athensCrayAlt, 
```

Notes:

These silvers aren't made to be simply random greyscale colours, just doesn't work to make them to be like that. Here's the design logic:

- The `mercury` is kinda the main silver for borders and lines that aren't too much related to text.
- The `athensCray` are used only in the Roadmap vertical line, because it's a lighter version of `mercury`.
- The `athensCrayAlt` color is used for silver lines/borders that are close to text, because it's slightly blue, things like horizontal rulers and input/textarea borders.
- The `porcelain` are used only on Milestone, could be easily replaces by `athensCray` (it looks better imho).

nicksp:
@bernardodiasc Did some small changes. Overall I really like how you put it and structured. Really convenient way. As one last improvement before merging I'm thinking about one thing that might be useful and handy inside `auto` repo. Given that we added `ColorPalette` component just to showcase existing color props, what if we export those variables from the component so that when imported into `auto` we can just re-use them :) We'll need to tweak colours used in `auto` anyways and to eliminate duplication we could just make the `ColorPalette` component usable. I find it smart and useful. Wdyt?

bernardodiasc:
Exposing colours from the component is a very smart idea indeed!... IF Auto will still going to contain UI parts. IMHO the better would be move everything UI-related to Auto UI, making this move unnecessary. 😄 

I know I'm repeating myself a lot about this. The whole point of having Auto UI is to not have any UI code at all in the Application. But yeah, there are some road ahead until get there, meanwhile exposed colours can come handy 👍

nicksp:
So, shall we plan it for a separate issue or would you prefer doing that in the scope of current PR?

bernardodiasc:
Please make separate issue. Otherwise will delay too much this one. This one is useful as it is already :)

nicksp:
Made some changes, as the `+/-` button on `CollapsibleSection` component was broken and not clickable when expanded. Fixed that now and updated build.

PR #68, opened 2017-12-07
Include footer component into pages

bernardodiasc:
https://github.com/x-team/auto/issues/722

### DIFF
--- a/lib/auto-ui.js
+++ b/lib/auto-ui.js
@@ -7136,7 +7136,7 @@ var FooterBrands = function (_PureComponent) {
               return _react2.default.createElement(
                 'li',
                 { className: cx.brand, key: id },
-                url ? _react2.default.createElement(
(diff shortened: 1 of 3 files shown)


nicksp:
I personally don't feel we need all these changes in a PR. `url` is required prop, and if it's empty will just render non-working `a`. So markup will stay valid and fully operational. So suggest closing without merging unless there're strong arguments

bernardodiasc:
Empty `url` will make links working for same page. This looks like buggy FE to me.

bernardodiasc:
@nicksp I see what you said. We can close this one I guess.

bernardodiasc:
I still prefer to make brand urls optional instead sending empty params and having non-working `<a>`... but let's just move on. :)

nicksp:
@bernardodiasc Good idea, we can get back to it when we feel it's not flexible. For current purposes it works well. Let's keep an eye on it 👍

PR #77, opened 2017-12-12
Review typography

### LINKED ISSUE #42: Review typography
Now that many of the components are in place, we need to do a review of the sketch file and ensure we are using the correct font faces / sizes / spacing.

nicksp:
Closes #42

Unifies and updates typography usage in the UI lib.

### DIFF
--- a/.snapguidist/__snapshots__/Roadmap-1.snap
+++ b/.snapguidist/__snapshots__/Roadmap-1.snap
@@ -6,7 +6,7 @@ exports[`Roadmap-1 1`] = `
     <div
       className="AutoUI_Roadmap-77">
       <section
-        className="AutoUI_RoadmapLevel-30 AutoUI_RoadmapLevel-45">
+        className="AutoUI_RoadmapLevel-31 AutoUI_RoadmapLevel-46">
         <svg
           height="43px"
           viewBox="0 0 37 43"
@@ -23,7 +23,7 @@ exports[`Roadmap-1 1`] = `
                 Object {
                   "color": Array [
                     246,
-                    58,
+                    57,
                     84,
                   ],
                   "model": "rgb",
@@ -49,17 +49,17 @@ exports[`Roadmap-1 1`] = `
           </g>
         </svg>
         <section
-          className="AutoUI_Text-21">
+          className="AutoUI_Text-26">
           <h2
-            className="AutoUI_Text-39">
+            className="undefined AutoUI_typo-102 AutoUI_typo-53 AutoUI_typo-22">
             The challenge
           </h2>
           <p
-            className="AutoUI_Text-47">
+            className="undefined AutoUI_typo-116 AutoUI_typo-53 AutoUI_typo-28">
             Level 1
           </p>
           <div
-            className="AutoUI_Text-54 AutoUI_typo-21 AutoUI_typo-15">
+            className="AutoUI_Text-41 AutoUI_typo-138 AutoUI_typo-53 AutoUI_typo-40">
             Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur, fuga suscipit rem voluptatibus sunt sapiente. Molestiae aliquam at libero repellat, asperiores, recusandae, praesentium ratione officia molestias nobis temporibus ipsam, repudiandae! Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur, fuga suscipit rem voluptatibus sunt sapiente. Molestiae aliquam at libero repellat, asperiores, recusandae, praesentium ratione officia molestias nobis temporibus ipsam, repudiandae!
           </div>
         </section>
@@ -96,7 +96,7 @@ exports[`Roadmap-1 1`] = `
     <div
       className="AutoUI_Roadmap-77">
       <section
-        className="AutoUI_RoadmapLevel-30 ">
+        className="AutoUI_RoadmapLevel-31 ">
         <svg
           height="43px"
           viewBox="0 0 30 43"
@@ -113,7 +113,7 @@ exports[`Roadmap-1 1`] = `
                 Object {
                   "color": Array [
                     246,
-                    58,
+                    57,
                     84,
                   ],
                   "model": "rgb",
@@ -141,17 +141,17 @@ exports[`Roadmap-1 1`] = `
           </g>
         </svg>
         <section
-          className="AutoUI_Text-21">
+          className="AutoUI_Text-26">
           <h2
-            className="AutoUI_Text-39">
+            className="undefined AutoUI_typo-102 AutoUI_typo-53 AutoUI_typo-22">
             Take the stage
           </h2>
           <p
-            className="AutoUI_Text-47">
+            className="undefined AutoUI_typo-116 AutoUI_typo-53 AutoUI_typo-28">
             Level 2
           </p>
           <div
-            className="AutoUI_Text-54 AutoUI_typo-21 AutoUI_typo-15">
+            className="AutoUI_Text-41 AutoUI_typo-138 AutoUI_typo-53 AutoUI_typo-40">
             Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur, fuga suscipit rem voluptatibus sunt sapiente. Molestiae aliquam at libero repellat, asperiores, recusandae, praesentium ratione officia molestias nobis temporibus ipsam, repudiandae! Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur, fuga suscipit rem voluptatibus sunt sapiente. Molestiae aliquam at libero repellat, asperiores, recusandae, praesentium ratione officia molestias nobis temporibus ipsam, repudiandae! Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur, fuga suscipit rem voluptatibus sunt sapiente. Molestiae aliquam at libero repellat, asperiores, recusandae, praesentium ratione officia molestias nobis temporibus ipsam, repudiandae!
           </div>
         </section>
@@ -188,7 +188,7 @@ exports[`Roadmap-1 1`] = `
     <div
       className="AutoUI_Roadmap-77">
       <section
-        className="AutoUI_RoadmapLevel-30 ">
+        className="AutoUI_RoadmapLevel-31 ">
         <svg
           height="36px"
           viewBox="0 0 37 36"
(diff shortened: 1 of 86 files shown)


nicksp:
@bernardodiasc @joshwnj This is ready for review.

.snapguidist/__snapshots__/Roadmap-1.snap
```diff
@@ -49,17 +49,17 @@ exports[`Roadmap-1 1`] = `
           </g>
         </svg>
         <section
-          className="AutoUI_Text-21">
+          className="AutoUI_Text-26">
           <h2
-            className="AutoUI_Text-39">
+            className="undefined AutoUI_typo-102 AutoUI_typo-53 AutoUI_typo-22">
```
bernardodiasc: There are many `undefined` classes in the snapshots, this can be caused by some inexistent class.

nicksp: Good catch mate. Let me double-check that and fix.

src/components/CollapsibleSection.js
```diff
@@ -80,6 +80,10 @@ const Content = elem.div(cmz(`
     width: calc(100% - 200px)
   }
 
+  & p {
```
bernardodiasc: Is there a guarantee that all text blocks will be used with `<p>`? I think it should, that's what's expect, but we need make sure of that, if someone includes text body inside only `<div>` the spacing harmony may result differently.

nicksp: Yea, I thought about that. Not sure how to make it work for each case and just for the child, as it's passed by the end user.

src/components/ApplicantBadge.js
```diff
@@ -37,11 +39,13 @@ const Info = elem.div(cmz(`
 `))
 
 const Name = elem.div(cmz(
-  family.smallHeading, `
+  typo.smallHeading, `
   letter-spacing: -.025em
   font-size: 1.2em
```
bernardodiasc: We have font-size and other props overriding the `typo.smallHeading` defaults, is this really desired?

I'm also worried about the usage of `em` and `rem` (relative units) here, since they will cause difference depending on the context. For instance, if this component is included inside another component that set the font-size to something different from the body (I guess is 20px https://github.com/x-team/auto-ui/blob/fa5a878855e247925381ceb5fc671ca0c01b9152/src/styles/typo.js#L142), it might result different from expected. The quick solution would be use `rem` instead, if still wants to use relative units. But I still unsure if these overidings should happen at all.

.snapguidist/__snapshots__/Roadmap-1.snap
bernardodiasc: There are many `undefined` classes in the snapshots, this can be caused by some inexistent class.

src/components/ApplicantBadge.js
bernardodiasc: We have font-size and other props overriding the `typo.smallHeading` defaults, is this really desired?

I'm also worried about the usage of `em` and `rem` (relative units) here, since they will cause difference depending on the context. For instance, if this component is included inside another component that set the font-size to something different from the body (I guess is 20px https://github.com/x-team/auto-ui/blob/fa5a878855e247925381ceb5fc671ca0c01b9152/src/styles/typo.js#L142), it might result different from expected. The quick solution would be use `rem` instead, if still wants to use relative units. But I still unsure if these overidings should happen at all.

src/components/CollapsibleSection.js
bernardodiasc: Is there a guarantee that all text blocks will be used with `<p>`? I think it should, that's what's expect, but we need make sure of that, if someone includes text body inside only `<div>` the spacing harmony may result differently.

src/components/Milestones.js
```diff
@@ -120,17 +120,18 @@ const styles = {
...
-    position: absolute;
-    padding-top: .2em;
-    width: 100%;
-    transition: all .25s ease-out;
-  `)
+    typo.tinyHeading,
+    `
+      display: block
+      text-transform: uppercase
+      font-size: 12px
```
bernardodiasc: This is another case with typography props being included together with predefined values, does that overridings are really needed? Isn't something that could be provided right away from typo sets?

nicksp:
Just wanted to make a quick note here. Regarding smth. that overrides fonts and other typography related props: If it's meant to be used on admin pages and our Sketch designs don't cover that, we shouldn't worry about that for now.

src/components/SolutionForm/index.js
```diff
@@ -46,53 +50,32 @@ const Form = elem.form([
...
-      padding: .75em 2.4em
-      transition: all .3s ease-out
-      font-weight: 700
+      display: block;
+      width: 100%;
+      height: 156px;
+      resize: none;
+      border: 1px solid ${theme.lineSilver3};
+      border-radius: 4px;
+      font-size: 18px;
```
bernardodiasc: This font-size here are really needed? I see this class inherits the typo.base, which has font-size = 20px, so I see that this is a variation, but question is: isn't better to avoid these overidings in favor of all typography styles consistently defined in the typo.js?

I understand if isn't the case, just making questions here :)

src/components/Milestones.js
bernardodiasc: I double checked `typo.tinyHeading`, it seems that it doesn't provide `font-size` and `color` by default, I guess it's not an override case here. Will leave the comment anyway if we can think in a way to move that to vars, or disregard the comment if it's just fine how it is. :)

src/components/Button.js
```diff
@@ -24,28 +24,24 @@ type Props = {
...
-      transition: all .3s ease-out;
-      white-space: nowrap;
-    }
+    background: transparent;
+    border-radius: 0;
+    border: 2px solid transparent;
+    color: ${theme.baseBrighter};
+    cursor: pointer;
+    display: inline-block;
+    font-weight: 400;
```
bernardodiasc: Maybe the typography props here could be provided by typo variables right away.

src/components/Button.js
```diff
@@ -95,12 +91,7 @@ const sizeStyles = {
     font-size: .75em;
   `),
 
-  large: cmz(
-    typo.family.action,
-    `
-      font-size: 1em;
-    `
-  ),
+  large: cmz(`font-size: 1em`),
```
bernardodiasc: Be careful with relative units, if this `em` goes in a non-root element, it possibly can result the expected size on first sight. But if the `em` is being used on the root element of the component, the results can be unpredictable, because will vary accordingly the font-size of the first parent that has font-size defined, that is unknown from this component perspective.

The easy solution is to set `rem` in the root element of the component, that way the unpredictability is narrowed down to the `body` font-size. Since relative units can cause a lot trouble if not using very consciously, I'd recommend to stick to absolute units to avoid that trouble.

bernardodiasc:
@nicksp one extra suggestion. You could create a component similar to `ColorPalette` for Typography variables. It's quite simple, just get the variables and loop through including styles in blocks of texts `The quick brown fox jumps over the lazy dog`. This will provide handy visualisation of all the typography sets available. :)

src/components/SolutionForm/index.js
nicksp: This is intentional. As per Sketch design, the base text (`typo.base`) is `20px`. But form values differ from the regular text by font-size only. I thought instead of specifying each possible typography definition, that may vary from other just slightly, I'd go with overrides. But after another examination of styles, and your note here, I realized, I'll define another typo class that will be composed of the base with some overrides. This way we'll have a single place to maintain typography.

nicksp:
@bernardodiasc Created a `Typography` visualisation component, thanks for suggestion 😛 It'll be extended automatically as we add more typography sets :)

localhost_6060_

bernardodiasc:
That's awesome! :D

src/components/SolutionForm/index.js
nicksp: Fixed that by defining a new typography set `typo.formText`

src/components/Milestones.js
nicksp: Moved that over to the typo set. But, given the `Button` component assumes dynamic sizing, still keep font-sizes there to override the one coming from typo set.

src/components/Button.js
nicksp: Yea, fixed that to use absolute units as part of revising `Button` component. Thanks for spotting 👍

src/components/Button.js
nicksp: Fixed during revising the `Button` component.

src/components/ApplicantBadge.js
nicksp: Updated.

nicksp:
@bernardodiasc I revised the complete typography sets and updated accordingly :-) Also I fixed the issue with `undefined`. The ones you still can see come from the old snaps. I guess they don't get updated. Latest snaps don't contain `undefined` as part of class name.

If you got some time, could you please re-review that, so this PR doesn't block others? Just want to get this out the door asap 😝

.snapguidist/__snapshots__/Roadmap-1.snap
nicksp: Should be fixed now in new spec files.

src/components/Button.js
```diff
@@ -91,26 +84,11 @@ const colorStyles = {
 
 // Size options
 const sizeStyles = {
-  normal: cmz(`
-    font-size: .75em;
-  `),
+  small: cmz(`font-size: 10px !important`),
```
bernardodiasc: I imagine why `!important` was required here. It's some kind of trouble I had a lot with CSS Modules as well. We don't have exact control of the order that classes are printed in the output and having low specificity the latest declarations overrides the first ones.

I think there is no much do do about this to avoid the `!important` tag, so I guess that's fine.

src/components/CollapsibleSection.js
bernardodiasc: Yeah, I understand what you said, it's kinda fine this way. Let's keep an eye in how the CollapsibleSection are used, if someone includes text blocks without `<p>` that results visually wrong, then we know where the problem is. :)

src/components/Button.js
nicksp: Right, `!important` here is to ensure exact font sizes are applied over the base ones. I tried getting control over the order of classes, but couldn't find a better way to do that. Will keep that as is for now.

src/components/CollapsibleSection.js
nicksp: indeed, let's keep an eye on it :-)

PR #100, opened 2018-01-04
Fix layout for roadmap on mobile screen

piotrpasich:
image

Fixes the  https://github.com/x-team/auto/issues/787

### DIFF
--- a/src/components/ui/RoadmapHero.js
+++ b/src/components/ui/RoadmapHero.js
@@ -27,7 +27,6 @@ const Root = elem.div(cmz(`
     display: flex
     flex-wrap: wrap
     justify-content: space-around
-    align-items: center
   }
 
   @media screen and (max-width: ${breakpoints.sm}) {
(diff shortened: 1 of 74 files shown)


src/styles/typo.js
```diff
@@ -184,6 +194,8 @@ export default {
     @media screen and (max-width: ${breakpoints.sm}) {
       & {
         width: 2rem
+        margin-left: auto
```
nicksp: That's no-no. Divider should only be centered if centered prop is passed to the text component. In other cases I don't think we're able to force each instance to be centered

piotrpasich: ok

nicksp:
@piotrpasich Do you need any help mate with progressing with this one? Just let me know if so 😏

piotrpasich:
@nicksp fixed

src/components/ui/RoadmapHero.js
```diff
@@ -52,14 +62,15 @@ class RoadmapHero extends PureComponent<Props, State> {
...
+    this.updateAlignment()
   }
 
   componentWillUnmount () {
     window.removeEventListener('resize', this.updateAlignment)
   }
 
   updateAlignment = () => {
-    const mq = window.matchMedia(`(max-width: ${breakpoints.xs})`)
+    const mq = window.matchMedia(`(max-width: ${breakpoints.lg})`)
```
nicksp: this should still be `xs` as we only want to center on mobile, and keep it left aligned on other screens.

nicksp:
Updated some files. Merging now

PR #101, opened 2018-01-15
Scroll back up on Roadmap causes blinking step icon

justtry:
https://github.com/x-team/auto/issues/818

Fixing scroll logic and preventing blinking

### DIFF
--- a/src/components/ui/Roadmap.js
+++ b/src/components/ui/Roadmap.js
@@ -157,15 +157,10 @@ class Roadmap extends PureComponent<Props, State> {
     window.removeEventListener('scroll', this.throttledScrollHandler)
   }
 
-  detectScrollDirection = () => {
-    // 1 — scroll down; -1 — scroll up
-    return this.lastScrollPos > window.scrollY ? -1 : 1
-  }
-
   handleDocumentScroll = () => {
     this.levelEl.forEach((element, index) => {
       const elNode = findDOMNode(element)
-      if (isScrolledIntoView(elNode, this.detectScrollDirection())) {
+      if (isScrolledIntoView(elNode)) {
         this.setState({ activeVisible: index })
       }
     })
--- a/src/components/ui/Roadmap.js
+++ b/src/components/ui/Roadmap.js
@@ -144,11 +144,9 @@ class Roadmap extends PureComponent<Props, State> {
   }
 
   levelEl = []
-  lastScrollPos = 0
   throttledScrollHandler = () => {}
 
   componentWillMount () {
-    this.lastScrollPos = window.scrollY
     this.throttledScrollHandler = throttle(this.handleDocumentScroll, 350)
     window.addEventListener('scroll', this.throttledScrollHandler)
   }
@@ -164,7 +162,6 @@ class Roadmap extends PureComponent<Props, State> {
         this.setState({ activeVisible: index })
       }
     })
-    this.lastScrollPos = window.scrollY
   }
 
   renderRoadmap = () => {
--- a/src/components/ui/RoadmapTimelineElement.js
+++ b/src/components/ui/RoadmapTimelineElement.js
@@ -71,20 +71,21 @@ const animatedStyles = {
       stroke-miterlimit: 10
       margin: 10% auto
       box-shadow: inset 0 0 0 ${theme.baseRed}
-      animation: fill .4s ease-in-out .4s forwards, scale .4s ease-in-out .4s both
+      transition: box-shadow ease-in-out .2s
     }
-
-    @keyframes fill {
-      100% {
-        box-shadow: inset 0 0 0 30px ${theme.baseRed}
-      }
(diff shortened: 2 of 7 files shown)


src/components/ui/RoadmapTimelineElement.js
```diff
@@ -71,20 +71,21 @@ const animatedStyles = {
...
+      transition: box-shadow ease-in-out .2s
     }
-
-    @keyframes fill {
-      100% {
-        box-shadow: inset 0 0 0 30px ${theme.baseRed}
-      }
+  `),
+  checkmarkActive: cmz(`
+    & {
```
nicksp: If changes are just affect the element itself, for a shorther form, `& {}` notation can be skipped, so the final css looks like this:

```
checkmarkActive: cmz('box-shadow: ....')
```

src/utils/helpers.js
```diff
@@ -11,15 +11,14 @@ export function throttle (callback, timeout) {
...
   // Viewport offset
   const scrollPosition = window.scrollY
   const docViewBottom = scrollPosition + window.innerHeight
 
   // Element's position relative to the document
-  const elemTop = scrollPosition + top
   const elemBottom = scrollPosition + bottom
 
-  return direction < 0 ? elemTop < docViewBottom : elemBottom <= docViewBottom
+  return elemBottom <= docViewBottom
```
nicksp: right, so you got fixed it by simplifying the condition :) That's what I had in mind when first developed it. Just wanted to have different behaviour based on scroll direction. But apparently, it was buggy :-)

If you do change this way, please update a bit further. You now no longer depend on a scroll direction (`direction` prop), so you'll also need to update a calling `Roadmap` module to remove unneeded code and logic around the calculation of scroll direction, as it's not needed anymore.

Also, whenever you update `auto-ui`, make sure to run a live version (`localhost:6060`) in a browser and update failing tests, so those get committed as well 😄

src/components/ui/RoadmapTimelineElement.js
```diff
@@ -99,8 +100,8 @@ class RoadmapTimelineElement extends PureComponent<Props> {
     const cx = {
       active: isDone ? animatedStyles.timelineActive : '',
       box: animatedStyles.checkmarkBox,
-      checkmark: isDone ? animatedStyles.checkmark : '',
-      checkmarkCheck: isDone ? animatedStyles.checkmarkCheck : '',
+      checkmark: `${animatedStyles.checkmark} ${isDone ? animatedStyles.checkmarkActive : ''}`,
+      checkmarkCheck: `${animatedStyles.checkmarkCheck} ${isDone ? animatedStyles.checkmarkCheckActive : ''}`,
```
nicksp: Also, could we please simplify classes used here, seems a bit confusing.
Is that possible to have two classes as before: `checkmark` - for a completed checkmark and `checkmarkCheck` for a completed underlying check icon?

src/utils/helpers.js
justtry: thx, I think in both direction behaviour should be same - that is most natural for viewer

src/components/ui/RoadmapTimelineElement.js
justtry: hey @nicksp, I think it's good to have default styles for such elements in default state (checkmark(for the circle) and checkmarkCheck(not checked state - just for check mark inside)) and for active state (checkmarkActive and checkmarkCheckActive), so it's more clear for styles editing

nicksp: Sounds reasonable Julia :-) Let's keep it the way you proposed then 👍

nicksp:
is it ready for re-review ;-) ?

src/components/ui/RoadmapTimelineElement.js
justtry: Updated, thx

justtry:
Yes, thanks^ ready for re-review

src/components/ui/Roadmap.js
```diff
@@ -157,15 +157,10 @@ class Roadmap extends PureComponent<Props, State> {
     window.removeEventListener('scroll', this.throttledScrollHandler)
   }
 
-  detectScrollDirection = () => {
-    // 1 — scroll down; -1 — scroll up
-    return this.lastScrollPos > window.scrollY ? -1 : 1
```
nicksp: Don't forget to remove this variable `lastScrollPos` also, as it was used for the purpose of detecting scroll direction.

justtry: Haven't noticed those tests at all :sweat_smile: now totally will pay attention to it, thx for screen @nicksp !

justtry:
Updated, ready for recheck.

src/components/ui/Roadmap.js
nicksp: no worries 😃

Issue #103, opened 2018-01-18
AttachFiles component


bernardodiasc:
Create new component: **AttachFiles**

## Preview

![attach](https://user-images.githubusercontent.com/131859/35111093-902e0c16-fc61-11e7-80ae-82196bf8ac1f.png)

## Props

```
files: [
  {
    filename: string,
    path: string,
    progress: number
  }
]
onUpload: function
onCancel: function
onDelete: function
```

## Behaviour

- the "Attach a File" button opens the standard file picker dialog
- files that are uploading show a progress bar, and can be canceled (with confirmation)
- files that have finished uploading can be deleted (with confirmation)

nicksp:
CLosed via https://github.com/x-team/auto-ui/pull/106

PR #106, opened 2018-01-24
AttachFiles component

bernardodiasc:
PR for #103 

screen recording 2018-01-24 at 11 50 am


bernardodiasc:
This is ready for review!

src/components/ui/AttachFiles.js
```diff
@@ -0,0 +1,143 @@
...
+        return FileProgress({
+          style: {
+            width: `${progress}%`
+          }
+        })
+      }
+    }
+
+    const renderFiles = (files: Files) => {
+      return files && files.length > 0 && (
```
jepz20: I think we could skip validating if files exists since files will at least by an empty Array as defined in the default props

bernardodiasc:
Branch rebased, ready for review.

src/components/ui/AttachFiles.js
```diff
@@ -0,0 +1,143 @@
...
+const ButtonLabel = elem.span(cmz(`
+  padding: 0 20px
+`))
+
+class AttachFiles extends PureComponent<Props> {
+  static defaultProps = {
+    files: [],
+    onUpload: () => {},
+    onCancel: () => {},
+    onDelete: () => {}
```
zachgoldstein: The typing here for `onCancel` and `onDelete` are missing strings I think.

src/components/ui/AttachFiles.js
```diff
@@ -0,0 +1,143 @@
...
+  width: 0
+  background-color: ${theme.baseRed}
+  transition: width 0.5s
+`))
+
+const ButtonLabel = elem.span(cmz(`
+  padding: 0 20px
+`))
+
+class AttachFiles extends PureComponent<Props> {
```
zachgoldstein: Would love to understand how this is getting used in the codebase a bit better. Is it only used by `AttachFiles.md` or is there another container component somewhere?

src/components/ui/AttachFiles.js
```diff
@@ -0,0 +1,143 @@
...
+              )
+            )
+          })
+        )
+      )
+    }
+
+    return Root(
+      renderFiles(files),
+      <Button outlined onClick={onUpload}>{ButtonLabel('Attach a file')}</Button>
```
zachgoldstein: When this is clicked and while an upload is in-flight, do we want to disable this button?

src/components/ui/AttachFiles.js
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type Files = Array<File>
+
+type Props = {
+  files: Files,
+  onUpload: Function,
+  onCancel: Function,
+  onDelete: Function
+}
```
zachgoldstein: The typing here for input to `onCancel` and `onDelete` are missing strings I think.

So `onUpload` doesn't have any input params? Is it meant to send a click event? Do we want to send the upload path like `onCancel` and `onDelete`?

src/components/ui/AttachFiles.js
bernardodiasc: The business logic that will use this UI component will be located in Portfolio container https://github.com/x-team/auto/issues/829, still not existent

src/components/ui/AttachFiles.js
bernardodiasc: The onUpload button is designed to open a file picker window, the logic doesn't belong to the UI but to the container, it should not be obstructive and you should be able to select more files even when some others is in uploading progress. Does that make sense?

src/components/ui/AttachFiles.js
bernardodiasc: I'm thinking, based on your comments, that the `onUpload` is a misleading name for this function. The behaviour of this function is to be open the file picker dialog. I made the UI component without the function to let possibilities open but maybe don't need that and the dialog method could work well inside this component. The reason is that I once did a component that instead a file picker dialog, it used a modal with some other options, but might never be the case here.

zachgoldstein: Maybe it'd refine the idea of this if it started as a component in `auto`, was merged there, and then you took the core of it and abstracted something general here? Or do we already have something in `auto`?

bernardodiasc: We have nothing that uses this component on `auto` yet. This is really just the UI render piece for the attaching files functionality that will exist on Portfolio screen.

zachgoldstein:
Took another look at this and LGTM. I'm still not super clear on the usage but it'll become more clear when it gets into auto.

src/components/ui/AttachFiles.js
nicksp: That makes sense. Would result in responsive UX.

jepz20:
Hey Team, this LGTM as well. I think we should merge it so we can go ahead with task x-team/auto#829

PR #108, opened 2018-01-24
Applicant's avatar is flickering bug

### LINKED ISSUE #107: Applicant's avatar is flickering
When used in combination with buttons, `ApplicantBadge` component looks odd in a way that avatar images flicker when the buttons do appear. The fix is to set image height (we generate `64px` height gravatar images) and remove extra top padding from the component text content.

nicksp:
Fixes #107 

**Release Type:** _Fix_


falcoagustin:
LGTM 👍

PR #110, opened 2018-01-26
added-checkbox-style

### LINKED ISSUE #109: Add Checkbox Style to InputField
x-team/auto#748 uses a checkbox input. If you try to use a checkbox with InputField its renders like
captura de pantalla 2018-01-25 a la s 11 08 02 p m

jepz20:
Resolves #109 

checkbox

### DIFF
--- a/src/components/forms/InputField.js
+++ b/src/components/forms/InputField.js
@@ -57,7 +57,7 @@ const radioInputStyles = {
 
   option: cmz(`
     margin-left: 50px
-  `),
+  `)
 }
 
 const label = cmz(
@@ -191,7 +191,7 @@ class InputField extends PureComponent<Props> {
       }
     }
 
-    if (elements.hasOwnProperty(type)) {
+    if (type === 'radio' || type === 'checkbox') {
       return (
         FieldRoot(
           Tag({
--- a/src/components/forms/InputField.js
+++ b/src/components/forms/InputField.js
@@ -72,8 +72,8 @@ const FieldRoot = elem.div(cmz(`
   display: inline-block
   position: relative
 `))
-const RadioCircle = elem.span(cmz(radioInputStyles.circle))
-const RadioLabel = elem.label(labelStyles)
+const RadioCircle = elem.span(radioInputStyles.circle)
+const Label = elem.label(labelStyles)
 
 const inputStyles = [
   typo.formText,
@@ -150,7 +150,6 @@ const checkboxInputStyles = {
 }
 
 const CheckboxTick = elem.span(checkboxInputStyles.tick)
-const Checkboxlabel = elem.label(labelStyles)
 
 const getTagName = type => type === 'textarea' ? 'textarea' : 'input'
 
@@ -159,13 +158,13 @@ const inputFactory = type => elem[getTagName(type)](inputStyles)
 const specialTypesDefinitions : Object = {
   radio: {
     className: radioInputStyles.input,
-    box: RadioCircle,
-    label: RadioLabel
+    ElemBox: RadioCircle,
+    ElemLabel: Label
   },
   checkbox: {
     className: checkboxInputStyles.input,
-    box: CheckboxTick,
-    label: Checkboxlabel
+    ElemBox: CheckboxTick,
+    ElemLabel: Label
   }
 }
 
@@ -195,10 +194,11 @@ class InputField extends PureComponent<Props> {
     const errorClassName = isInvalid ? errorInput : ''
 
     if (isSpecialType(type)) {
+      const { ElemBox, ElemLabel, className } = specialTypesDefinitions[type]
       return (
         FieldRoot(
           Tag({
-            className: specialTypesDefinitions[type].className,
+            className,
             type,
             name,
             id: inputId,
@@ -207,8 +207,8 @@ class InputField extends PureComponent<Props> {
             'aria-labelledby': labelId,
             ...rest
           }),
-          specialTypesDefinitions[type].box(),
-          specialTypesDefinitions[type].label(label)
+          ElemBox(),
+          ElemLabel(label)
         )
       )
     }
--- a/src/components/forms/InputField.js
+++ b/src/components/forms/InputField.js
@@ -39,6 +39,7 @@ const radioInputStyles = {
     top: 6px
     border: 1px solid ${theme.formBorder}
     box-sizing: border-box
+    left: 0
   `),
 
   input: cmz(`
@@ -122,7 +123,6 @@ const checkboxInputStyles = {
   `),
   tick: cmz(`
     & {
-      cursor: pointer
       position: absolute
       width: 18px
       height: 18px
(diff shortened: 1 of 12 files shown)


src/components/forms/InputField.js
```diff
@@ -57,23 +57,23 @@ const radioInputStyles = {
...
+)
+
 const ComponentRoot = elem.div()
 const FieldRoot = elem.div(cmz(`
   display: inline-block
   position: relative
 `))
 const RadioCircle = elem.span(cmz(radioInputStyles.circle))
-const RadioLabel = elem.label(cmz(radioInputStyles.label))
+const RadioLabel = elem.label(cmz(label))
```
falcoagustin: Label is already a cmz, it does not need to be cmz'ed again 🙂

src/components/forms/InputField.js
```diff
@@ -107,6 +107,51 @@ const errorInput = cmz(`
...
+      border-top: none
+      border-right: none
+      transform: rotate(-45deg)
+      transition: all 0.25s ease
+    }
+  `)
+}
+
+const CheckboxTick = elem.span(checkboxInputStyles.tick)
+const Checkboxlabel = elem.label(cmz(label))
```
falcoagustin: same here

src/components/forms/InputField.js
```diff
@@ -57,23 +57,23 @@ const radioInputStyles = {
...
-  label: cmz(
-    typo.baseText,
-    `
-      margin-left: 30px
-    `
-  )
+  `)
 }
 
+const label = cmz(
```
falcoagustin: We could rename this variable labelStyles, what do you think ?

jepz20: make sense, thanks 😄

jepz20:
Hey @falcoagustin thanks for the review, I have updated the requested changes, could you check it again, please 😄

src/components/forms/InputField.js
```diff
@@ -107,6 +107,51 @@ const errorInput = cmz(`
...
+    }
+
+    &:after {
+      opacity: 0
+      content: ' '
+      position: absolute
+      width: 8px
+      height: 4px
+      top: 5px
+      left: 5px
```
falcoagustin: Can you change this to 4px? See this difference, I know its minimal, but I looked at it at the very first try,
screen shot 2018-01-29 at 4 34 32 pm
vs
screen shot 2018-01-29 at 4 34 36 pm
Can you spot the difference?

src/components/forms/InputField.js
```diff
@@ -170,15 +227,15 @@ class InputField extends PureComponent<Props> {
 
     const inputId = id || name
     const labelId = inputId ? `label-${inputId}` : ''
-    const isRadio = type === 'radio'
-    const RootComponent = isRadio ? FieldRoot : ComponentRoot
+    const isRadioOrCheckbox = (type === 'radio' || type === 'checkbox')
```
falcoagustin: We use this already in two places, what do you think about extracting this into a common function? This way you can use it when checking what type to render and underneath,
```js
const RootComponent = isRadioOrCheckbox() ? FieldRoot : ComponentRoot
```

nicksp:
Guys, sorry to disturb you. I'm late to the party and do not want to break things... but have you checked I've already got `InputField` component that supports textareas, regular inputs, radio buttons? It can easily be extended to support checkboxes. This way we'll have a generic component that can handle all the required logic in a common manner, consistently and with no duplication. Wdyt? 😄

nicksp:
Ignore my prev. comment. Late time here 😆

src/components/forms/InputField.js
jepz20: Thanks for the comment @falcoagustin, this actually made me rethink the name to a more generic one. So I change it to `isSpecialType` and is based on the `specialTypesDefinitions` object.

src/components/forms/InputField.js
jepz20: Nice catch, thanks 😄

jepz20:
Thanks again, @falcoagustin mind to give it another go 🙂

src/components/forms/InputField.js
```diff
@@ -107,10 +107,70 @@ const errorInput = cmz(`
...
+}
+
+const CheckboxTick = elem.span(checkboxInputStyles.tick)
+const Checkboxlabel = elem.label(labelStyles)
+
 const getTagName = type => type === 'textarea' ? 'textarea' : 'input'
 
 const inputFactory = type => elem[getTagName(type)](inputStyles)
 
+const specialTypesDefinitions : Object = {
```
jepz20: There is an issue with flow where if not setting specialTypesDefinitions to `Object` it throws an error that types doesn't exist even though it is already being validated in isSpecialType base on [this answer](facebook/flow#1606 (comment)), I set the specialTypesDefinitions type as object (an unsealed object) to fix the issue

piotrpasich:
What's the status of this PR ? What do we need to do to make it ready ?

nicksp:
will give it a check later today 😄

falcoagustin:
Looks good to go from my side 😄

src/components/forms/InputField.js
```diff
@@ -57,23 +57,23 @@ const radioInputStyles = {
...
+    margin-left: 30px
+  `
+)
+
 const ComponentRoot = elem.div()
 const FieldRoot = elem.div(cmz(`
   display: inline-block
   position: relative
 `))
 const RadioCircle = elem.span(cmz(radioInputStyles.circle))
```
nicksp: actually `radioInputStyles.circle` is also wrapped with `cmz` already. You can remove double wrapper from this line 😄

src/components/forms/InputField.js
```diff
@@ -107,10 +107,70 @@ const errorInput = cmz(`
...
+      border-top: none
+      border-right: none
+      transform: rotate(-45deg)
+      transition: all 0.25s ease
+    }
+  `)
+}
+
+const CheckboxTick = elem.span(checkboxInputStyles.tick)
+const Checkboxlabel = elem.label(labelStyles)
```
nicksp: this is similar to the prev. definition `const RadioLabel = elem.label(labelStyles)`. We can define generic label definition once to remove duplication, right 😃 ? Smth. like:

```js
const Label = elem.label(...)
```

src/components/forms/InputField.js
```diff
@@ -134,11 +194,11 @@ class InputField extends PureComponent<Props> {
     const labelId = inputId ? `label-${inputId}` : ''
     const errorClassName = isInvalid ? errorInput : ''
 
-    if (type === 'radio') {
+    if (isSpecialType(type)) {
```
nicksp: Let's do object destructuring after this line, to pull props off the object. This way we can condense the below render code to this:

```js
const { className, box: ElemBox, label: ElemLabel } = specialTypesDefinitions[type]
<...>
Tag({
  className,
  type,
  <...>
}),
ElemBox(),
ElemLabel(label)
```

Wdyt?

src/components/forms/InputField.js
jepz20: nice catch thanks 😄

src/components/forms/InputField.js
jepz20: you are right, makes sense. Thanks

src/components/forms/InputField.js
jepz20: this does look nicer, thanks for the recommendation 😄

jepz20:
Hey @nicksp thanks for the review 😄 .

Ready for another re-review, I did your recommendation to have the whole label clickable, since I think it indeed provides a better experience for the user.

nicksp:
Updated snapshots and we're ready for merge 😄

PR #113, opened 2018-01-26
Textarea component

### LINKED ISSUE #102: Textarea component
Create new UI component: **Textarea**

## Preview

textarea

## Props

```
placeholder: string
children: node
onChange: function
onFocus: function
onUnfocus: function
charLimit: number (default to 1000)
```

## Behaviour

- The Textarea component should be based on https://github.com/ngs/draft-js-markdown-shortcuts-plugin (need trial with that)
- When Textarea changes, trigger action from container given on `onChange`
- When Textarea gain focus, trigger action from container given on `onFocus`
- When Textarea loses focus, trigger action from container given on `onUnfocus`
- Textarea should limit the characters to 1000 (variations can be set by `charLimit`)
- The character counter should appear when textarea is focused

falcoagustin:
Closes #102 

## Impact
Creates a component that contains two tabs, one for writing markdown, and one that previews the markdown rendered. *EDIT:* From discussion on slack, and below, the idea was to integrate a wysiwyg editor. We decided to include Medium-editor library.

## For the reviewer
Regarding styles, there can be improvements, any improvement is more than suggested, the initial borders of the tabs are just as a matter of sampling, we can improve them the way we like. The component is shown below.

This initial component uses a wrapper that we can extend in a future to ensure it is more extensible, but its not the purpose of this PR.

Normal component:

screen shot 2018-02-04 at 6 22 30 pm

When focusing, a colored count of words appear, optionally passed by parameter, 1000 characters is the actual default.

screen shot 2018-02-04 at 6 22 53 pm

When selecting some words, It'll pop up a tab bar with options in order to style your text input:

screen shot 2018-02-04 at 6 23 00 pm

## Aditional notes

In order to include the `Medium-editor` library there where some adjustments I had to make, them being to use a css loader. More details go [here](https://x-team.slack.com/archives/C7KMMSD7Y/p1517445780000191)


zachgoldstein:
So in `auto` we currently have components that appear similar, `PortfolioItem` and `MarkdownEditor`, how is this different? Or is this meant to generalise those components a bit?
Current components in `auto`:
editor

src/components/ui/MarkdownTextarea.js
```diff
@@ -0,0 +1,226 @@
+import React, { PureComponent } from 'react'
```
zachgoldstein: I think we should throw a `// @flow` at the top here and build with typing from the start.

src/components/ui/MarkdownTextarea.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  onChange = ({ target }: Object) => {
+    this.setState({
+      text: target.value
+    })
+    const { onChange } = this.props
+    if (onChange) {
+      onChange(target)
+    }
```
zachgoldstein: How does `onChange` not conflict here? It's a function calling itself with a constant also named after itself?

src/components/ui/MarkdownTextarea.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  onFocus = ({ target }: Object) => {
+    this.setState({
+      shouldShowTextLength: true
+    })
+    const { onFocus } = this.props
+    if (onFocus) {
+      onFocus(target)
```
zachgoldstein: Same as above. I think it'd be more clear to change the name of that constant

src/components/ui/MarkdownTextarea.js
zachgoldstein: Ohh we're calling a passed in method here b/c it's not on `this`. I gotcha. Might be nice to change the names a bit for clarity but not a biggie.

src/components/ui/MarkdownTextarea.js
```diff
@@ -0,0 +1,226 @@
...
+        currentButton: name
+      })
+    }
+  }
+
+  render () {
+    const {
+      placeholder = 'Enter your response here.',
+      charLimit = 1000
+    } = this.props
```
zachgoldstein: Is there a reason why these constants are here in render vs at the top of the class where `WRITE_BUTTON_TEXT` is?

falcoagustin: They are properties from the component, not constants 😄 . The word const is a constant reference to the variable, that is where your confusion may come from 🙂. This means it cannot be reassigned, without meaning its value does not change, check out this:
```js
const obj = {}
obj.a = 1
// works!
obj = 3
Raises uncaught TypeError: Assignment to constant variable.
```
Another thing is that those props are injected from what you pass to the component defined. This is `<MyComponent placeholder='somePlaceholder'/>` and they are injected to the component inside. 😄

falcoagustin:
Regarding the component we have in `auto` you are completely right, lets wait for @bernardodiasc approval, but my guess is that as there is no component in auto-ui we want that mapping to it, but no certainty of this 😄. Thanks for the review!

src/components/ui/MarkdownTextarea.js
bernardodiasc: no need to stick to the issue's specs that way, suggestions for changes are always welcomed :)

src/components/ui/MarkdownTextarea.js
nicksp: I think what Zach meant is why do we have `1000` defined inline and not in the shared constants file so that it can be re-assigned easier if we want?

bernardodiasc:
To be fair, I didn't create the specs, just translated to a task. The Markdown component located on `auto` could be used but the real point of this task is to have this component in `auto-ui` with only UI logics. I believe I missed the point in checking compatibility with existing component with the specific requirements for Portfolio Screen. My suggestion to create a new one was based on a Draft.js requirements that @falcoagustin realised overcomplicated, and I agreed. Although this will make us having 2 similar (they aren't identical at this point) components in codebase, at some point would be wise to refactor any Markdown usage in `auto` to import from `auto-ui` and deprecate the old component.

nicksp:
@zachgoldstein @falcoagustin @bernardodiasc That's what I meant when I said I already had created `Markdown` component in Auto. Now to replicate similar behavior and UI we introduced one more dependency and wrote similar code. What we needed to do is just move a component from `auto` over to `auto-ui`, tweak it as required and that's it.This way we could prevent spnding dev time and re-inventing what is already done. So I'm a bit skeptical about merging this code that will eventually duplicate existing Auto component.

@joshwnj Wdyt?

bernardodiasc:
> What we needed to do is just move a component from auto over to auto-ui, tweak it as required and that's it.

Yes, my bad, should have realised that when created the task.

I just looked at https://github.com/x-team/auto/wiki/Portfolio-UI "let's trial a markdown textarea with https://github.com/ngs/draft-js-markdown-shortcuts-plugin and see whether it's going to give a good experience.", in fact @falcoagustin gave a try and realised not giving good experience.

What would be the challenge on moving and adapting exiting Markdown for #102 specs?

nicksp:
> What would be the challenge of moving and adapting existing Markdown for #102 specs?

That's what we should have investigated :-) I can't answer right now 😄  Need to take a better look, but I'm sure it won't take that much time and effort and we'll keep existing code and previous effort that we put in it.

bernardodiasc:
On a quick sight, the challenges would be:

- Recreate https://github.com/x-team/auto/tree/master/src/client/components/MarkdownEditor and https://github.com/x-team/auto/tree/master/src/client/components/Tabs on `auto-ui`;
- Adapt MarkdownEditor to Portfolio Screen specs;
- Refactor `auto` to import both MarkdownEditor and Tabs from `auto-ui`;

Both recreating UI components and refactoring application imports would require some specific efforts, for instance `isAmbassador` props should be renamed to `allowEdit`, also some styles should be moved/rewritten on auto-ui, etc. But IMHO, it's a good move to have these components on auto-ui, where they belong :)

falcoagustin:
I agree with what you say @nicksp, at least we know that the component similar to Draft.js results in a terrible user experience. We should refactor current component and map it to `auto-ui` so that we have the most possible amount of components there. Let me know what do you decide and I'll gladly take care of it 😄

src/components/ui/MarkdownTextarea.js
joshwnj: Let's go with `onBlur` as it's a familiar DOM term

falcoagustin:
@nicksp ready for re-review!

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,176 @@
...
+  margin-bottom: -1px
+  padding: 0 10px
+`)
+
+const navContainerStyles = cmz(`
+  margin-bottom: 10px
+  border-bottom: 1px solid ${theme.lineSilver3}
+`)
+
+const navButtonStyles = cmz(`
```
bernardodiasc: I'm not sure why you made this way... I think you could safely get rid of `!important` using something like this:

```css
border-left: 1px solid ${theme.lineSilver3}
border-top: 1px solid ${theme.lineSilver3}
border-right: 1px solid ${theme.lineSilver3}
border-bottom: 0
background-color: ${theme.baseBrighter}
```

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,176 @@
...
+
+  & .editable {
+    height: 100%
+    outline: none
+  }
+
+  & p {
+    margin: 0
+  }
+`)
```
bernardodiasc: More small aesthetic bits :)

When including text, if I use a H2 in the beginning, it adds too much space on the top (the same happens if the element is in the bottom), a simple solution to treat that is adding following:

```css
& :first-child {
    margin-top: 0
}


& :last-child {
    margin-bottom: 0
}
```

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,176 @@
...
+
+const Root = elem.div([
+  utilStyles.maxWidth,
+  cmz(`
+    font-family: "Source Sans Pro", "Helvetica Neue", Helvetica, Arial, sans-serif
```
bernardodiasc: It would be better to inherit typography styles from `typo.formText` here (remove redundancy).

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,176 @@
...
+  `)
+])
+
+class TextareaEditor extends PureComponent<Props> {
+  state: State = {
+    text: '',
+    shouldShowTextLength: false
+  }
+
+  _onChange = (text: string) => {
```
bernardodiasc: This is another matter of taste suggestion for syntax, the underscore is generic character that have different meanings on it's usages here and there in the codebase. I would choose something more literal to use on cases like this, for instance: `handleOnChange`, `handleOnFocus`, `handleOnBlur`.

These cases are common in React components, when you have a function that exists to **handle** another function that is usually passed by props.

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,176 @@
...
+
+    const options = {
+      placeholder: {
+        text: placeholder
+      }
+    }
+
+    return Root(
+      <div className={navContainerStyles}>
+        <nav className={navStyles}>
```
bernardodiasc: I like this tab part, looks pretty but this one have no functionality. I think it could be removed. :)

src/components/ui/TextareaEditor/TextareaEditor.md
```diff
@@ -0,0 +1,17 @@
+Normal textarea example :
```
bernardodiasc: The demos are inverted :)

bernardodiasc:
Code review is quite simple, the code looks overall very good. The user experience have a problem I identified, please take a look at the Gif bellow:

screen recording 2018-02-07 at 01 45 am

Explaining: the problem happens when I switch focus from one textarea to another, the current implementation forces me to click twice, doesn't look much natural behaviour.
  - more detailed in the gif, it lost focus when I click anywhere outside, that's correct
  - it doesn't lost focus when I click on another textarea, but the log seems to expose that the other textarea was focused, but it's not, if I type something it affect the first textarea

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    this.medium = new _MediumEditor('.editable', this.props.options)
+    this.medium.subscribe('editableInput', e => {
+      const { text, charLimit } = this.props
+      if (this.input.textContent.length > charLimit) {
+        this.input.textContent = text
+      }
+
+      let { textContent } = this.input
```
nicksp: Can this be changed to `const` ?

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,56 @@
...
+
+  componentDidMount = () => {
+    const subscribeFunction = fun => event => {
+      if (fun) {
+        fun(this.input)
+      }
+    }
+
+    this.medium = new _MediumEditor('.editable', this.props.options)
+    this.medium.subscribe('editableInput', e => {
```
nicksp: I reckon with `this.medium.destroy()` we remove all the assigned event listeners, right? Just to make sure we keep memory in order after component unmounts.

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,56 @@
+import React, { PureComponent } from 'react'
+
+import 'medium-editor/dist/css/medium-editor.css'
+import 'medium-editor/dist/css/themes/default.css'
+
+const _MediumEditor = require('medium-editor')
+
+class MediumEditorWrapper extends PureComponent {
```
nicksp: Let's add some props validation/typing to the component ;-) ?

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,56 @@
...
+      }
+
+      let { textContent } = this.input
+      const { onChange } = this.props
+      if (onChange) {
+        onChange(textContent)
+      }
+    })
+
+    this.medium.subscribe('focus', subscribeFunction(this.props.onFocus))
```
nicksp: We could pull `onFocus` and `onBlur` off the `this.props` object above, right?

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,56 @@
...
+    this.medium.restoreSelection()
+  }
+
+  componentWillUnmount = () => {
+    this.medium.destroy()
+  }
+
+  componentDidMount = () => {
+    const subscribeFunction = fun => event => {
+      if (fun) {
```
nicksp: Can `fun` be undefined in this case? Would it make sense to subscribe without passing a listener? In our case, we pass it all the time, so might make sense to remove the `if` check. Typing would also help catching possible errors. Wdyt?

src/components/ui/TextareaEditor/TextareaEditor.js
nicksp: Good point 👍

src/components/ui/TextareaEditor/TextareaEditor.js
nicksp: I also tend to follow similar pattern: have event-like props be named like `onEventName`, and respective handlers like `handleEventName` (without `onPrefix` as we're handling an event 😄 )

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,176 @@
...
+    this.setState(() => ({ shouldShowTextLength: false }))
+    const { onBlur } = this.props
+    if (onBlur) {
+      onBlur(target)
+    }
+  }
+
+  render () {
+    const {
+      placeholder = 'Enter your response here.',
```
nicksp: Seems like it should be responsibility of another component.

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,56 @@
+import React, { PureComponent } from 'react'
+
+import 'medium-editor/dist/css/medium-editor.css'
+import 'medium-editor/dist/css/themes/default.css'
+
+const _MediumEditor = require('medium-editor')
```
nicksp: Let's call it simply `MediumEditor` as it's just a normal imported module?

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,56 @@
...
+    const tag = 'div'
+
+    const childProps = {
+      ref: node => { this.input = node },
+      className: 'editable'
+    }
+    if (this.medium) {
+      this.medium.saveSelection()
+    }
+    return React.createElement(tag, childProps)
```
nicksp: Why did we decide not to use existing `elem.div()` function 😄 ?

src/components/ui/TextareaEditor/TextareaEditor.js
bernardodiasc: yeah yeah, `handleChange`, `handleFocus` and `handleBlur`, I mistyped :)

bernardodiasc:
Hey @nicksp, I have some comments/questions about your some of your comments 😄 

> 3. When you click outside textarea, the counter is gone and I'm confused where it is 😛 . Is this a feature or a bug? I'd suggest looking around for real-world examples and try to find a similar pattern. Twitter compose area is a quite popular example which is familiar to millions of users. It always shows a counter, and I'd expect to have similar behavior. Wdyt guys?

As per spec: "The character counter should appear when textarea is focused", but there's nothing about vanishing when blur. I'd say to let counter visible if `chars > 0`. 

> 4. Why we hardcoded a counter? Was it by design? That limits the usage of the component in case we want to inject similar component without having a counter at all? That would be treated as thinking forward

The counter is part of the component's spec. I think if a use cases of the component without the counter appears, we should include a flag prop to ignore max chars and keep counter hidden.

> 5. Also, as a general suggestion, I'd propose having a composable generic TextArea component component which can be extended to render markdown (or potentially anything else in the future) and whether to show a counter, as this it an extra option for the textarea and might not be required in some cases.

I'm not sure I get this... Why such abstractions? What would be use cases? Are you talking about a generic `<TextArea />` that wrap and mimic the native HTML `<textarea>`?

---

About the other points, I agree with them.

src/components/ui/TextareaEditor/TextareaEditor.js
bernardodiasc: I think it makes sense to have `placeholder` as a prop with some defaultProps value defined.

nicksp: Right, as well as any other common user-defined props like `maxLength`, `alt`, `disabled`, `autofocus`, whatever else is possible. No need to limit the user to choose how to config textarea. Let them pass it if they wish, smth like `...rest` taken off the props might work here 😄

nicksp:
> As per spec: "The character counter should appear when textarea is focused", but there's nothing about vanishing when blur. I'd say to let counter visible if chars > 0.

Either like that or as I suggested, to make this U pattern familiar for users.

> The counter is part of the component's spec.

Alright, makes sense. But we're free to suggest improvements and think forward a bit 😄 

> I'm not sure I get this... Why such abstractions?

It's not an abstraction. It's an idea for a component which can accept a render prop to render based on the option. Right now we have markdown renderer, but can easily be extended to accept other custom renders. That's not much work as it seems but can open up more opportunities for future even if we don't have such requirements right now.

src/components/ui/TextareaEditor/TextareaEditor.js
bernardodiasc: Agreed 👍

bernardodiasc:
> The counter is part of the component's spec.
> Alright, makes sense. But we're free to suggest improvements and think forward a bit 😄

Sure thing! I'm just trying to understand the use cases and what changes in the specs. And if your suggestion is simply a matter of not having the counter, if a flag prop wouldn't fit the case. :)

falcoagustin:
Thanks for both of your comments @nicksp and @bernardodiasc. I will tackle this after I'm done with the e-mail templates. 😄

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,176 @@
...
+  padding: 8px 12px
+  outline: none
+  cursor: pointer
+`)
+
+const editorContainerStyles = cmz(`
+  & {
+    display: block
+    width: 100%
+    height: 156px
```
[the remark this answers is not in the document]
falcoagustin: Hey @bernardodiasc, this was this way in order for the component not to extend and have a scrolling in order to control the size of it not to grow with every `<p>` added. Do you suggest anything else?

falcoagustin:
Hey @bernardodiasc and @nicksp, all of the stuff you mentioned are done, except for the charLimit implementation tidying. The 👍 in your comments means that it has been taking care of.
If someone could pick it up from here as I am now going to another project I'd appreciate it very much.
Thanks!

zachgoldstein:
I've fixed the build here and added flow typing. I think this is now good to go. Would love a pair of eyes on it to finalise.

@jepz20 sorry, just noticed you self assigned this a few mins ago.

jepz20:
Awesome @zachgoldstein 😄  although we still need to find a fix for the char limit issue,(if you hit the char limit it sends the cursor to the beginning and loses the styles) which is what I'm working on right now

jepz20:
The issue with the charLimit was more complicated than expected.
Had to take into account when:

### Typing a new char
 straightforward just need to listen to the input event and stop if the limit has been reached
### Pasting text
To handle the paste event and take in consideration the charLimit it was necessary to extend the `paste extension` provided by `medium-editor`

There are two types of paste:
   #### As plain text  
   straightforward just need to cap the plain text based on the current count and the char limit in the `doPaste` method
   #### With HTML 
   This is more complicated. We need to cap the text limit base not on the char count of the HTML but in the actual text being displayed. 

To achieve it I created a method that traverses the HTML counting the characters being displayed and when the char limit has been reached it "crops" the HTML to that point.

Because medium-editor does some parsing in the `pasteHTML` method 
and this is the last step before finally pasting the HTML, I had to copy all of the logic from 
`https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/paste.js` and add the crop logic
 
This option provides a better user experience, for example, you can copy text from one textarea to another without losing the style

copy with style

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,146 @@
...
+        if (workEl.nodeName.toLowerCase() && this.getEditorOption('targetBlank') === 'a') {
+          MediumEditor.util.setTargetBlank(workEl)
+        }
+
+        MediumEditor.util.cleanupAttrs(workEl, options.cleanAttrs)
+        MediumEditor.util.cleanupTags(workEl, options.cleanTags)
+        MediumEditor.util.unwrapTags(workEl, options.unwrapTags)
+      }
+      const selectionCount = MediumEditor.selection.getSelectionRange(this.document).toString().length
+      const cropFragment = getDelimitedHTML(fragmentBody, charLimit + selectionCount, currentText.length)
```
jepz20: copy and adapted to our style the logic from 
`https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/paste.js` 
to add this line and show the cropFragement, it was not possible to extend it like on the `doPaste` method because this is the last step which calls directly Medium util

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,146 @@
...
+        workEl = elList[i]
+        if (workEl.nodeName.toLowerCase() && this.getEditorOption('targetBlank') === 'a') {
+          MediumEditor.util.setTargetBlank(workEl)
+        }
+
+        MediumEditor.util.cleanupAttrs(workEl, options.cleanAttrs)
+        MediumEditor.util.cleanupTags(workEl, options.cleanTags)
+        MediumEditor.util.unwrapTags(workEl, options.unwrapTags)
+      }
+      const selectionCount = MediumEditor.selection.getSelectionRange(this.document).toString().length
```
jepz20: take into consideration the currently selected text, since this will be replaced with the pasted text

jepz20:
@zachgoldstein @nicksp this should be ready to be review again 😄

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,153 @@
...
+  handleBlur = (target: HTMLTextAreaElement) => {
+    const { onBlur } = this.props
+    if (onBlur) {
+      onBlur(target)
+    }
+  }
+
+  render () {
+    const {
+      charLimit = 1000,
```
nicksp: I think it's better set it as a defaultProp.

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,153 @@
...
+            text={text}
+            html={html}
+            charLimit={charLimit}
+            options={options}
+            onChange={this.handleChange}
+            onFocus={this.handleFocus}
+            onBlur={this.handleBlur} />
+        </div>
+        <div className={textCountStyles}>
+          {shouldShowTextLength && <p>
```
nicksp: Shouldn't we hide it on blur? cc/ @joshwnj Do we still want the counter to stay in view once showed?

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,149 @@
...
+      const PasteWithCharLimitExtension = pasteWithCharLimitExtension(charLimit)
+      defaultOptions.extensions = {
+        paste: new PasteWithCharLimitExtension()
+      }
+    }
+    this.medium = new MediumEditor('.editable', { ...defaultOptions, ...options })
+    this.medium.setContent(html || text)
+
+    if (charLimit != null) {
+      this.medium.on(this.input, 'keypress', (event) => {
```
nicksp: We can provide [a type](https://flow.org/en/docs/react/events/) for `event` object here, otherwise, parenthesis can be omitted.

src/components/ui/TextareaEditor/MediumEditorWrapper.js
```diff
@@ -0,0 +1,149 @@
...
+    const nodeCopy = node.cloneNode(false)
+    parent.appendChild(nodeCopy)
+    if (node.childNodes.length === 0) {
+      let content = node.textContent
+      nodeCopy.textContent = getCropText(content, charLimit, textCount)
+      textCount += content.length
+      return
+    }
+
+    Array.prototype.forEach.call(node.childNodes, it => getClone(nodeCopy, it))
```
nicksp: Can we just write `node.childNodes.forEach(...)` for a shorter call 😄 ?

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,153 @@
...
+    const { onChange } = this.props
+    if (onChange) {
+      onChange(text, html)
+    }
+  }
+
+  handleFocus = (target: HTMLTextAreaElement) => {
+    this.setState(() => ({ shouldShowTextLength: true }))
+    const { onFocus } = this.props
+    if (onFocus) {
```
nicksp: I'd prefer checking possible null functions with a strict check: `if (typeof onFocus === 'function') {...}`

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,153 @@
...
+  state = {
+    text: '',
+    html: '',
+    shouldShowTextLength: false
+  }
+
+  handleChange = (text: string, html: string) => {
+    this.setState(() => ({ text }))
+    const { onChange } = this.props
+    if (onChange) {
```
nicksp: I'd prefer checking possible null functions with a strict check: if (typeof onChange === 'function') {...}

src/components/ui/TextareaEditor/TextareaEditor.js
jepz20: Hey @nicksp I think this may be a decision made by the user when calling the component so I added a prop called `hideTextLengthOnBlur`.

src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  state = {
+    text: '',
+    html: '',
+    shouldShowTextLength: false
+  }
+
+  changeShouldShowTextLength = (val: boolean) => {
+    setTimeout(() => this.setState(() => ({ shouldShowTextLength: val })))
```
jepz20: I added a setTimeout because when setting the state directly on the `onBlur` or `onFocus`  events for some reason medium triggers them again and you have to double-click the element to gain focus.

jepz20:
Hey @nicksp thanks, fix the last issue, change it to have the same default id when no id is provided (the previous implementation change base on time) and works fine now, please have another look 😄

src/components/ui/TextareaEditor/TextareaEditor.js
joshwnj: @nicksp @jepz20 yeah I think we only need to see the counter when the textarea is in focus. However one thing we want to avoid is causing the layout to jump.

So perhaps hide the counter with css (eg. `visibility: hidden`) rather than removing it altogether?

nicksp: @joshwnj Made a commit to make counter visible on focus only and removed extra visibility prop https://github.com/x-team/auto-ui/commit/acd56e77378cebe2498d3f5527c5b90abd1f8274

Until we need this explicitly, no need to overcomplicate the component with additional props to show/hide counter text.

joshwnj: Sounds good 👍

nicksp: I reverted that and made it explicit to only show counter on focus. No need to introduce booelan flags to add functionality we're not asked for...

PR #117, opened 2018-02-07
Add datatest attributes

danturchenko:
Most data-test attributes are being added in auto, but some components could only have their root elements accessed here. This branch adds friendly readable data-test selectors to some elements important for test automation to clean up the css selectors and improve test reliability


joshwnj:
LGTM

PR #124, opened 2018-02-21
Make InputField labels inherit display property

zachgoldstein:
Closes: https://github.com/x-team/auto-ui/issues/123
This is causing availability fields to cut off on mobile in auto: https://github.com/x-team/auto/issues/916

Before:
screen shot 2018-02-21 at 12 29 40 pm

After:
screen shot 2018-02-21 at 12 32 20 pm


nicksp:
@zachgoldstein When we create `auto-ui` PR, we also update changed snapshots and build app with `npm run build` command. 😄

PR #126, opened 2018-03-28
add id prop to AttachFile Component

### LINKED ISSUE #125: Change attachment file to support a file id
AttachFile component returns the path and name of the file as `path/file`, however, the path is not accessible via the browser. 
Add support to send an id as a unique identifier that could be used on the onCancel and onDelete callbacks.

jepz20:
Fixes #125 

add an id prop and return it on the onCancel and onDelete callbacks


joshwnj:
LGTM

Issue #132, opened 2018-05-04
Add toggle switch component


nicksp:
This is to accompany https://github.com/x-team/auto/issues/965 feature

So the task assumes creating a customized version (maybe by setting an additional prop) of a checkbox component (`/src/components/forms/InputField.js`)

The design is as follow (can be created with pure css3 features):

![image](https://user-images.githubusercontent.com/579331/39612102-0bd02e20-4f65-11e8-81e6-26d1447099cd.png)

The complete design Sketch file for this feature can be found here: https://cl.ly/2o112I2o3X3b

nicksp:
@rpunkfu I guess this is ready for review ;-) Don't forget to add labels to issues as well 😄

PR #133, opened 2018-05-04
Add toggle switch component

rpunkfu:
Resolves: #132

This toggle switch component is implement as variation over simple `checkbox` input.
I added definitions object for these types variation, so `sliding-checkbox` type becomes `checkbox` in the DOM, but we can style it differently and use slightly different syntax.
This was particularly useful for this component, where I wanted underlying logic to be the same as if it was a checkbox, but wanted some complicated styling.

#### Video with demo
https://cl.ly/360b3J0A2f0M/Screen%20Recording%202018-05-04%20at%2003.34%20PM.mov

### DIFF
--- a/.snapguidist/__snapshots__/InputField-11.snap
+++ b/.snapguidist/__snapshots__/InputField-11.snap
@@ -1,8 +1,21 @@
 exports[`InputField-11 1`] = `
 <div>
-  <input
-    aria-labelledby=""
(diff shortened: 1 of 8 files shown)


nicksp:
Just waiting on @jepz20's approval before merging back to `master` :)

jepz20:
LGTM 🎉

PR #139, opened 2018-05-18
Add avatar component

slawekkolodziej:
PR for issue x-team/auto-ui#137

### DIFF
--- a/src/components/ui/Avatar.js
+++ b/src/components/ui/Avatar.js
@@ -1,57 +1,44 @@
+// @flow
+
 import { PureComponent } from 'react'
 
 import elem from '../../utils/elem'
 
-type Props = {
-  src: string,
-  alt: ?string
-}
-
 const cmz = require('cmz')
 
-// Size options
-const sizeStyles = {
-  small: cmz(`
-    width: 32px
-    height: 32px
-  `),
-  normal: cmz(`
-    width: 64px
-    height: 64px
-  `)
+type Props = {
+  alt?: string,
+  src: string,
+  size: number
 }
 
 const Root = elem.div(cmz(`
-  border-radius: 50%
   display: inline-block
+  border-radius: 50%
   overflow: hidden
 `))
 
 const Image = elem.img(cmz(`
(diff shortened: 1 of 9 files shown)


src/components/ui/Avatar.js
```diff
@@ -0,0 +1,60 @@
+import { PureComponent } from 'react'
```
nicksp: Don't forget to type annotate all ya components as we do with the rest of them. Also, add `// @flow` annotation at the top to allow typecking that file.

src/components/ui/Avatar.js
```diff
@@ -0,0 +1,60 @@
...
+`))
+
+const Image = elem.img(cmz(`
+  width: 100%
+  height: 100%
+`))
+
+class Avatar extends PureComponent<Props> {
+  static defaultProps = {
+    size: 'normal',
```
nicksp: You use `size` here but forgot to add that to Props.

src/components/ui/Avatar.js
```diff
@@ -0,0 +1,60 @@
...
+const cmz = require('cmz')
+
+// Size options
+const sizeStyles = {
+  small: cmz(`
+    width: 32px
+    height: 32px
+  `),
+  normal: cmz(`
+    width: 64px
```
nicksp: since you duplicate width and height, it's much more flexible to pass size number and just use that. Simple enough to allow any possible sizes without introducing more size options 😏

src/components/ui/Avatar.js
```diff
@@ -0,0 +1,60 @@
...
+
+const Root = elem.div(cmz(`
+  border-radius: 50%
+  display: inline-block
+  overflow: hidden
+`))
+
+const Image = elem.img(cmz(`
+  width: 100%
+  height: 100%
```
nicksp: Just setting width and height to 100% isn't enough and would break image constraints when parent's div sizes are different and not equal each other. The more robust way is to make contained images responsive with remaining proportions.

PR #141, opened 2018-05-22
Add Note component

### LINKED ISSUE #127: Note component
Mockup:

image

A `Note` component has:

- Avatar
- Full name
- Date (prop should be a `Date` object, and we will render as a relative time, eg. _"1 h ago"_ or _"10 m ago"_
- Body (render as markdown)

## Behaviour

- if the Date is less than 1 minute ago, it should always be rendered as _"just now"_
- if the Date is more than 24 hours ago, it should be rendered as "dd mmm  yy"
- the Date should be periodically re-rendered, so that as time passes the correct relative time is shown.

bernardodiasc:
Relates to the part of #127 

Please run `npm install` for new dependencies before run on this branch.

## Preview

screen shot 2018-05-24 at 2 39 10 pm

---

Thanks @skvale for the great work. The code was really good and polished, I just included the existing Avatar component in place.

### DIFF
--- a/src/components/ui/Text.js
+++ b/src/components/ui/Text.js
@@ -5,6 +5,7 @@ import { PureComponent } from 'react'
 import elem from '../../utils/elem'
 
 import typo from '../../styles/typo'
+import theme from '../../styles/theme'
 
 import type { Element } from 'react'
 
@@ -43,7 +44,19 @@ const Content = elem.div(cmz(
   'margin: 15px 0'
 ))
 
-const PureContent = elem.span(cmz(typo.baseText))
+const PureContent = elem.span(cmz(
+  typo.baseText,
+  `
+    a {
+      color: ${theme.baseRed}
+      text-decoration: none
+    }
+
+    a:hover {
+      text-decoration: underline
+    }
+  `
+))
 
 const centerAlign = cmz(`text-align: center`)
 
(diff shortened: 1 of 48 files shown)


nicksp:
I agree that the original code was of a really good quality.

nicksp:
@skvale You're right. We'll fix that. We're working on that actually. Thanks for the contribution!

nicksp:
@bernardodiasc As discussed, please investigate the possibility to re-use existing `MarkdownEditor` component instead of introducing a new dependency.

bernardodiasc:
I believe no action is required, after taking a look at this possibility I concluded that the `medium-editor` is not purposed to be used as a markdown renderer. On other hand `markdown-to-jsx` is one of the best (if not the best) lib to deal with this, introducing this dependency can be valuable if we need more powerful text rendering options in the future.

src/components/ui/Text.js
```diff
@@ -43,7 +44,19 @@ const Content = elem.div(cmz(
...
+const PureContent = elem.span(cmz(
+  typo.baseText,
+  `
+    a {
+      color: ${theme.baseRed}
+      text-decoration: none
+    }
+
+    a:hover {
+      text-decoration: underline
```
nicksp: By introducing this global change you broke UI of some of the admin pages where PureComponent is used. On hover, links became underlined with red where it shouldn't. Be careful next time introducing global changes for existing components! Reverted that change in `master`.

PR #142, opened 2018-05-23
Include NotesFeed component

bernardodiasc:
screen recording 2018-05-24 at 03 38 pm


bernardodiasc:
This PR is ready for Review. 👍

src/components/ui/NotesFeed.js
```diff
@@ -0,0 +1,75 @@
...
+          )
+        ),
+        this.showViewMore((notes && notes.length) || 0) && (
+          <Button
+            outlined
+            block
+            color='silver'
+            onClick={this.viewMore}
+            className={buttonClass}
+          >
```
slawekkolodziej: Let's add `type='button'` to prevent submitting forms

bernardodiasc:
@slawekkolodziej good call, I updated the code with your suggestion 👍

src/components/ui/Loader.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import { PureComponent } from 'react'
+import elem from '../../utils/elem'
+
+type Props = {}
```
nicksp: No need to define empty Props. Besides such simple component as Loader is better to write in a functional form.

PR #146, opened 2018-05-29
Include read only mode on TextareaEditor

bernardodiasc:
Resolve x-team/auto-ui#143

screen shot 2018-05-29 at 11 32 12 am


bernardodiasc:
Thanks for including the build, forgot that :)

PR #147, opened 2018-05-29
Extend TextareaEditor with a size option

Pr0x1m4:
This PR adds large prop to `TextEditorArea` to facilitate height change but also includes default height


nicksp:
@bernardodiasc Re: "this component don't use textarea element, it's actually a div with contenteditable attribute."

I can see that now. What I meant to say, is that we should ideally be doing better investigations by checking all the possible scenarios with things like I mentioned above. If that's not possible we could mention that as one of the options investigated and move to the next option 😄 And I totally agree with your comments here regarding changing the prop to  be a boolean one and using a class instead of inline styles.

cc/ @Pr0x1m4

Pr0x1m4:
@nicksp @bernardodiasc I agree with changing `large` to be a boolean, makes more sense, I'll make the changes

nicksp:
Also I changed the PR title to a proper one.

PR #151, opened 2018-06-07
Create Dropdown component

bernardodiasc:
This is very simple component that can be reused with different icons and labels and content.

The preview bellow seems buggy, but it's correct. The reason that the content appears floating and with transparent background is because this component is not responsible for the styles of it's children.

screen recording 2018-06-07 at 04 24 pm


nicksp:
@bernardodiasc Do we expect that childs could have unique backgronds? Shouldn't we define some base styles for the inner content in that component, so when using it we just define childs and they already look fine?

bernardodiasc:
I'm not sure it's this component responsibility to assume the styles of another component that will appear in the floating space. It actually takes care of the position, it's a transparent place for children appear, we can deal with visual of the children.

I'm trying to think about how the styles of that transparent box could be, but background, margins, paddings, borders, colours in general, all seems to belong to the children in the way I see. Let me know if you have some use cases to add default styles, I can make the changes based on that. :)

src/components/ui/Dropdown.js
```diff
@@ -0,0 +1,90 @@
...
+  state = {
+    open: false
+  }
+
+  toggle = () => this.setState((prevState: State) => ({ open: !prevState.open }))
+
+  render () {
+    const { icon, label, children } = this.props
+    const { open } = this.state
+    return (label && children) ? (
```
slawekkolodziej: How about we let dropdown to have icon only? That way I can simply reuse this component here: https://github.com/x-team/auto/issues/1122

slawekkolodziej: cc @bernardodiasc

nicksp:
I am also thinking given the case of x-team/auto#1122 if we should show up/down icon based on a prop only, so we can have other options to reveal hidden content.

bernardodiasc:
hey @slawekkolodziej and @nicksp please let me know if this matches with what's desired:

screen recording 2018-06-08 at 11 52 am

slawekkolodziej:
@bernardodiasc, looks good to me

nicksp:
Perfect 👍 And positioning will be specified by the child component, right? A parent just have `position: relative` mainly, correct?

nicksp:
@bernardodiasc I set the label back to `in progress` to indicate you're still working on it :)

bernardodiasc:
@nicksp yeah, but I'm accepting suggestions about how we'll use the prop. I used `position="right"` and currently is the only variation regarding positioning, the default is aligned to left. I didn't wanted to over complicate this without more use cases. Suggestions welcomed.

nicksp:
`left`/`right` seems correct to me, at least it cover most of the use cases we have atm. The only other thing that I'm thinking of is the way you render up/down buttons. Feels a bit smart. I'd prefer a plain approach of having an extra prop or smth specifying we actually need those icons.

bernardodiasc:
I see, yeah I considered the up/down to be part of the label because I can't imagine using it without that. But we can include a boolean prop `withArrows`, default to false. Wdyt?

nicksp:
I can think of use cases where we'd want to have a clickable label without any indications. I mean that' very possible. So to be clear with component's API, let's have a prop named like `visibilityIndicator` or smth?

bernardodiasc:
Ok, one last question about behaviour, when opened, by clicking outside the dropdown it should close? This require an extra script, I have the snippet made somewhere already.

nicksp:
I’d expect it to be closed when clicked outside.It just   feels natural. We have this functionality in Auto. We use react-click-outside lib from TJ in there. So we can reuse the sane HOC in here as well. It’s a tiny lib we can pull from npm 



Sent from myMail for iOS


Friday, 8 June 2018, 18:04 +0300 from notifications@github.com  <notifications@github.com>:

bernardodiasc:
Cool, I'm going to use the lib you suggested. Just for the record I found the snippet:

```
  document.body.addEventListener('click', function(event) {
    var hasParent = false;
    for (var node = event.target; node != document.body; node = node.parentNode) {
      if (node === $menu) {
        hasParent = true;
        break;
      }
    }
    if (!hasParent) {
      $menu.classList.remove('menu-guide__menubar--visible');
    }
  });
```

nicksp:
@bernardodiasc Cool, that's basically a base for that lib 😄 Thanks for sharing. But I tend to believe it's good to use similar solutions for same problems across all the repos of an entire app.

bernardodiasc:
@nicksp I totally agree, thanks for letting me know that lib is in use, it was much faster to use anyway.

This PR is ready for review again. /cc @slawekkolodziej 

I included a `padded` prop too, because the click area of the link can be too small in the hamburger case, it's something I'm afraid to result bad experience on mobile or even with cursor, but for regular cases that padding is not required.

## Preview

screen recording 2018-06-08 at 01 07 pm

bernardodiasc:
Included fading effect in the component:

screen recording 2018-06-09 at 12 57 pm

nicksp:
Looks great Bernardo :)



Sent from myMail for iOS


Saturday, 9 June 2018, 19:07 +0300 from notifications@github.com  <notifications@github.com>:

bernardodiasc:
This is strange, it seems a bug on `react-click-outside` side. Will fix it and send PR to the lib.

nicksp:
@bernardodiasc Why was this closed 😱 ? Could you find a root cause for the issue I mentioned?

nicksp:
PR should be re-opened.

bernardodiasc:
@nicksp I closed this because all the commits are on the #152, the Dropdown component is a requirement for the SelectBox, I'm working on the fix will send one more commit in the other branch. The rebasing was causing lots of conflicts, I tho this was a good idea.

nicksp:
@bernardodiasc Ah, so you merged two things together? Alright then. Closing this again.

nicksp:
RE: the issue, I just wonder if it fails only inside the Styleguidist playground? Have you tried that on Auto? We use similar lib on Auto for collapsible textarea and things seem working fine.

bernardodiasc:
Good suggestion, will try that.

nicksp:
Yea, it's just smth I learned :) I run into several cases when things that were behaving incorrectly inside Stylegudist, run properly when inside Auto 😄 Not saying this is the case, but worth trying that.

nicksp:
Hey @bernardodiasc Have you tried that? What's the outcome?

Issue #148, opened 2018-06-05
Allow supporting html tags on Text component


richa:
### Notes
In the issue https://github.com/x-team/auto/issues/1094, we wanted to add a Text component which should look like the screenshot below:

https://user-images.githubusercontent.com/579331/40840069-cd96dfa0-65ad-11e8-9409-701d8ee65549.png

Since the current `Text` component doesn't support html tags support, we could not add `<b>` or `<strong>`

### User Story

As a Developer, I should be able to add html tags support for Text components such that I can directly add html tags to `Text` component when used in the `auto` repo.

bernardodiasc:
Hey here, just a heads up about the styles of HTML tags. We had a situation where I included styles for `<a>` that affected unexpected places. I believe this can be the moment to include one new sample in the `Text.md` covering the most common HTML tags in the body content. :)

nicksp:
@bernardodiasc This isn't exactly styles as per se, but mostly supporting regular HTML styling tags with its default styles. So, we're unlikely to break anything. We just want to make sure that if elements like `<b>` or `<i>` are being used, they're rendered as they should by the browser. We can even support just a limited set of attributes for now. So we should be fine with this change.

nicksp:
In the prev case that you mentioned things were broken due to custom styles. And `<a>` won't be in the list of supported tags for now.

PR #153, opened 2018-06-12
Text: Support html with whitelisted tags

skvale:
Fixes: #148


nicksp:
@skvale Thanks for the fix, but we have a defined dev cycle. First, we define requirements and delegate to the dev team. So, we'll work on that when it's planned work.

nicksp:
Closed as we didn't plan work on the ticket yet.

PR #154, opened 2018-06-15
View attachments links in notes

richa:
### Screenshots
screen shot 2018-06-15 at 5 53 05 pm

### `npm run test`
screen shot 2018-06-15 at 5 55 18 pm


nicksp:
Reloved merge conflict.

src/components/ui/FileLinks.js
```diff
@@ -0,0 +1,40 @@
+// @flow
```
nicksp: This files can be removed now along with their docs, as they aren't in use anymore

src/components/ui/FileLink.js
```diff
@@ -0,0 +1,49 @@
+// @flow
```
nicksp: This file can be removed now along with their docs, as they aren't in use anymore

PR #157, opened 2018-06-28
Fix click-outside module export

keremgocen:
I wanted to reuse the Dropdown component from here but failed with flow errors during `auto` integration.

Changing how `'react-click-outside'` is exported (same as in `auto`) allowed me to render Dropdown component properly.

This component was introduced a few days ago and I'm not sure if it works out of the box. Please see https://github.com/x-team/auto/pull/1260 @nicksp @bernardodiasc


nicksp:
Hey @keremgocen I'll check things and reply back.

keremgocen:
I used the `Link` component in auto instead of creating a new one here. And created a `DropdownMenu` component in auto so this PR will be sufficient to use the `Dropdown` component in auto repo.

nicksp:
I actually took this ticket as a chance to introduce proper Flow types definitions by using a recommended `flow-typed install lib_name@version` or if there's no official definition exists `flow-typed create-stub lib_name@version`. Also fixed lint errors along the way.

PR #166, opened 2018-07-05
Create new list

justtry:
SelectBox changed to simpler version for redux usage.

Removed functions as we use redux actions for all api things and present code is not working with redux

### DIFF
--- a/src/components/ui/SelectBox.js
+++ b/src/components/ui/SelectBox.js
@@ -361,7 +361,20 @@ class SelectBox extends Component<Props, State> {
     const { onCreateNew } = this.props
     const { search } = this.state
 
-    onCreateNew && onCreateNew(search)
+    if (onCreateNew) {
+      const pr = onCreateNew(search)
+
+      // checking if pr is a Promise
+      if (pr && typeof pr.then === 'function') {
+        this.setState(() => ({ creating: true }))
+        Promise.resolve(pr).then((reponseItem) => {
+          this.setState(() => ({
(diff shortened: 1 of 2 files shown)


src/components/ui/SelectBox.js
```diff
@@ -266,18 +267,26 @@ class SelectBox extends Component<Props, State> {
...
-      }))
+      this.setState((prevState, props) => {
+        const newState = {
+          ...prevState,
+          items: viewItems,
+          view: viewItems,
+          expanded: this.props.expanded
+        }
+
+        if (typeof props.creating !== 'undefined') {
```
keremgocen: just out of curiosity, is there any chance `props.creating` can be undefined? Maybe define it in `defaultProps` so it's either true or false rather than undefined, is it possible?

src/components/ui/SelectBox.js
```diff
@@ -359,16 +368,21 @@ class SelectBox extends Component<Props, State> {
...
-        this.handleSearch('')
-      }) // to do: catch state
+    const { search } = this.state
+
+    if (onCreateNew) {
+      const pr = onCreateNew(search)
+
+      // checking if pr is a Promise
+      if (pr && typeof pr.then === 'function') {
+        this.setState(() => ({ creating: true }))
```
keremgocen: 👍

src/components/ui/SelectBox.js
nicksp: Seems being answered in related `auto` PR :)

PR #178, opened 2018-08-01
Fix SelectBox async methods

bernardodiasc:
This PR is required for https://github.com/x-team/auto/issues/1358 and for https://github.com/x-team/auto/issues/1407

It updates the async methods of SelectBox and fix some hidden bugs here and there.

@nicksp as I mentioned before, Styleguidist have some limitations so the showcases for SelectBox now miss the "full UX" for QA since it requires a container to hold transitional states. This is not a problem and can be worked around. I'm just not investing more time on this now to get things (ListsEditor) going, so please keep that limitation in mind when checking this PR. :) 

I tested this on the application and this is how it looks:

screen recording 2018-07-31 at 09 13 pm

Still need to deal with transitional states on AddToList container, so the https://github.com/x-team/auto/issues/1358 can remain open.

## Auto-ui preview

screenshot-localhost-6060-2018 07 31-21-49-31


bernardodiasc:
I'm not seeing this error on my log 🤔 

It doesn't look related to any change from this PR. Will double check...

nicksp:
Thanks @bernardodiasc also, given that we aren't that interested in having transitional states in auto-ui, would is still remaining on https://github.com/x-team/auto/issues/1358? Could you leave a note there?

And I'm curious if these transitional state will work when integrated into Auto ?

bernardodiasc:
This is how it looks:

screen recording 2018-07-31 at 09 13 pm

The UI will respond only after server response with the updates in this PR. It's still required to make some updates on Auto to make transitions work on AddToList, the effort would be XS. I mentioned this in the https://github.com/x-team/auto/issues/1358, I suggested to repurpose that issue to address this detail, since the SelectBox are now stable. And it's ok to do that task later because the feature works fine without it (it's more an enhancement than a bug).

nicksp:
@bernardodiasc Could you see the error I showed above? If not, then it might be smth related to my machine and we're good to merge changes.

PR #180, opened 2018-08-03
Exclude registered updates

keremgocen:
- Add `isHovering` prop function to `ApplicantBadge` for checking mouse hover state and showing checkbox dropdown


src/components/ui/ApplicantBadge.js
```diff
@@ -28,7 +28,8 @@ type Props = {
   avatar?: Element<*>,
   children?: Element<*> | string,
   onClick?: Function,
-  exclusionFormRender?: Function | void
+  exclusionFormRender?: Function | void,
+  isHovering?: Function
```
bernardodiasc: Hey @keremgocen, what are you trying to accomplish here? If I get right, you want to show control buttons when hovering the ApplicantBadge, thats right? I believe I had implemented this previously in pure CSS, I can see a `cx.displayControlsOnHover` being used but not declared, I think this behaviour was lost unintentionally at some moment.

nicksp:
@keremgocen Please resolve merge issues.

src/components/ui/ApplicantBadge.js
keremgocen: That function is getting mouseOver state from auto

src/components/ui/ApplicantBadge.js
```diff
@@ -308,7 +310,22 @@ class ApplicantBadge extends PureComponent<Props> {
             )}
           </div>
         )}
-        <div className={cx.controls}>
+        <div
+          className={cx.controls}>
+          {exclusionFormRender &&
+            isHovering != null &&
```
nicksp: I guess `hover` is simple enough to check using css `:hover` pseudo-class....

nicksp: wouldn't it be enough to just check like that `isHovering && isHovering()`?

src/components/ui/Button.js
```diff
@@ -230,8 +230,8 @@ class Button extends PureComponent<Props> {
     const colorClassName = colorStyles[color] || ''
     const sizeClassName = sizeStyles[size] || ''
     const extraClassName = [
-      outlined && extraStyles.outlined,
-      outlined && 'outlined',
+      (outlined || pseudolink) && extraStyles.outlined,
```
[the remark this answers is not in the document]
keremgocen: hmmm, I think I only edited this to fix merge conflicts and it was `link` before I changed to `pseudolink` so I am not sure if it's required at all. I will try to handle this whole hover thing with using css only, as in I will try to force the `ApplicantBadge` component to display it's buttons when mouse is hovered on it. That way I would not need to make all these changes mentioned in this review.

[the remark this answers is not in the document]
nicksp: before it was just `outlined`, `pseudolink` is already defined below. Check, code is similar

src/components/ui/ApplicantBadge.js
keremgocen: I think I got it, but I had to use id with `#id` to reference another class https://github.com/x-team/auto-ui/pull/182

keremgocen:
Closing this as I got hover problem fixed using only CSS https://github.com/x-team/auto-ui/pull/182 and the branch is based on latest master.

PR #182, opened 2018-08-04
Display controls on hover

keremgocen:
- Use CSS only for displaying/hiding dropdown icons
- Separate render logic for displaying dropdown menus

peek 2018-08-05 03-27

### DIFF
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -28,7 +28,8 @@ type Props = {
   avatar?: Element<*>,
   children?: Element<*> | string,
   onClick?: Function,
-  exclusionFormRender?: Function | void
+  renderCheckDrop?: Function | void,
+  renderXDrop?: Function | void
 }
 
 const cmz = require('cmz')
@@ -243,7 +244,8 @@ class ApplicantBadge extends PureComponent<Props> {
       tags,
       avatar,
       children,
-      exclusionFormRender
+      renderCheckDrop,
+      renderXDrop
     } = this.props
 
     const cx = mode === 'card' ? cardTheme : tabularTheme
@@ -316,7 +318,19 @@ class ApplicantBadge extends PureComponent<Props> {
           </div>
         )}
         <div id='dropdown-controls' className={cx.controls}>
-          {exclusionFormRender && (
+          {renderCheckDrop && (
+            <Dropdown
+              tooltip
+              label={(
+                <span className={cx.control}>
+                  <SvgIcon icon='check' />
+                </span>
+              )}
+            >
+              {renderCheckDrop()}
+            </Dropdown>
+          )}
+          {renderXDrop && (
             <Dropdown
               tooltip
               label={(
@@ -325,7 +339,7 @@ class ApplicantBadge extends PureComponent<Props> {
                 </span>
               )}
             >
-              {exclusionFormRender()}
+              {renderXDrop()}
             </Dropdown>
           )}
         </div>
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -105,14 +105,14 @@ const cardTheme = {
     }
   `),
 
-  controls: cmz(`
+  controls: cmz('controls', `
     grid-area: control
     display: flex
     visibility: hidden
   `),
 
   displayControlsOnHover: cmz(`
-    &:hover #dropdown-controls {
+    &:hover .controls {
       visibility: visible
     }
   `),
@@ -317,7 +317,7 @@ class ApplicantBadge extends PureComponent<Props> {
             )}
           </div>
         )}
-        <div id='dropdown-controls' className={cx.controls}>
+        <div className={cx.controls}>
           {renderApprovalDropdown && (
             <Dropdown
               tooltip
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -28,8 +28,8 @@ type Props = {
   avatar?: Element<*>,
   children?: Element<*> | string,
   onClick?: Function,
-  renderCheckDrop?: Function | void,
-  renderXDrop?: Function | void
+  renderApprovalDropdown?: Function,
+  renderExclusionDropdown?: Function
 }
 
 const cmz = require('cmz')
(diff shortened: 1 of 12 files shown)


src/components/ui/ApplicantBadge.js
```diff
@@ -109,6 +105,18 @@ const cardTheme = {
...
   `),
 
+  controls: cmz(`
+    grid-area: control
+    display: flex
+    visibility: hidden
+  `),
+
+  displayControlsOnHover: cmz(`
+    &:hover #dropdown-controls {
```
bernardodiasc: hey @keremgocen , lets avoid use ID selectors, what you can do instead is:

```
  controls: cmz('dropdown-controls', `
    grid-area: control
    display: flex
    visibility: hidden
  `),

  displayControlsOnHover: cmz(`
    &:hover .dropdown-controls {

....
```

src/components/ui/ApplicantBadge.js
```diff
@@ -28,7 +28,8 @@ type Props = {
   avatar?: Element<*>,
   children?: Element<*> | string,
   onClick?: Function,
-  exclusionFormRender?: Function | void
+  renderCheckDrop?: Function | void,
```
relztic: Let's make these prop names more explicit:

```js
renderApprovalDropdown?: Function,
renderExclusionDropdown?: Function
```

Also; notice how we can just accept a `Function`, since they are optional in the first place.

src/components/ui/ApplicantBadge.js
```diff
@@ -308,8 +317,20 @@ class ApplicantBadge extends PureComponent<Props> {
             )}
           </div>
         )}
-        <div className={cx.controls}>
-          {exclusionFormRender && (
+        <div id='dropdown-controls' className={cx.controls}>
```
relztic: As per @bernardodiasc comment, plz remove the `id` prop.

src/components/ui/ApplicantBadge.js
keremgocen: I kinda wanted to keep them a bit more abstract, as this `exclusion` and `approval` semantically enforces business logic.

keremgocen: But yeah, when you think about it you can't pass in custom icons either so I guess we might as well sprinkle business logic around:)

src/components/ui/ApplicantBadge.js
keremgocen: this sounds good but I can't make it work without using `id` prop, this is what I've tried so far;
```
controls: cmz('dropdown-controls', `
  grid-area: control
  display: flex
  visibility: hidden
`),

displayControlsOnHover: cmz(`
  &:hover .dropdown-controls {
    visibility: visible
  }
`),
```
and I have class name only inside div,

```
<div onClick={handleClick} className={[cx.mode, cx.displayControlsOnHover, active ? cx.active : ''].join(' ')}>
.
.
```
```
<div className={cx.controls}>
  {renderCheckDrop && (
    <Dropdown
      tooltip
      label={(
        <span className={cx.control}>
          <SvgIcon icon='check' />
        </span>
      )}
    >
      {renderCheckDrop()}
    </Dropdown>
  )}
  {renderXDrop && (
    <Dropdown
      tooltip
      label={(
        <span className={cx.control}>
          <SvgIcon icon='x' />
        </span>
      )}
    >
      {renderXDrop()}
    </Dropdown>
  )}
</div>
```
am I missing something @nicksp ?

src/components/ui/ApplicantBadge.js
keremgocen: please see my comment https://github.com/x-team/auto-ui/pull/182#discussion_r207717653

bernardodiasc: Well, we can't use ID's for elements that repeat on the page, so we need to figure out how to solve this with class :)

nicksp: Agree with guys here. Since we render a bunch of ApplicantBadge's on the page, we can have only one component with this ID.

src/components/ui/ApplicantBadge.js
nicksp: This way of constructing a component can even be beneficial when we'll need to have more controls to manipulate the applicant. So API will stay the same and UI component implementation will almost remain the same, we 'll just need to extend the array we pass to the UI component

keremgocen: @nicksp @bernardodiasc  sounds good but I won't be doing that within the scope of this ticket/PR as it's already delayed quite a lot. But feel free to create a new one for that feature. I just want to get hover working with CSS and without ids and move on.

src/components/ui/ApplicantBadge.js
keremgocen: peek 2018-08-05 03-27
seems good to me, at least until we find a solution to not use `id`s with cmz, otherwise the solution in pure css seems easy http://jsfiddle.net/persianturtle/J4SUb/. Is it possible we can use pure css for that particular area @nicksp ?

keremgocen:
cheers for https://github.com/x-team/auto-ui/pull/182/commits/c577a9297bffb15288ed8f0771a28e8447c2d4e6 @bernardodiasc :clinking_glasses:

keremgocen:
Updated this PR to use conditional props for displaying icons.

src/components/ui/ApplicantBadge.js
```diff
@@ -309,7 +324,19 @@ class ApplicantBadge extends PureComponent<Props> {
...
+              tooltip
+              label={(
+                <span className={cx.control}>
+                  <SvgIcon icon='check' />
+                </span>
+              )}
+            >
+              {renderApprovalDropdown()}
+            </Dropdown>
+          ) : null}
```
nicksp: this is optional actually

src/components/ui/ApplicantBadge.js
```diff
@@ -318,9 +345,9 @@ class ApplicantBadge extends PureComponent<Props> {
                 </span>
               )}
             >
-              {exclusionFormRender()}
+              {renderExclusionDropdown()}
             </Dropdown>
-          )}
+          ) : null}
```
nicksp: this is optional and can be written in the form of `{booleanFlag && (...markup here)}`

PR #185, opened 2018-08-10
Add on click callbacks for applicant badge actions

relztic:
**Release Type:** *New feature* 

Fixes x-team/auto#1294

## Description

- Adds on click callbacks for applicant badge actions.

## Checklist

## Steps to Test or Reproduce

- Navigate to the `ApplicantBadge` component section.
  - Make sure the component passes regression.

## Related PRs

branch | PR
------ | ------
`1294-exclude-registered-from-list--release-aug-6-12` | [Auto](https://github.com/x-team/auto/pull/1447)
`1294-exclude-registered-from-list--release-aug-6-12` | [Auto API](https://github.com/x-team/auto-api/pull/466)

### DIFF
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -323,7 +323,7 @@ class ApplicantBadge extends PureComponent<Props> {
         )}
         {actions.length > 0 && (
           <div className={cx.controls}>
-            {actions.map(({ key, icon: Icon, onClick, render }) => (
+            {actions.map(({ key, icon: Icon, onClick = null, render }) => (
               Icon && (
                 <Dropdown
                   key={key}
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -2,7 +2,6 @@ import React, { PureComponent } from 'react'
 import md5 from 'crypto-js/md5'
 
 import Avatar from './Avatar'
-import SvgIcon from './SvgIcon'
 import TruncatedList from './TruncatedList'
 import Dropdown from './Dropdown'
 
@@ -17,6 +16,12 @@ type Info = {
   tip?: string
 }
 
+type Action = {
+  icon?: Function,
+  onClick?: Function,
+  render?: Function
+}
+
 type Props = {
   id: number,
   mode?: string,
@@ -28,12 +33,7 @@ type Props = {
   avatar?: Element<*>,
   children?: Element<*> | string,
   onClick?: Function,
-  displayApprovalIcon?: boolean,
-  onClickApprovalIcon?: Function,
-  renderApprovalDropdown?: Function,
-  displayExclusionIcon?: boolean,
-  onClickExclusionIcon?: Function,
-  renderExclusionDropdown?: Function
+  actions?: Array<Action>
 }
 
 const cmz = require('cmz')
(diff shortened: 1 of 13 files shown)


src/components/ui/ApplicantBadge.js
```diff
@@ -328,24 +332,24 @@ class ApplicantBadge extends PureComponent<Props> {
             <Dropdown
               tooltip
               label={(
-                <span className={cx.control}>
+                <span className={cx.control} onClick={onClickApprovalIcon}>
```
bernardodiasc: hey @relztic, what do you think about including a prop `onClick` in the Dropdown component and calling it here: https://github.com/x-team/auto-ui/blob/master/src/components/ui/Dropdown.js#L193

relztic: Hi @bernardodiasc,

IMHO, I think it might be better to leave things as they currently are!

Due to the following...

Placing the `onClick` callback _directly_ on the `Dropdown` component could cause issues regarding event bubbling (propagation).

Apart from _not_ having to bypass the callback down to yet another component. wdyt? 😄

relztic:
@nicksp review this please 😄

src/components/ui/ApplicantBadge.js
bernardodiasc: The result here is a `onClick` with another `onClick` on an element inside, I think this results the opposite of what you said regarding event bubbling, sorry if I'm confused.

This is what I suggested, it's not adding any extra event listener:

```
    const handleClick = (e: any) => {
      e && e.stopPropagation()
      const { onClick } = this.props
      onClick && onClick()
      return toggle ? this.toggle() : this.open()
    }
```

If you need the click to not cause the toggle in the dropdown effect, the declaration is like this `<Dropdown toggle={false} />`, this makes sure the clicking will keep open. not sure if it's the case.

There's something else I didn't get about the implementation of this Dropdown, why have `displayApprovalIcon` if I could get the same result checking boolean of `renderApprovalDropdown` is there a case to use `onClickApprovalIcon` even with `displayApprovalIcon` being false?

bernardodiasc: > Apart from not having to bypass the callback down to yet another component.

You are already sending `label` as a callback render for Dropdown, the concept is the same, including `onClick` would grow the component API instead create an exception :)

nicksp:
So guys, what do we do here regarding props? To be honest, I find it really cumbersome and unclear to have these 3 props for each of actions: `display*`/`onClick`/`render*`. What if we decide to extend a set of actions at some point in future?! That would simply result in props grown and even more confusion of what each prop does. 

image

The more logical way, imo, would be to define the following contract for the component.

```js
const actions = [
   exclude: {
     icon: reference to SVGIcon UI component
     action: onHandleExclusion | null
     render: renderExclusionPanel
   },
  approval: {
    icon: reference to SVGIcon UI component
    action: onHandleApproval | null,
    render: renderExclusionPanel
  }
]

<ApplicantBadge {..rest} actions={actions} />
```

Then we just loop over the actions array and just render `<Dropdown>` component based on what was passed. `ApplicantBadge` should no minimum if what is gonna be rendered, so we get these data from the passed `actions` prop. Also, that scales nicely.

So, my suggestion is to implement this way instead.

/cc @bernardodiasc @relztic

relztic:
@nicksp review this please 😄

nicksp:
Hey @bernardodiasc You requested changes a while ago. Could you please check if changes look good to you?

bernardodiasc:
Looks good to me 👍

nicksp:
Will merge once https://github.com/x-team/auto/pull/1447#pullrequestreview-146774782 is addressed.

PR #194, opened 2018-08-24
Tabular view mode for Search form

bernardodiasc:
Dependency for https://github.com/x-team/auto/issues/1416

### SearchForm

- http://localhost:6060/#!/SearchForm

screenshot-localhost-6060-2018 08 24-03-40-36

### Keywords

- http://localhost:6060/#!/Keywords

screen shot 2018-08-24 at 3 55 12 am


bernardodiasc:
@nicksp about the position of the elements, right now this might be problematic to match exactly with mocks. One reason is because this search form suffered changes since the mocks was created, it would be good to check again with @mccrodp or @ryanchartrand if the tabular view can use same positions for the fields as used on card view.

The other reason is that we have 2 buttons in this form, I see a task about eliminating one https://zube.io/x-team/xp-formerly-auto/c/2291, but it's not on backlog radar yet. The point is that the Search form was re-arranged, I'm not sure how strictly can be done accordingly the mocks now.

Anyway, it's possible to update positions with little effort (thanks flexbox), I'm just not sure if will look good (I know that will not match mocks exactly), so I'm double checking.

nicksp:
@bernardodiasc [It's already agreed](https://github.com/x-team/auto/issues/1425#issuecomment-415724133) that we'll proceed with the task of removing a button.

I can see this restructure of UI being not-blocking, so in order not to hold back further progress, let's get it merged as it is, and get back once #1425 is resolved and we have a new Tabular view of a search form in place. I hope that should be OK, @mccrodp

PR #211, opened 2018-09-10
Lists screen - tabular mode layout

bernardodiasc:
- Task https://x-team-internal.atlassian.net/browse/XP-2223
- `Auto` PR https://github.com/x-team/auto/pull/1518

## PR Description

> Styleguidist don't work well having components with `position: fixed` because it's a position related to the screen. As mentioned before, this tool don't provide scope isolation as Storybook does, using `<iframe>`. So it will be hard to provide valuable QA cases for `AdminScreen` and `ListsScreen`. This PR will be better QA'd on front-end: https://github.com/x-team/auto/pull/1518

### TBP

- [x] provide further description of this PR content: https://github.com/x-team/auto-ui/pull/211#issuecomment-420470944


nicksp:
@bernardodiasc How is it going? What's left uncovered?

BTW, please make sure to rebase with release branch to discard merge conflicts 😄

nicksp:
btw, just noticed that I mentioned as a person who closed the PR, but I didn't do anything at all! Weird.

bernardodiasc:
> btw, just noticed that I'm mentioned as a person who closed the PR, but I didn't do anything at all! Weird.

@nicksp this happened before, I think it's because the branch of previous release was deleted. So I opened again and rebased with new release.

> How is it going? What's left uncovered?

New components that holds markup layout of `admin` and `admin/search` routes:

- `ui/AdminScreen` - basically the fixed header and the content
- `ui/Modal` - modal now is declared on AdminScreen
- `ui/ListsScreen` - the layout that toggles display `list` and `tabular`
- `ui/ProfileHeaderLinks` - this was moved to be declared on ListsScreen

These components received small updates regarding list/tabular display shifting:

- `ui/ApplicantGrid`
- `ui/ApplicantBadge`

These components got visual improvements:

- `ui/SearchForm` - it's 100% done in `list` display, in progress in `tabular`
- `ui/ApplicantGridHeader` - this was moved to be declared on SearchForm

These components received updates on animation speed, to make feel faster:

- `ui/Dropdown`
- `ui/SelectBox`
 
This is how the `?display=list` looks atm:

screen recording 2018-09-11 at 08 38 pm

> It's almost the same as current, as it should, but it's after the refactor. At this moment this view mode is stable*. It's easier to control the page layout now using UI screens components, **today I'm focused on finishing the tabular visual fixes (see more info below)**.

This is how the `?display=tabular` looks atm:

screen recording 2018-09-11 at 08 57 pm

---

stable*: well, not yet, the applicant exclusion dropdown need better position UX, this is something that might take more time to have it done, it can be for a follow up and use current solution for this PR (the current position have it's code bytes on `auto` PR)

The completeness of this PR relies on completeness of https://github.com/x-team/auto/pull/1518, because the limitations of QA platform as described previously. To put simple, things that remain for completeness of the task:

- display toggle action, that's made on redux side, also require routes updates (`auto` PR - working on this today)
- visual fixes in tabular composition (in this PR - working on this today)
- applicant click on tabular should toggle display and select applicant (this could be made with https://x-team-internal.atlassian.net/browse/XP-2234)
- applicant exclusion dropdown position (this already have a follow up https://x-team-internal.atlassian.net/browse/XP-2155 and https://x-team-internal.atlassian.net/browse/XP-2165 - I think this actually could be moved to used an approach more like the modal instead use Dropdown, this could result in much smaller render tree).

Let me know anything, it would be great if you have some time to take a look at what already have done, feedback welcomed! :)

bernardodiasc:
@nicksp this is ready for review. Let me know any question!

src/components/ui/SvgIcon.js
```diff
@@ -484,6 +484,30 @@ const getIcon = ({ icon, color, hover }) => {
       <svg width='14px' height='14px' viewBox='0 0 24 24'>
         <path className={fillClassName} transform='scale(1.333) translate(-3, -3)' d='M20.54 5.23l-1.39-1.68C18.88 3.21 18.47 3 18 3H6c-.47 0-.88.21-1.16.55L3.46 5.23C3.17 5.57 3 6.02 3 6.5V19c0 1.1.9 2 2 2h14c1.1 0 2-.9 2-2V6.5c0-.48-.17-.93-.46-1.27zM12 17.5L6.5 12H10v-2h4v2h3.5L12 17.5zM5.12 5l.81-1h12l.94 1H5.12z' />
       </svg>
+    ),
+
+    tabular: (
```
nicksp: Suggest renaming to `list`, as that's what the icon contains and better represents the icon meaning 😄

src/components/ui/SvgIcon.js
```diff
@@ -484,6 +484,30 @@ const getIcon = ({ icon, color, hover }) => {
...
+          <rect className={fillClassName} x='3' y='7' width='12' height='2' />
+          <rect className={fillClassName} x='3' y='14' width='12' height='2' />
+          <rect className={fillClassName} x='0' y='0' width='2' height='2' />
+          <rect className={fillClassName} x='0' y='7' width='2' height='2' />
+          <rect className={fillClassName} x='0' y='14' width='2' height='2' />
+        </g>
+      </svg>
+    ),
+
+    cards: (
```
nicksp: can be renamed to `grid`, as the icon represents a grid view, but can be left as it is.

src/components/ui/SelectBox.js
```diff
@@ -53,29 +53,32 @@ const cx = {
...
 
   selectsEmpty: cmz(`
     & > div:first-of-type {
-      transition: color .15s ease-out, font-size .15s ease-out
+      transition: color .10s ease-out, font-size .10s ease-out
     }
 
     & > div:last-of-type {
-      transition: visibility 1s, opacity .15s ease-out, padding .15s ease-out
+      transition: visibility 0s, opacity .10s ease-out, padding .10s ease-out, height .10s ease-out
```
nicksp: units can be omitted on `0`s

src/components/ui/SelectBox.js
```diff
@@ -53,29 +53,32 @@ const cx = {
...
     & > div:last-of-type {
       width: calc(100% - 20px)
+      height: 24px
       white-space: nowrap
-      overflow: auto
+      overflow-x: overlay
+      overflow-y: hidden
       padding: 0 0 10px
-      transition: visibility 0, opacity .15s ease-out, padding .15s ease-out
+      transition: visibility 0s, opacity .10s ease-out, padding .10s ease-out, height .10s ease-out
```
nicksp: units can be omitted on `0`s

nicksp: We don't need to transition `visibility` here as `opacity` should be enough and more performant.

src/components/ui/SelectBox.js
nicksp: Same as above regarding the reason to not transition `visibility`

src/components/ui/SearchForm.js
```diff
@@ -154,36 +211,69 @@ class SearchForm extends PureComponent<Props> {
...
-              }
-            />
+      <div className={themeClasses.searchFormContainer}>
+        <form onSubmit={onSubmit} className={themeClasses.searchForm}>
+          <div className={themeClasses.selectLists}>
+            <div className={themeClasses.listsSelector}>
+              <SelectBox
+                placeholder='Select Lists'
+                items={lists}
+                visibleItems={3}
```
nicksp: Maybe it makes sense to extract `3` to a constant, so we can set it separately if required to change?

src/components/ui/SearchForm.js
```diff
@@ -154,36 +211,69 @@ class SearchForm extends PureComponent<Props> {
...
     } = this.props
 
-    const theme = mode === 'card' ? cardTheme : tabularTheme
+    const themeClasses = mode === 'tabular' ? tabularTheme : listTheme
+
+    const renderDislpaySwitchButtons = () => (
+      <div className={themeClasses.displayButtons}>
+        <a
+          className={themeClasses.displayButton}
+          onClick={(e) => this.handleSwitchDisplay(e, 'tabular')}
```
nicksp: For performance reasons (to not create new functions on each render and thus force reconciliation), I suggest refactoring that handler to a class prop initializer, so we create a function once: `handleSwitchDisplay = (mode: string) => (e: Object) => {...body}`. So when assigning that to the `onClick` event, you won't need to pass event object explicitly (extra bonus 😏). And you use that like that: `onClick={this.handleSwitchDisplay('tabular')}`

src/components/ui/SearchForm.js
```diff
@@ -154,36 +211,69 @@ class SearchForm extends PureComponent<Props> {
...
+        >
+          <SvgIcon
+            icon='tabular'
+            color={mode === 'tabular' ? 'default' : 'grayscale'}
+            hover='default'
+          />
+        </a>
+        <a
+          className={themeClasses.displayButton}
+          onClick={(e) => this.handleSwitchDisplay(e, 'list')}
```
nicksp: same suggestion as before

src/components/ui/SearchForm.js
```diff
@@ -136,14 +183,24 @@ class SearchForm extends PureComponent<Props> {
     onSelectField: () => {},
     onSubmit: () => {},
     openListEditorModal: () => {},
-    renderApplicantsStatusFilter: null
+    renderApplicantsStatusFilter: null,
+    headerColumns: [],
+    sortBy: '',
+    sortDirection: SORT_DIRECTIONS.ASCENDING,
+    onSortingChange: () => {},
```
nicksp: We could shave off several lines of code, by removing default empty values for callbacks. If they aren't required we just don't pass them and before calling them we check for their existence. So we would be safe. Setting default empty functions is useless as it's treated as a new function each time and might cause extra re-renders.

src/components/ui/SearchForm.js
```diff
@@ -106,14 +139,18 @@ const tabularTheme = {
...
     width: 100%
-    background-color: ${theme.baseBrighter}
-    padding: 10px 30px 30px
+    padding: 20px 30px 30px
     box-sizing: border-box
   `)
 }
 
+type SortDirections = {
+  ASCENDING: 'asc',
```
nicksp: a bit confused here. Since this is a type, we should specify the type of each prop and not the value. Otherwise, we're just duplicating line 170 here 😄

src/components/ui/ListsScreen.js
```diff
@@ -0,0 +1,179 @@
...
+
+  searchForm: cmz(`
+    box-sizing: border-box
+  `),
+
+  applicantGrid: cmz(`
+    box-sizing: border-box
+  `),
+
+  applicant: cmz(``)
```
nicksp: Is this in use?

src/components/ui/Dropdown.js
```diff
@@ -193,7 +208,8 @@ class Dropdown extends PureComponent<Props, State> {
     ].join(' ')
 
     const handleClick = (e: any) => {
-      e && e.stopPropagation()
+      e && e.preventDefault() && e.stopPropagation()
```
nicksp: I think the event always presents here (as it's a callback for a native event) so we can remove the check for its existence.

src/components/ui/Dropdown.js
```diff
@@ -54,7 +54,7 @@ const styles = {
     z-index: 999
     visibility: hidden
     opacity: 0
-    transition: visibility 0s linear 0.2s, opacity 0.2s linear
+    transition: visibility 0s linear 0.1s, opacity 0.1s linear
```
nicksp: we can omit units for `0`s

src/components/ui/ApplicantGridHeader.js
```diff
@@ -152,32 +161,42 @@ class ApplicantGridHeader extends PureComponent<Props> {
...
     })
+
+    const renderCell = (headerColumn) => {
+      if (Array.isArray(headerColumn)) {
+        return (
+          <div key={`grouped${headerColumns.indexOf(headerColumn).toString()}`} className={cx.grouped}>
+            {headerColumn.map(renderCell)}
+          </div>
+        )
+      } else {
```
nicksp: since we do a simple `if () { return .. } else { return .. }`, we can get rid of one nesting by doing just that: 
```js
if () {
  return (
    ...
  )
}

return (
...
)
```

nicksp: easy to read and scan 😏

src/components/ui/ApplicantGrid.js
```diff
@@ -61,9 +80,10 @@ type State = {
 class ApplicantGrid extends PureComponent<Props, State> {
   static defaultProps = {
     items: [],
-    mode: 'card',
+    mode: DISPLAY_MODES.LIST,
     visible: 50,
-    increment: 50
+    increment: 50,
+    onKeyPress: () => {}
```
nicksp: I made a comment before, no real need to specify default values for undefined callbacks. Just don't pass them and check for existence before invoking them.

src/components/ui/ApplicantGrid.js
```diff
@@ -44,14 +52,25 @@ const tabularTheme = {
   `)
 }
 
+type DisplayModes = {
+  LIST: 'list',
```
nicksp: Same as before, this is supposed to be a type representation of props and not their values. Or you had an intention to says that only a specific string value is allowed for a prop?

src/components/ui/ApplicantBadge.js
```diff
@@ -448,61 +461,60 @@ class ApplicantBadge extends PureComponent<Props> {
...
+                  </span>
+                )}
+              >
+                {render && render()}
+              </Dropdown>
+            )
+          ))}
+        </div>
+        <div className={cx.infos}>
+          {info && info.length > 0 && mapInfosToRender(info)}
```
nicksp: Let's copy the `size` function from `auto` app over to `auto-ui`'s helpers module to simplify checks like `if (info && info.length > 0)` to just: `if (size(info)) {}`, or, as in your case, to: `size(info) > 0 && ...`

src/components/ui/ListsScreen.js
```diff
@@ -0,0 +1,179 @@
...
+
+type Props = {
+  applicant?: Element<*>,
+  search?: Element<*>,
+  result?: Element<*>,
+  mode?: $Values<DisplayModes>, // eslint-disable-line no-undef
+}
+
+const DISPLAY_MODES: DisplayModes = {
+  LIST: 'list',
```
nicksp: I noticed that we keep duplicating that definition for display modes. Let's extract that to a single place to re-use across different modules.

src/components/ui/ApplicantBadge.js
```diff
@@ -448,61 +461,60 @@ class ApplicantBadge extends PureComponent<Props> {
...
-          </span>
+          </div>
         )}
       />
     )
 
-    const handleClick = (e) => {
-      e.stopPropagation()
+    const handleClick = (e: any) => {
+      e && e.stopPropagation()
```
nicksp: I think the event always presents here (as it's a callback for a native event) so we can remove the check for its existence.

src/components/ui/ApplicantBadge.js
```diff
@@ -448,61 +461,60 @@ class ApplicantBadge extends PureComponent<Props> {
...
             {`+ ${amount} more`}
-          </span>
+          </div>
         )}
       />
     )
 
-    const handleClick = (e) => {
-      e.stopPropagation()
+    const handleClick = (e: any) => {
```
nicksp: type can be more specific here, like `Object`

src/components/ui/AdminScreen.js
```diff
@@ -0,0 +1,59 @@
...
+    content: Element<*>
+  }
+}
+
+class AdminScreen extends PureComponent<Props, void> {
+  static defaultProps = {
+    header: null,
+    children: null,
+    modal: {
+      onClose: () => {},
```
nicksp: Please check other places, if we don't use a callback, there's no need to pass it. So please remove it from here.

src/components/ui/SvgIcon.js
bernardodiasc: Well, this might be confusing...

```
list icon => tabular view
grid icon => list view
```

nicksp: Agree, what if we change both icon names to smth: `list` and `grid` as it seems to be common names for such icons and representations

image

nicksp: So that list icon -> list view, grid icon -> table view. Should be better now 😄

src/components/ui/SearchForm.js
bernardodiasc: nice, good stuff to know :)

src/components/ui/ApplicantGrid.js
bernardodiasc: > Or you had an intention to says that only a specific string value is allowed for a prop?

That was the case, but I agree that it's kinda strange this way.

nicksp: Maybe we improve to state that it accepts a `string` value at least or smth? Wdyt?

src/components/ui/SvgIcon.js
bernardodiasc: That's a good point. @mccrodp once bring same questioning regarding the icons usage.

I think the "grid" is out of place as an icon choice, because the tabular view looks much more with the list icon anyway, and also there's no grid view, what is called list view is more like a cards view. 

"List" is a word heavily used in this app haha better be careful and double check the decisions. Please let me know what's exactly the final call for this :)

nicksp: I can see your point, Bernardo. It's a bit confusing. But given the icons used in the design, I'd prefer calling them as shown in the picture above.

mccrodp: >So that list icon -> list view, grid icon -> table view. Should be better now 😄

Yep, exactly my concern, but I take your point bernardo and also the designs were like that. But I think @nicksp's suggestion above makes more sense to me and easiest change to make at this point. The list icon should be the list view and the grid at least gives the idea of a table, although it may not exactly look like that 😛

PR #215, opened 2018-09-13
New Pencil Button component

darieldejesus:

Fixes [XP-2159](https://x-team-internal.atlassian.net/browse/XP-2159)

## Description

This is the button that we're gonna be using in the applicant form.

## Checklist

- [ ] set yourself as an assignee
- [ ] set appropriate labels for a PR (initially it's mostly `[zube]: In Review`)
- [ ] set `In Review` status for respective issue as well
- [ ] make sure your code lints (`npm run lint`)
- [ ] Flow typechecks passed (`npm run typecheck`)

## Related PRs

List related PRs against other Auto repos if applicable:

branch | PR
------ | ------
xp-2159-allow-editing-fields | [Auto](https://github.com/x-team/auto/pull/1521)

## Screenshots
making changes

### DIFF
--- a/src/components/ui/PencilButton.js
+++ b/src/components/ui/PencilButton.js
@@ -1,0 +1,60 @@
+// @flow
+/* global SyntheticEvent, HTMLButtonElement */
+
+import React, { Component } from 'react'
+
+import type { Color } from './SvgIcon'
+import SvgIcon from './SvgIcon'
+
+const cmz = require('cmz')
+
+const styles = {
+  editButton: cmz(`
+    background: transparent
+    border: none
+    cursor: pointer
+    font-size: 15px;
+    height: auto !important
+    opacity: 0;
+    transition: opacity .25s linear;
+  `),
+
+  editButtonVisible: cmz(`
+    opacity: 1;
+  `)
+}
+
+type Props = {
+  onClick?: Function,
+  visible?: boolean,
+  color?: Color,
+  hover?: Color
+}
+
+class PencilButton extends Component<Props, void> {
+  static defaultProps = {
+    visible: true,
+    color: 'default',
+    hover: 'default'
+  }
+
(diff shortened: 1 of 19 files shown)


src/components/ui/PencilButton.js
```diff
@@ -0,0 +1,60 @@
...
+
+const cmz = require('cmz')
+
+const styles = {
+  editButton: cmz(`
+    background: transparent
+    border: none
+    cursor: pointer
+    font-size: 15px;
+    height: auto !important
```
joshwnj: @darieldejesus what's the reason for needing `!important` here?

src/components/ui/PencilButton.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  render () {
+    const { visible, color, hover } = this.props
+    return (
+      <button
+        onClick={this.handleClick}
+        className={`${styles.editButton} ${visible ? styles.editButtonVisible : ''}`}
+      >
+        <SvgIcon icon='edit' color={color} hover={hover} />
```
joshwnj: I'm not seeing any hover-states when viewing in Firefox. Chrome works fine.

Strangely, when I view the `SvgIcon` by itself, hover works in both Firefox & Chrome.

darieldejesus: That's weird. Good catch btw. I'm gonna take a look

darieldejesus: I had to use a `<span>` instead of the `<button>` tag that I was using previously. The reason I took this approach is that Firefox doesn't pass the `:hover` to the children if the parent already took it. As you saw, the `SvgIcon` works fine but with the previous version of the `PencilButton`, where the `SvgIcon` was wrapped by a `<button>`, the `:hover` never reached the `SvgIcon` component.

PR #220, opened 2018-09-18
Add shouldSortItems flag prop to SelectBox component

xvicmanx:
**Release Type:** *Breaking change*

## Description

These changes add a control flag (shouldSortItems) for items in the select box component to control whether the items should be sorted or not.

### DIFF
--- a/src/components/ui/SelectBox.js
+++ b/src/components/ui/SelectBox.js
@@ -363,6 +363,7 @@ type Props = {
   width?: number,
   visibleItems?: number,
   expanded?: boolean,
+  shouldSortItems?: boolean,
   hasSearch?: boolean,
   lined?: boolean,
   search?: string,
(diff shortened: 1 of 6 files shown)


src/components/ui/SelectBox.js
```diff
@@ -918,28 +921,33 @@ class SelectBox extends Component<Props, State> {
...
-              </span>
-            )}
-          </li>
-        )}
-        {sortByCreatingFirst(filteredItems.sort(sortById)).map(renderItem)}
-      </ul>
-    )
+    const renderItems = () => {
+      const items = shouldSortItems
+        ? sortByCreatingFirst(filteredItems.sort(sortById))
```
bernardodiasc: hey @xvicmanx, I'm not sure about this, because `sortByCreatingFirst` is about a visual behaviour, it's not related to any business logic. This sorting method is used to display the "Creating 'a new item'..." label, this is something that need to be present no matter the business sorting rules. What can be conditional is only the `filteredItems.sort(sortById)` part. Does that make sense?

xvicmanx:
Thanks @nicksp. I will do that.

src/components/ui/SelectBox.js
xvicmanx: Great observation @bernardodiasc. I will fix that. Thanks!

xvicmanx:
Fixed now

PR #223, opened 2018-09-19
Add status indicator to applicant badge component

xvicmanx:
**Release Type:** *Breaking change*

## Description

These changes add a status indicator to the applicant badge component.
This status indicator is a circle that can have three colors depending on the status:
 - Accepted:  green
- Excluded:  red
- Pending: yellow

## Checklist

- [x] Flow typechecks passed (`npm run typecheck`)

## Impacted Areas

Applicant Badge Component.

## Screenshots
lists-status-color-auto-ui

### DIFF
--- a/src/styles/theme.js
+++ b/src/styles/theme.js
@@ -35,7 +35,10 @@ const palette = {
   porcelain: '#E9EDEE',
   brickRed: '#D32F3B',
   wePeep: '#F7D9DC',
-  bombay: '#B2B6BC'
+  bombay: '#B2B6BC',
+  green: '#5fcf21',
+  mustard: '#f8e71c',
+  red: '#f63a55'
 }
 
 export default wrap({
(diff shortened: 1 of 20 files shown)


src/styles/theme.js
```diff
@@ -35,7 +35,10 @@ const palette = {
   porcelain: '#E9EDEE',
   brickRed: '#D32F3B',
   wePeep: '#F7D9DC',
-  bombay: '#B2B6BC'
+  bombay: '#B2B6BC',
+  green: '#5fcf21',
```
bernardodiasc: Can we use naming from http://chir.ag/projects/name-that-color/ instead `green`, `mustard` or `red`? These general color names can cause a lot trouble when introduce more shades of the same color, that's the reason we are setting all color as constants with most unique as possible names for the color palette.

If you notice, the color red here have hex `#f63a55` while the radicalRed is `#F63954`. While the HEX is slightly different, the actual color seen by the eyes is pretty much the same, I suggest reuse the existing red instead include new one.

xvicmanx: Thanks for the feedbacks bernardo. Let's apply the suggested feedbacks!

xvicmanx:
Suggested changes have been applied. Great tool the `name that color` one. Thanks!

PR #225, opened 2018-09-21
Change the method to compare values

darieldejesus:
**Release Type:** *Bug fix* 

Fixes [XP-2159](https://x-team-internal.atlassian.net/browse/XP-2159)

## Description

Apart from plain values like string and numbers, we're also using objects as values. For that reason, the approach to compare if the **editValue** is equal to the field **value** needs to be changed. Now I installed lodash.isEqual which automatically will compare two values, no matter if they are **string**, **number**, **boolean**, **null**, **undefined** or **objects** with one or more properties.

## Checklist

## Related PRs

List related PRs against other Auto repos if applicable:

branch | PR
------ | ------
xp-2159-allow-editing-fields-master | [Auto](https://github.com/x-team/auto/pull/1530)

## Steps to Test or Reproduce

1. Activate the edition mode by clicking the pencil icon.
2. The **SAVE** button should be disabled when the field value and the current editing value are equal.

## Screenshots
inline editor


src/components/ui/InlineEditor.js
```diff
@@ -1,6 +1,8 @@
 // @flow
 
 import React, { PureComponent } from 'react'
+
```
nicksp: no need to separate as those are all 3rd party libs 😄

PR #237, opened 2018-10-04
Fix Expand / Contract of the "Show" button

reyronald:
**Release Type:** *Non Breaking Feature*

Fixes https://x-team-internal.atlassian.net/browse/XP-2294
(Tabular view: Expand / Contract of the “Show” button when switching views)

## Description

Now, when switching between List and Tabular view,
the "Show" button will not expand, contract, move,
or show any animation. Basically the button animation
was disabled for this scenario.

Also, making the height of the button consistent
among the two different views

## Checklist

- [x] Flow typechecks passed (`npm run typecheck`)
- [x] if any snapshots have been changed, verify that component still works and looks as expected and update changed snapshot

## Related PRs

N/A

## Steps to Test or Reproduce

Switch between List and Tabular view. The "Show" button should not jerk in position or change size with an animation.

* `SearchForm` component in Tabular and List view

## Screenshots

Before: https://www.useloom.com/share/ed18ec4fdf2940cabc1b55fa3c39f598

After:

xp-2294 b


reyronald:
Merged and tested successfully in Chrome, Firefox and Edge

src/components/ui/SearchForm.js
```diff
@@ -145,6 +146,7 @@ const tabularTheme = {
     margin: 0 10px
     height: 58px
     padding: 10px 40px
+    transition: margin, padding 0s
```
nicksp: When we use `0` we don't need to specify any units. It's just useless 😄

reyronald:
@nicksp That's because the `s` unit is necessary fir the `transition` property to work 😄 

before

Fixed:

image

I'm going to add the unit back

reyronald:
Check the docs here: https://developer.mozilla.org/en-US/docs/Web/CSS/transition-delay#Formal_syntax

src/components/ui/SearchForm.js
reyronald: That's the case for size units, but not for time units 😅 . Check my comment here: https://github.com/x-team/auto-ui/pull/237#issuecomment-427159086

Removing the unit is what actually caused the bug that you saw

nicksp:
@reyronald Ah, really, good to know. Actually, I recommend turning off transition which seems to do the same tricks but a bit more logical step.

PR #267, opened 2018-11-05
Simplify Search

reyronald:

Fixes https://x-team-internal.atlassian.net/browse/XP-2389

## Description

Modifies the "Select Fields" select so that it only accepts one item at a time, not multiple.

## Checklist

## Related PRs

branch | PR
------ | ------
XP-2389-simplify-search | [Auto](https://github.com/x-team/auto/pull/1628)
XP-2389-simplify-search | [Auto API](https://github.com/x-team/auto-api/pull/552)

- The Search Form component

## Screenshots

image


reyronald:
Hello Nick, this is ready for review again. The fix for the bug we discussed on Slack is here: https://github.com/x-team/auto-ui/pull/267/commits/b1cc8a66401598d6a7e649b476c4eb1b3c098b86. Like I said it's a little bit difficult to explain so if you need assistance let me know and we can discuss it synchronously.

The workaround to make the select auto-close after selecting an item without that fix would require us specifying a `areItemsToggleable={false}` prop in the "Select Field" component, but that would mean that that items are not toggleable so a user wouldn't be able to clear the input after choosing any item. If that's required, a `hasClear` prop can be used, but that requires passing an `onClear` event and there would be an "X" icon in the placeholder.

bernardodiasc:
Hi @reyronald, I haven't tested this feature on my local yet but I have some questions for you.

> Modifies the "Select Fields" select so that it only accepts one item at a time, not multiple.

About this description, I can understand the context of the changes, but it's not clear what was in fact changed in this PR, because "accepts one item at a time, not multiple" is a characteristic of one particular implementation of SelectBox on auto, not the default behaviour, is that correct? In that case, what are the changes made in this PR and how it affects the SelectBox component?

The reason I'm asking this, is because SelectBox is very complex component, and in order to avoid regressions it's better if we reinforce the QA. It would be great if you can include or update some demos in SelectBox.md to showcase your changes also using edge cases to make sure it would cause any unwanted regression in other behaviours of the component and other implementations in the app.

By reading the code it wasn't clear from where `internalCloseDropdown` comes from, is it a prop? But anyway I get the gist of it. My question is, the "collapse selectbox on selecting an item" became a default behaviour? Will take some time to test on my local, right now just these questions. Cheers :)

bernardodiasc:
Awesome, thanks for the clarification and for the extra example in showcases @reyronald, cheers!

bernardodiasc:
We may find the need of be able to "close dropdown from a children component" much more often in the next features, do you have some suggestion of a different approach other than the "ghost" prop `closeDropdown`? :) 

As a side note, I found that the https://x-team-internal.atlassian.net/browse/XP-2461 issue is potentially related to the way `areItemsToggleable` was used here https://github.com/x-team/auto-ui/blob/master/src/components/ui/SelectBox.js#L574, not sure if this bug affects this task.

reyronald:
With this fix we are already able close the dropdown in any situation. But if you can give me an example of what you have in mind maybe I can provide a better answer.

Regarding the different approach, the "ghost" prop is the behavior that is already in place since a long time ago, removing it would require a refactor of the component and possibly changing the way it is used in other places. This will not be too hard though because there are only like 4 or 5 places were it is used in the entire app.

Perhaps there is a shortest route related to `areItemsToggleable` and event propagation but I didn't investigate that possibility because it could've introduced a regression for components that are using that prop and I didn't have the time when working in this ticket to test it.

Either way if Nick and you don't feel comfortable with the fix in this PR then we would need to dedicate time to investigate other approaches. We could do that in a separate ticket if needed.

Let me know what you guys think

bernardodiasc:
I'm fine with your current implementation :) thanks for clarifying my questions 👍

nicksp:
Resolved merge issues

PR #288, opened 2018-11-25
Add support for Knobs to dynamically edit components in runtime

nicksp:
Fixes https://x-team-internal.atlassian.net/browse/XP-2523 and add support for dynamically editing component right inside Storybook runtime.

**NOTE:** In order to change any specific component, this component should be altered to support knobs. Given that, I suggest doing that incrementally when there's a particular need and when writing new components' stories.


nicksp:
Merging to the parent branch as support for knobs has been added successfully as this can be seen in the Storybook UI:

image

PR #291, opened 2018-11-26
Include children prop at ProfileHeaderLinks

bernardodiasc:
**Release Type:** *non-breaking changes* 

Fixes https://x-team-internal.atlassian.net/browse/XP-2519

## Description

Extend the ProfileHeaderLinks api to accept children props.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-2519-move-the-redesigned-shortcut-element-into-profile-header` | [Auto](https://github.com/x-team/auto/pull/1657)

## Steps to Test or Reproduce

Add any amount of children elements to ProfileHeaderLinks.

`ProfileHeaderLinks`

## Screenshots

screen shot 2018-11-26 at 1 49 19 pm


bernardodiasc:
Hey @galileo, thanks for the feedback!

> What about the storybook snap update, we don't need to do it in that case?

The `ProfileHeaderLinks.stories.js` have the showcase desired for Storybook with sample for the updated API.

> But wouldn't it be better to keep the API consistent, so adding new elements would be defined only by the links property?

That's a good question, I choose this approach in order to keep it open for Elements that comes with own styles and have independent behaviour. The case of links and anchors that we already have in the ProfileHeaderLinks, that's specific kind of elements that behaves in a limited and expected way.

So the update makes the API to use `links` array that becomes external links or internal anchors while the `children` prop expects one or more element that will be rendered after links without any particular styles (only using `display: inline-block` in order to preserve the `<span>` effect used for links - also to force a `<div>` to display inline).

Hope that helps clarify, let me know any other question or suggestion. Cheers!

galileo:
Thanks for the feedback.
Yes, I agree with both of them. I think you took the best approach for our state right now, other approaches would be over complicated and not worth implementing right now. Thanks for the clarification :ok_hand:

src/components/ui/ProfileHeaderLinks.js
```diff
@@ -43,13 +45,18 @@ const cx = {
...
+
   activeLink: cmz(`
     color: ${theme.typoParagraph}
   `)
 }
 
 type Props = {
-  links: Array<Object>
+  links: Array<Object>,
+  children?: ChildrenArray<Element<*>|string>|Element<*>|string
```
reyronald: Hey @bernardodiasc, the `React.Node` type is the type for the `children` props of React components, which includes all of these possible cases. Did you avoid using it for a specific reason?

https://github.com/facebook/flow/blob/1721a0019f213b80b2780930db521cc01269eb7e/lib/react.js#L13-L20

https://github.com/facebook/flow/blob/1721a0019f213b80b2780930db521cc01269eb7e/lib/react.js#L244

src/components/ui/ProfileHeaderLinks.js
```diff
@@ -106,9 +113,26 @@ class ProfileHeaderLinks extends PureComponent<Props, State> {
     })
   }
 
+  renderChildren = (): any => {
+    const { children } = this.props
+
+    return Children.map(children, (child, i) => {
+      return (
+        <li key={i} className={cx.linkWrapper}>
```
reyronald: Perhaps this style shouldn't be called `linkWrapper` anymore since it doesn't only wrap links now. what do you think?

bernardodiasc:
@reyronald thanks for reviewing, your suggestions was updated, please take a look again when you have some time, thanks!

src/components/ui/ProfileHeaderLinks.js
```diff
@@ -106,9 +113,26 @@ class ProfileHeaderLinks extends PureComponent<Props, State> {
     })
   }
 
+  renderChildren = (): any => {
+    const { children } = this.props
+
+    return Children.map(children, (child, i) => {
```
nicksp: Great use of `Children.map` 👍 Also, we could utilize ES6 implicit return to simplify code.

PR #293, opened 2018-11-27
Address remaining visual updates for Tabular view filtering

reyronald:
**Release Type:** *Non Breaking Changes*

Fixes https://x-team-internal.atlassian.net/browse/XP-2498

## Description

This PR moves the filter icon of the tabular view to the left of the column name, instead of the right.

## Checklist

## Related PRs

branch | PR
------ | ------
XP-2498-address-tabular-ui-changes | [Auto](https://github.com/x-team/auto/pull/1663)

Tabular view

## Screenshots

**Before**

image

**After**

image


src/components/ui/ApplicantGridHeader.js
```diff
@@ -148,9 +151,7 @@ const cx = {
   `),
   filter: cmz(`
     & {
-      position: absolute
-      top: 0
-      right: -10px
+      margin-right: 10px;
```
nicksp: It's pretty safe to omit `;` when writing styles with `cmz` as it adds them automatically. Just a little trick to save some keystrokes ;-)

reyronald: I'm aware of that, but I would actually prefer that we don't do this because that's not valid CSS :S. 

We would be relying on `cmz` internals to modify the styles so that they work. If we ever want to migrate to another solutions or migrate the styles somewhere else we would have a lot of work to do adding those semicolons back.

If it was my choice I would stick to only writing valid CSS and always include the semicolons.

nicksp: We already rely on cmz and it would be lots of work anyway to migrate from cmz to any other solution as our styles are written in a specific format. The rest of styles already written without semicolons, so it's a consistency factor as well and the underlying idea behind using `cmz` not to type extra chars :) It's like using CSS preprocessor in some way.

nicksp: That's also what we started doing from the beginning.

nicksp: Just mentioning the story behind :)

nicksp: To be fair, that doesn't mean I don't agree with you 😛 You're right in what you're saying.

reyronald:
> Could we place add some padding below the button as it looks like it's cliped

I did that on purpose so that the "Clear" text was centered in the remaining space of the drop-down at the bottom like it is in the designs, otherwise it looked a bit weird to me.

Adding the padding below this is how it looks now:

**Before, no padding, text is centered**

image

**After, padding, text is not centered, doesn't match designs**

image

(this change is in the Auto PR by the way, not this one)

> In the designs the sorting arrow is centered vertically, could we fix that also please:

You are correct, I broke this! Sorry ✋

nicksp:
> I did that on purpose so that the "Clear" tex

That's probably because we shouldn't use a button but a pure text with styles so it's centered properly ;-)

reyronald:
> That's probably because we shouldn't use a button but a pure text with styles so it's centered properly ;-)

Not sure what you mean. The fact that is not centered properly has nothing to do with the fact it is a button, it has to do with the fact there's an extra padding below the "box" that contains it, which would happen the same way with just a pure text.

Can you elaborate a bit more on your expectations for this part ?

src/components/ui/ApplicantGridHeader.js
reyronald: Yeah I get that too. Just saying why I wasn't comfortable doing that :P. And to be fair I didn't really mention my biggest issue with it, which is that it breaks CSS highlighting when there are no semicolons. But that fine I'll learn to live with that for the sake of following the rules :P

nicksp:
> Can you elaborate a bit more on your expectations for this part ?

Sorry, wasn't super clear. But let's try to center it vertically. Doesn't look good after you applied a fix ;-)

reyronald:
> Sorry, wasn't super clear. But let's try to center it vertically. Doesn't look good after you applied a fix ;-)

Ok, but then it won't appear centered when you hover over it. Should we remove the hover as well? We don't have that state in the design so I'm not sure what we should do.

Your call

nicksp:
> Ok, but then it won't appear centered when you hover over it.

Let's just add spacing above the button and it will be centered.

reyronald:
Hey Nick, just did, this is how it looks now with hover state. Let me know how you feel about it.

https://github.com/x-team/auto/pull/1663/commits/807f333d17d6f1a7e5c56e310871690b0fe4d603

image


image

nicksp:
Appreciate @reyronald Looks clean now UI-wise. Exactly what I had in mind regarding the change 😛

PR #296, opened 2018-11-29
Include readOnly prop in the Button component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-2449

## Description

Included `readyOnly` flag. If this prop is set to true, the hover action will result in none visual changes.

## Checklist

`Button.js`

## Screenshots

screen recording 2018-11-29 at 12 19 pm


src/components/ui/Button.js
```diff
@@ -272,7 +279,8 @@ class Button extends PureComponent<Props> {
     selected: false,
     block: false,
     wide: false,
-    selectbox: false
+    selectbox: false,
+    readOnly: false
```
reyronald: Just mentioning that: If you are going to include a default value for this new prop, then `readOnly` it not really optional anymore from Flow's perspective. There's no way that `readOnly` can be `undefined` inside the definition of the component.

nicksp:
Hey @bernardodiasc Is this ready for review? The ticket isn't in review yet, so wondering what's the correct state :)

bernardodiasc:
> Hey @bernardodiasc Is this ready for review? The ticket isn't in review yet, so wondering what's the correct state :)

@nicksp the PR is ready but not the ticket, there's another PR for `auto` in progress. Will let you know when the ticket is ready for review, thanks!

src/components/ui/Button.js
bernardodiasc: This is interesting. So the flow definition here should not be set as optional, right? Because if I get it correctly, Flow will alert of undefined value in dev runtime of auto-ui, but unexpected usages are still "allowed" outside auto-ui, is this correct?

reyronald: You should not type any prop as optional if you are going to include a default value for it.

If you declare it as optional, Flow will alert of undefined like you said, but this is **impossible**, because the prop has a default value ;). That's why that shouldn't be done.

From the component's usage perspective, Flow is smart enough to understand that a prop is optional when you include the default value for it, so that you can render it without passing that prop.

Check it out:

https://flow.org/try/#0PTAEAEDMBsHsHcBQiCWBbADrATgF1AFSgCGAzqAEoCmxAxvpNrGqAOTY32sDcyuAnhiqgACkwzkAvKADeiUKEixYALlAA7AK5oARlWyhEAX16Ja0MuQCy-AMLMs6quvxUAHrmcATctTq4AOntMWCcXAB4xWAkAPll5UFJcYlwUWlAvKkhiTWhcKIlQaTkFBSVVUABGABoEk2NkEFAAQTyAC1hNAHM20AADcr7QFHJ1WHxo1NDiaGH1UFw24T6C0iGBIQyslHUUKfVEJsWRjNgqUnVWBmIUWZ0nWhzSYQAxOHhh8lI0YjxQZ06PQWsFAmnUmWwSWI4MOYEWKVA-E6p0u+CcVC8wNAGEs-UGoCWHFAekemmeC0JVFY5GI-Uy2Vy+XEa0UOGG+G4AEpENxwjZgo5nPhgDFkEA

@bernardodiasc

bernardodiasc: > You should not type any prop as optional if you are going to include a default value for it.

That's perfectly reasonable. 

I'm going to update Button to reflect this with other props with default value as well:

```diff
diff --git a/src/components/ui/Button.js b/src/components/ui/Button.js
index 9a59f49..0d582d5 100644
--- a/src/components/ui/Button.js
+++ b/src/components/ui/Button.js
@@ -17,19 +17,19 @@ type Props = {
   className: string | CmzAtom,
   size: Size,
   color: Color,
-  outlined: ?boolean,
-  rounded: ?boolean,
-  raised: ?boolean,
-  pseudolink: ?boolean,
-  selected: ?boolean,
-  disabled: ?boolean,
-  block: ?boolean,
-  wide: ?boolean,
-  selectbox: ?boolean,
+  outlined: boolean,
+  rounded: boolean,
+  raised: boolean,
+  pseudolink: boolean,
+  selected: boolean,
+  disabled: boolean,
+  block: boolean,
+  wide: boolean,
+  selectbox: boolean,
   component: string,
   children?: Node,
   tag?: string,
-  readOnly: ?boolean
+  readOnly: boolean
 }
 
 const baseStyles = {
```

Thanks for the feedback 👍

nicksp: @bernardodiasc That's tru. Here's a a thread about that https://github.com/facebook/flow/issues/4871

src/components/ui/Button.stories.js
```diff
@@ -16,16 +51,32 @@ storiesOf('UI Components/Button/Color Options', module)
...
+  ))
+  .add('wide size', () => (
+    <Button wide>Wide size</Button>
+  ))
 
 storiesOf('UI Components/Button/Extra States', module)
-  .add('disabled', () => <Button disabled>Disabled state</Button>)
-  .add('rounded default', () => <Button rounded>Rounded default state</Button>)
-  .add('raised default', () => <Button raised>Raised default state</Button>)
+  .add('disabled', () => (
```
nicksp: We coud simplify that to write one-line :) Committed the fix.

bernardodiasc: sure, I was about to revert this too.
the change reason was to put the `<Component` as the first thing in the line, it's easy to see all cases source code :)

bernardodiasc:
@nicksp @reyronald I've sent an extra commit here https://github.com/x-team/auto-ui/pull/296/commits/aaaac14ee53f3919594f18aa6f2d570e8f4347c1, this is about making a single component with Storybook capabilities to be reused in all showcases. The way to check this working is simple, check out the knobs panel, it will be available in all stories. :)

bernardodiasc:
@nicksp I could not reproduce the bug you mentioned.

> The only concern I have is that we fixed it in a way that now I can't seem to copy a text from buttons, which isn't what the ticket were sayng about.

The approach was to ignore hover effect when read only, this issue should not be related with this PR, I can't see any change that could cause this bug 🤔 

See my test, works fine here:

screen recording 2018-11-29 at 03 25 pm

> We just need to remove the hover state and we'll be fine

Are you saying that removing the shadow effect on hover will not cause any regression in other places? What about the skills that are not read only? Not sure if this gonna work.

nicksp:
> Are you saying that removing the shadow effect on hover will not cause any regression in other places? What about the skills that are not read only?

I was saying only about hover effect from one particular place: skills editor in read mode.

> I could not reproduce the bug you mentioned

Interesting. Let me re-check that once again.

bernardodiasc:
@nicksp sent the fix here https://github.com/x-team/auto-ui/pull/296/commits/4ccfc3c50bec7ebf4b55e31fb46d597e6ba80f17

see how it looks now:

screen recording 2018-11-29 at 03 59 pm

PR #298, opened 2018-11-30
Extend body chart in note editor to 5000

migueliriano:

Fixes https://x-team-internal.atlassian.net/browse/XP-2400

## Checklist

## Related PRs

*N/A*

## Steps to Test or Reproduce
1- Go to admin page
2- In the note section try to update a note.
3- You should be able to see a limit of 5000 characters

- Admin >NoteFeed section > Note update

## Screenshots
changed_words_limit


nicksp:
Hey @migueliriano Just wondering if it's ready for review? ToDo checkboxes are left unticked. Is that on purpose?

migueliriano:
@nicksp - Yes, it's ready for review.

reyronald:
I tested this and it works as expected. The edited note is saved to the database properly and the UI allows you to reach 5000 characters and no more.

This change is only applied to the note editing though, it is still possible to CREATE a note with more than 5000 characters, and in that case when trying to update it you won't be able to change anything in it unless you delete characters until you reach below 5000. I know this is not mentioned in the ticket but wanted to bring it up just in case, is this OK from the user's perspective? @nicksp 	

Also not part of the ticket but that I noticed, the text of long notes overflows to the right:

image

Other than those issues the ticket works fine, but don't know if I should approve. Let me know Nick!

migueliriano:
@reyronald - Thank you for a test this 👍. Here I left what I think.

**text of long notes overflows to the right:** I noticed that you use a long word without spaces instead of a big paragraph, I don't think so that in production we are going to have a case of a note with a long text without spaces between words. This is an example with a normal paragraph.

example_text

**It is still possible to CREATE a note with more than 5000 characters:** I think you're right, we can add the same limit to **CREATE** note to keep the standard and not lose information in the update state.

I can do that in this ticket, but I think we need we need to wait for @mccrodp or @nicksp to confirm this point.

nicksp:
Hey @migueliriano Re: CREATE a note with more than 5000 characters, let's fix it here as it should be a quick qin and a good example of thinking forward. 

RE: `Text of long notes overflows to the right`, let's create a separate ticket if it's an issue.

Wdyt?

nicksp:
And yeah, the code looks correct to me 😄

nicksp:
As stated here https://x-team.slack.com/archives/CDFLP79D1/p1543605649008200 we decided to create a separate ticket as we noticed an increased complexity in Note creation. So we decided to keep the current ticket under the same size and scope.

nicksp:
Rebuild and got less changes 😄

PR #301, opened 2018-12-04
Create ActivityLogsDisplay component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-2551

## Description

Create ActivityLogsDisplay component

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-2551-update-activity-module-frontend` | [Auto](https://github.com/x-team/auto/pull/1693)

## Steps to Test or Reproduce

1. Pull latest from this branch and run `npm install` to install `faker`
2. Run `npm start` and reach out to http://localhost:9001/?selectedKind=UI%20Components%2FActivityLogsDisplay&selectedStory=complete
3. Observe that, in order to expose visual typography vulnerabilities, the data is generated randomly
    - If only `label` is present, display it with the light font-weight
    - If `label` and `value` is present, display the label in light font-weight and value with bold font-weight, separated by `:`
    - If only `value` is present, display it with the light font-weight 

- new component `ActivityLogsDisplay`

## Screenshots

screen recording 2018-12-07 at 11 12 am


bernardodiasc:
This PR is ready for review, I've updated the description section `Steps to Test or Reproduce` with further instructions on how to test this PR independently of the entire feature (that should be tested integrated on app from https://github.com/x-team/auto/pull/1693).

src/components/ui/ActivityLogsDisplay.js
```diff
@@ -0,0 +1,136 @@
...
+type ActivityLogType = {
+  label: string,
+  value: string
+}
+
+type Props = {
+  logs: Array<ActivityType>
+}
+
+export const ActivityLog = ({ label, value }: ActivityLogType) => (
```
reyronald: Do we need to export this component?

src/components/ui/ActivityLogsDisplay.js
```diff
@@ -0,0 +1,136 @@
...
+class ActivityLogsDisplay extends PureComponent<Props, void> {
+  static Log = ActivityLog
+  static defaultProps = {
+    logs: []
+  }
+
+  render () {
+    const { logs } = this.props
+
+    const renderItem = ({ date, activity, user }: ActivityType) => (
```
reyronald: Can we move this outside the `render`? It could be an instance field or actually maybe even defined outside of the current component as a component of its own.

Creating functions is one of the most expensive operations in JavaScript, these `render` methods execute a bunch of times and it might hurt performance a little. I wouldn't do it unless it was necessary

src/components/ui/ActivityLogsDisplay.js
```diff
@@ -0,0 +1,136 @@
...
+        <div className={cx.activity}>{activity}</div>
+        <div className={cx.user}>{user}</div>
+      </div>
+    )
+
+    return (
+      <TruncatedList
+        visible={4}
+        increment={4}
+        items={logs.map(log => renderItem(log))}
```
reyronald: Not a requested change, but this could be:

```js
items={logs.map(renderItem)}
```

This would also perform better since it wouldn't be creating a new callback function on each render.

src/components/ui/ActivityLogsDisplay.js
```diff
@@ -0,0 +1,136 @@
...
+    {value && label ? (
+      <Fragment>
+        <span>: </span>
+        <b>{value}</b>
+      </Fragment>
+    ) : value}
+  </div>
+)
+
+class ActivityLogsDisplay extends PureComponent<Props, void> {
```
reyronald: If you decide to move the `renderItem` function outside, then we could turn this into a function instead of a class.

There's a lot of compilation overhead with classes, look at the generated code from Babel for a class vs. the one for a function:

class

function

So I wouldn't create classes unless its absolutely necessary 😅

src/components/ui/ActivityLogsDisplay.js
```diff
@@ -0,0 +1,136 @@
...
+  static Log = ActivityLog
```
reyronald: Nice usage of compound components 👍

src/components/ui/ActivityLogsDisplay.js
nicksp: That's a neat approach to write it as Ronald suggested in case our function accepts the same argument as the callback gives us.

src/components/ui/ActivityLogsDisplay.js
nicksp: Agree with this point 👍 We should be careful when creating functions, especially when we're inside methods that run very often.

src/components/ui/Note.js
```diff
@@ -174,7 +174,7 @@ class Note extends PureComponent<Props, State> {
 
     const content = (() => {
       try {
-        return markdownCompiler(text)
+        return markdownCompiler(text || '')
```
nicksp: Out of curiosity: is this change relevant to the task scope?

bernardodiasc: I had to update this to make flow tests pass.

src/components/ui/ActivityLogsDisplay.js
bernardodiasc: I used this as first approach, so I was importing as `import ActivityLogsDisplay, { ActivityLog } from './ActivityLogsDisplay'` on Storybook, but that didn't worked as expected when I tried `import ActivityLogsDisplay, { ActivityLog } from 'auto-ui'` at `auto`, of course. 😄So I switched to `<ActivityLogsDisplay.Log />` that looks a good option for sub-components like this, I think there's this patterns already somewhere else in the project.

src/components/ui/ActivityLogsDisplay.js
bernardodiasc: I've refactored the component to be a function, I guess it works fine 👍

bernardodiasc:
After some discussions and refactoring to adjust the desired specs, this PR is ready for review again! Thanks!!!

Demonstration with randomised data:

screen recording 2018-12-11 at 07 43 pm

The `ActivityLog` is expected to be a sub component of `ActivityLogsDisplay`, so it's grouped in it's parent component in the Storybook navigation and there's 2 showcases for the single/grouped variations:

screen recording 2018-12-11 at 07 58 pm

> note in the gif above, there's one situation that appears quickly in the grouped demo that shows only the main label without collapsing effect, the reason is that it's lacking the values, so the demo shows that things don't go crazy or blank if missing data

reyronald:
Did you guys forget to commit the built lib? I don't see them in the recent commits after the changes and I think I had to build it manually locally to make it work. @bernardodiasc 	@galileo

bernardodiasc:
Nice catch @reyronald, just sent the build!

src/components/ui/ActivityLog.js
```diff
@@ -0,0 +1,118 @@
...
+      <div
+        className={logClasses.join(' ')}
+        onClick={() => this.toggleItem()}
+      >
+        <span>{label}</span>
+        {size(value) > 0 && (
+          <Fragment>
+            <span>: </span>
+            <b>
+              {value.map((detail, i) => detail.label).join(', ')}
```
nicksp: Since`i` index isn't in use, we could omit that I guess. But not a request change.

src/components/ui/ActivityLog.js
```diff
@@ -0,0 +1,118 @@
...
+    const { label, value } = this.props
+    const logClasses = []
+    if (size(value) > 0) {
+      logClasses.push(cx.log)
+      logClasses.push(this.state.collapsed ? cx.logCollapsed : cx.logExpanded)
+    }
+    return Array.isArray(value) ? (
+      <div
+        className={logClasses.join(' ')}
+        onClick={() => this.toggleItem()}
```
nicksp: It's good to remember not to create callback functions on each render as it might be memory consuming and useless. We could easily do the same with: `onClick={this.toggleItem}` which means we only have one instance of a callback. Wdyt?

src/components/ui/ActivityLog.js
```diff
@@ -0,0 +1,118 @@
...
+      >
+        <span>{label}</span>
+        {size(value) > 0 && (
+          <Fragment>
+            <span>: </span>
+            <b>
+              {value.map((detail, i) => detail.label).join(', ')}
+            </b>
+            <ul className={[cx.details, this.state.collapsed ? cx.collapsed : ''].join(' ')}>
+              {value.map((detail, i) => (
```
nicksp: As a further improvement, there's a possibility to destructure `details` when defining `map`. This way we can refer to those props directly.

```suggestion
              {value.map((({ label, value }), i) => (
```

src/components/ui/ActivityLog.js
```diff
@@ -0,0 +1,118 @@
...
+                <li className={cx.detail} key={i}>
+                  {detail.label}: <b>{detail.value}</b>
+                </li>
+              ))}
+            </ul>
+          </Fragment>
+        )}
+      </div>
+    ) : (
+      <div>
```
nicksp: Couldn't we use `Fragment` here as well? Or do we need this `div` for stylistic purposes?

src/components/ui/ActivityLogsDisplay.js
```diff
@@ -0,0 +1,102 @@
...
+  activity: Node,
+  user: string
+}
+
+type Props = {
+  logs: Array<ActivityType>
+}
+
+const RenderItem = ({ date, activity, user }: ActivityType, i: number) => (
+  <div className={cx.item}>
```
nicksp: Didn't we forget to specify an index on a `div` since it's being rendered from the loop? Otherwise, we don't seem to need an `i` parameter for the `RenderItem` function.

src/components/ui/ActivityLog.js
bernardodiasc: Thanks for that reminder, I need to keep this in mind indeed!

bernardodiasc:
@nicksp thanks for such careful code review, appreciate that! The PR is ready for review again! Thanks!

PR #302, opened 2018-12-05
Add Shift + Enter key combination support for InlineEditor

galileo:

Fixes https://x-team-internal.atlassian.net/browse/XP-2492

## Description

Added the possiblitiy to save the `InlineEditor` content only when the `shift + enter` combination will be performed.

## Checklist

- [x] component's documentation (`.stories.js` file) is changed or added accordingly to reflect any new or up
dated use cases or variants usage.

## Related PRs

branch | PR
------ | ------
XP-2492-move-cursor-to-next-line | [Auto](https://github.com/x-team/auto/pull/1699)

## Steps to Test or Reproduce

1. Edit notes and perform hit the enter key, it should affect with adding new line instead of saving the note. 
2. To perform the save use the `Shift+Enter` key combination

`InlineEditor`, `Note`


src/components/ui/InlineEditor.js
```diff
@@ -152,16 +155,30 @@ class InlineEditor extends PureComponent<Props, State> {
...
       this.saveChanges()
     } else if (isKeyOfType(keyCode, ESCAPE)) {
       this.abortChanges()
     }
   }
 
+  handleKeyUp = (event: Object) => {
+    const { keyCode } = event
+
+    this.pressedKeys[keyCode] && delete this.pressedKeys[keyCode]
```
nicksp: Once you started collecting all the clicked keys in a new state object called `pressedKeys` that means we need to main that and add/remove keys on specific conditions, and this is kinda complicated and sometimes error-prone. What you can do instead is to check the `shift+Key` combination differently. Please check these resources for context:

- https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/shiftKey
- https://stackoverflow.com/a/6015906

The code will be much easier I assume and you won't need any extra management of `pressedKeys`.

Wdyt? Have you thought about that?

galileo: That is awesome. Actually, I wasn't aware about it. I searched for solutions and this is what I found to tackle this issue that I need to store the pressed keys. So yeah I think this will work much better. I will change the implementation. Thanks @nicksp

nicksp: Yeah, keys are usually easy to handle, so I had a feeling there's a different way to handle that 😄

galileo: It is a great day, this is so easy but I am so happy that you shared with me this knowledge!

I fixed this issue and published changes again.

src/components/ui/InlineEditor.js
```diff
@@ -155,7 +155,11 @@ class InlineEditor extends PureComponent<Props, State> {
     const { shouldSaveOnEnter } = this.props
     const { ENTER, ESCAPE } = KEY_CODES
 
-    if (isKeyOfType(keyCode, ENTER) && shouldSaveOnEnter) {
+    const saveChanges = shouldSaveOnEnter
```
nicksp: @galileo This is looking much cleaner and easier to manage, isn't it 😄 ?

PR #315, opened 2018-12-31
 Create configurable component that implements ToastUI text editor

reyronald:
**Release Type:** *Non Breaking Feature*

Fixes https://x-team-internal.atlassian.net/browse/XP-2672

## Description

Created `RichTextEditor` component that uses ToastUI Editor under the hood.

Docs:

https://nhnent.github.io/tui.editor/api/latest/ToastUIEditor.html#ToastUIEditor

## Checklist

## Steps to Reproduce

* When there's a character limit, the editor shouldn't allow the user to type any more characters into it after the limit has been reached.
* If we load an initial value to the editor and its contents exceed the configured character limit, then the content should be properly loaded but shouldn't be possible any more additional characters. ([see this comment](https://x-team-internal.atlassian.net/browse/XP-2656?focusedCommentId=13547&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13547))

## Screenshots

rich


bernardodiasc:
Great stuff @reyronald, I tested this component on my machine and it seems very good one and well implemented. I tried to use on some different cases and tested the different stories you provided. Overall works as expected, but I found an unexpected behaviour, see screen recording bellow:

screen recording 2019-01-02 at 12 56 pm

Continuing...

screen recording 2019-01-02 at 12 56 pm 1

The problem appears in the Markdown view, I could not use it at all. The wysiwyg view seems to be flawless.

reyronald:
Hey @bernardodiasc 	, thank you for reviewing, testing, and finding that bug!

Thankfully it was easy to fix, I just forgot to add the CodeMirror styles. After including them it now seems to work as expected:

https://github.com/x-team/auto-ui/pull/315/commits/ecb6f93953a4d13c37752e1dde7ff32bb5c3ceb0

Can you re-check on your side?

src/components/ui/RichTextEditor.js
```diff
@@ -0,0 +1,144 @@
...
+
+type State = {
+  characterCount: number
+}
+
+class RichTextEditor extends Component<Props, State> {
+  static defaultProps = {
+    characterLimit: Infinity,
+    hideModeSwitch: true,
+    handleMarkdownChange: () => {},
```
galileo: Great work @reyronald, my only concert for this component are those handle*Change methods, do we really need to have all of them? Wouldn't be better to keep the API for this component as clean as possible? 
I would propose to have only one handleChange method which will take an object 
```
{ markdown, html, text }
```
It can be benefitial as we are calling only one function and the API is much cleaner. But feel free to leave it as it is right now.

nicksp: I agree with @galileo Since we're interested in one mode only, we can provide one callback. We can always introduce new stuff as we go and on demand. Right now, we're fine with having an API proposed by Kamil above.

reyronald:
@galileo 	@nicksp 	Thank you guys, I've applied your suggestion

src/components/ui/RichTextEditor.js
```diff
@@ -0,0 +1,138 @@
+// @flow
+/* global HTMLDivElement */
+
+import React, { Component } from 'react'
+import TUIEditor from 'tui-editor'
```
nicksp: Suggest to name it in a generic name, so if we switch a library, we can keep the variable in the code intact.

PR #320, opened 2019-01-07
Truncate long attached file names

migueliriano:

Fixes https://x-team-internal.atlassian.net/browse/XP-2493

## Checklist

## Related PRs

*N/A*

## Steps to Test or Reproduce
1- Run `npm start`
2- Go to `AttachFiles` component and open an example with files name
3- Reduce the width of the browser.
4 The file name should be truncated and appear a title when you make a hover.

- `AttachFiles` component

#### `AttachFiles` component in  `auto-ui`

truncate_file_name

####  `AttachFiles` component used it in `note` section
truncate_file_name_in_note_section


src/components/ui/AttachFiles.js
```diff
@@ -50,9 +50,12 @@ const FileName = elem.a(cmz(
     & {
       width: 100%
       font-size: 16px
-      line-height: 1
+      line-height: 20px
```
galileo: I am only wondering what is the reason behind setting the size for line-height with a static value. Is there no a way to set up a common multiplier for this kind of values. For example, in this case, we can add the value `1.25` so if we will change one time the font-size. We will not be forced and think about updating the line-height, and the typography will stay the same.

nicksp: Totally valid point. I agree with @galileo. `line-height` should always be set as a [unitless number](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values)

Here's some more details: https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/

migueliriano: Thank you guys, comment addressed on this commit: 7632b8e22268f2fbb3225a0e05708eaac42e566a

PR #324, opened 2019-01-10
 Add Activity Header

galileo:

Fixes https://x-team-internal.atlassian.net/browse/XP-2696

## Description

Add ActivityBar

## Checklist

## Related PRs

branch | PR
------ | ------
auto_branch_name | [Auto](https://github.com/x-team/auto/pull/1769)

## Screenshots

zrzut ekranu z 2019-01-10 22-51-29


nicksp:
Could we please handle conflicting files?

galileo:
Resolved

src/components/ui/ActivityBar.js
```diff
@@ -0,0 +1,60 @@
+import React, { PureComponent } from 'react'
```
reyronald: We forgot to enable `// @flow` here

src/components/ui/ActivityBar.js
```diff
@@ -0,0 +1,60 @@
...
+const cx = {
+  container: cmz(`
+    & {
+      flex: 1 0 100%
+      flex-flow: row
+      display: flex
+      align-items: center
+      font-family: Source Sans Pro
+      font-weight: 400
+      font-size: 14px
```
reyronald: I think font size units should be `rem`, right @bernardodiasc 	?

src/components/ui/ActivityBar.js
```diff
@@ -0,0 +1,60 @@
...
+    max-width: 300px
+    padding: 0 8px
+   
+    white-space: nowrap
+    overflow:hidden
+    text-overflow: ellipsis
+  `),
+  time: cmz(`
+    padding: 1px 8px
+    color: #B2B6BC
```
reyronald: Let's use `${color.bombay}` here

src/components/ui/ActivityBar.js
```diff
@@ -0,0 +1,60 @@
...
+    border-left: 1px solid #C4C4C4
```
reyronald: This is a kind of `silver`, let's create a property for it in the `color`s object and use it here (`${color.silver}` or `${color.silver2}` is there's another one)

src/components/ui/ActivityBar.js
```diff
@@ -0,0 +1,60 @@
...
+  `)
+}
+
+type Props = {
+  text: string,
+  datetime: Date | string | number
+}
+
+class ActivityBar extends PureComponent<Props, void> {
```
reyronald: Let's make this a function component since it's not using state nor instance fields, better for bundling size and speed.

src/components/ui/ActivityBar.stories.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+
+import ActivityBar from './ActivityBar'
+
+storiesOf('UI Components/ActivityBarsDisplay/ActivityBar', module)
```
reyronald: Not sure why the name of the story is `'UI Components/ActivityBarsDisplay/ActivityBar'` instead of just `'UI Components/ActivityBar'` as usual, I'll change this.

src/components/ui/ActivityBar.stories.js
```diff
@@ -0,0 +1,15 @@
...
+  .add('Basic usage', () => (
+    <ActivityBar
+      text='Added to List: Registered'
+      datetime='2019-01-09 11:11:11'
+    />
+  ))
+  .add('missing props (does component explode?)', () => (
+    <ActivityBar />
```
reyronald: If the props are optional then the `Props` flow type should reflect that, I'll fix.

src/components/ui/ActivityBar.js
```diff
@@ -0,0 +1,60 @@
...
+    & {
+      flex: 1 0 100%
+      flex-flow: row
+      display: flex
+      align-items: center
+      font-family: Source Sans Pro
+      font-weight: 400
+      font-size: 14px
+      line-height: 1
+      min-width: 0
```
reyronald: Does anybody know if this `min-width` is necessary?

migueliriano: I think is not necessary too, we can try to remove it and confirm if not affect anything.

nicksp:
> The current implementation is re-using the timeago functionality from the Notes component, so if we fix the common function, it'll be changed how it's displayed there as well, are we ok with that @nicksp ? If not, we are going to have to write a different function for the use-case of this component that is exactly the same except for that extra space there.

I don't think we should change anything in other parts of the app. Instead, I suggest the third option. We add an extra param that specifying whether or not we need an extra space. It's gonna have a default value equal to how it behaves in already existing parts so we don't need to rewrite anything. Does that sound fine?

reyronald:
> I had a similar issue before and I fixed it using line-height: 1.25;.

I'm no CSS expert but I think that's a good solution. The line-height of the Figma file is higher than the one in the current implementation anyway (22px in the Figma file vs 14px here), so I'll use `line-height: 1.6;` which is the closest ratio (22/14 = 1.57).

nicksp:
> I'm no CSS expert but I think that's a good solution. The line-height of the Figma file is higher than the one in the current implementation anyway (22px in the Figma file vs 14px here), so I'll use line-height: 1.6; which is the closest ratio (22/14 = 1.57).

It's totally fine to use float values like `1.57` for a line-height so it closely matches designs 😏

src/components/ui/ActivityBar.js
```diff
@@ -0,0 +1,58 @@
...
+    
+    & > * {
+      flex-shrink: 0
+    }
+  `),
+
+  activity: cmz(`
+    line-height: 1.57
+    flex-shrink: 1
+    max-width: 300px
```
bernardodiasc: I believe we don't need to set `max-width` here. When removed the `max-width` it behaves respecting parent's width:

screen recording 2019-01-11 at 03 10 pm

src/components/ui/SvgIcon.stories.js
```diff
@@ -83,6 +84,7 @@ storiesOf('UI Components/SvgIcon', module)
       <SvgIcon icon='redirect' color='inverted' />
       <SvgIcon icon='filter' color='inverted' />
       <SvgIcon icon='link' color='inverted' />
+      <SvgIcon icon='time' />
```
nicksp: seems like we missed `inverted` here. Gonna fix while resolving a conflict

nicksp:
Successfuly resolved all the conflicts

src/utils/helpers.js
```diff
@@ -50,3 +54,28 @@ export function stopPropagation (event: ?SyntheticEvent<HTMLElement>) {
...
+  }
+
+  const minutesDelta = differenceInMinutes(now, date)
+
+  if (minutesDelta >= 60) {
+    return addSpaceAfterNumber ? `${hoursDelta} h ago` : `${hoursDelta}h ago`
+  }
+
+  if (differenceInSeconds(now, date) >= 60) {
+    return addSpaceAfterNumber ? `${minutesDelta} m ago` : `${minutesDelta}m ago`
```
nicksp: BTW, we could just make a space optional without duplicating `${minutesDelta} ago` part. Let's leave it like that cause it super late and I don't want to fix anything

galileo: :thinking: So in that case I would say that we could introduce a `template` property. 

And use it this way. Of coures names are only examples.
`%number%%type% ago`
`%number% %type% ago`

Also I would say that better approach would be to ask Product Owner if we can not align all those places because maybe this is only an design inconsistency which wasn't found on design level and we would be able to omit the additional complexity introduced here.

src/components/ui/ActivityBar.js
galileo: Do we have a place with common font sizes, wouldn't be beneficial to introduce one and use it wherever we want to use a font size? If new size will arise we will just add a new property into the typography files?
@bernardodiasc @reyronald @nicksp @migueliriano ?
If there is such a place maybe we should spread the knowledge in each place when we will spot this kind of issues?

src/components/ui/ActivityBar.js
galileo: Great catch! I will use this file definetely next time! Thanks for that sugestion @reyronald

PR #325, opened 2019-01-14
Update top profile design

galileo:
**Release Type:** *Non-Breaking feature* 

Fixes https://x-team-internal.atlassian.net/browse/XP-2573

## Description

We are adding here all new components which are used for the new Applicant profile top redesigns,

## Checklist

## Related PRs

branch | PR
------ | ------
auto_branch_name | [Auto](https://github.com/x-team/auto/pull/1771)

### DIFF
--- a/src/components/ui/ActivityBar.js
+++ b/src/components/ui/ActivityBar.js
@@ -1,0 +1,57 @@
+// @flow
+
+import React from 'react'
+import SvgIcon from './SvgIcon'
+
+import { timeSince } from '../../utils/helpers'
+import theme from '../../styles/theme'
+
+const cmz = require('cmz')
+
+const cx = {
+  container: cmz(`
+    & {
+      flex: 1 0 100%
+      flex-flow: row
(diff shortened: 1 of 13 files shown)


reyronald:
Hey @nicksp , this is how it looks after the fix on the no video case. Let me know if it looks good this way or you'd like another change.

**Wide view**:

image


**Narrow view**:

image

---

I noticed something else while playing with the viewport. Depending on the length of the name and the width of the screen, there are cases where the Base icon can overlap on top of the Status dropdown like this:

image

Originally I thought about increasing the media query breakpoint size, or reducing the right-margin of the Base icon, but this wouldn't be a fit-all solution because the screen size at where this happens depends on the length of applicant's name.

What do you think? Is this an issue we should address? Do you have any suggestion? @nicksp

galileo:
I updated this branch with the latest develop and fixed builds and tests. It looks good to me, but I can not approve this one as I created it.

PR #326, opened 2019-01-14
Add default toolbar items

migueliriano:

Fixes https://x-team-internal.atlassian.net/browse/XP-2705

## Description
1- Removed all of these buttons from the default editor config

* Task
* Indents [left & right]
* Table
* Image
* CodeBlock

2- Added feature to configure with a different set of buttons in the `RichTextEditor.js` component.

## Checklist

## Steps to Test or Reproduce
1- Go to `RichTextEditor` component on storybook
2- Confirm that the toolbar buttons listed in the description are not there.

- RichTextEditor

#### Old

old_default_toolbar_items

#### New
current_default_toolbar

### New toolbar buttons in `Note` section
toolbar_in_note_section

### DIFF
--- a/src/components/ui/RichTextEditor.js
+++ b/src/components/ui/RichTextEditor.js
@@ -13,7 +13,7 @@ require('tui-editor/dist/tui-editor.css') // editor ui
 require('tui-editor/dist/tui-editor-contents.css') // editor content
 require('highlight.js/styles/github.css') // code block highlight
 
-const cx = ('RishTextEditor', {
+const cx = ('RichTextEditor', {
   root: cmz(`
     & * {
       box-sizing: border-box;
@@ -42,7 +42,7 @@ const cx = ('RishTextEditor', {
 
 const excludedFromDefaultItems = ['codeblock', 'task', 'indent', 'outdent', 'table', 'image']
 
-export const toolbarItems = [
+const toolbarItems = [
   'heading',
   'bold',
   'italic',
@@ -65,7 +65,7 @@ export const toolbarItems = [
   'codeblock'
 ]
 
-export const defaultToolBarItems = toolbarItems.filter(item => !excludedFromDefaultItems.includes(item))
+const defaultToolBarItems = toolbarItems.filter(item => !excludedFromDefaultItems.includes(item))
 
 type Props = {
   disabled: boolean,
--- a/src/components/ui/RichTextEditor.js
+++ b/src/components/ui/RichTextEditor.js
@@ -13,7 +13,7 @@ require('tui-editor/dist/tui-editor.css') // editor ui
 require('tui-editor/dist/tui-editor-contents.css') // editor content
 require('highlight.js/styles/github.css') // code block highlight
 
-const cx = {
+const cx = ('RishTextEditor', {
   root: cmz(`
     & * {
       box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


src/components/ui/RichTextEditor.js
```diff
@@ -13,7 +13,7 @@ require('tui-editor/dist/tui-editor.css') // editor ui
 require('tui-editor/dist/tui-editor-contents.css') // editor content
 require('highlight.js/styles/github.css') // code block highlight
 
-const cx = {
+const cx = ('RishTextEditor', {
```
reyronald: Why are we including this?

src/components/ui/RichTextEditor.js
```diff
@@ -38,13 +38,41 @@ const cx = {
     font-size: 12px;
     line-height: 20px;
   `)
-}
+})
+
+const excludedFromDefaultItems = ['codeblock', 'task', 'indent', 'outdent', 'table', 'image']
+
+export const toolbarItems = [
```
reyronald: Why are we exporting this? If it's so that a consumer can use it then you have to export it in the bundle as well at `src/index.js`

src/components/ui/RichTextEditor.js
bernardodiasc:
```suggestion
const cx = ('RichTextEditor', {
```

src/components/ui/RichTextEditor.js
```diff
@@ -38,13 +38,41 @@ const cx = {
...
+  'heading',
+  'bold',
+  'italic',
+  'strike',
+  'divider',
+  'hr',
+  'quote',
+  'divider',
```
reyronald: I see that `divider` is included here several times, is that intentional?

src/components/ui/RichTextEditor.stories.js
```diff
@@ -26,6 +26,11 @@ storiesOf('UI Components/RichTextEditor', module)
       characterLimit={10}
     />
   ))
+  .add('with custom toolbar items', () => (
+    <RichTextEditor
+      toolbarItems={['heading', 'bold', 'italic']}
+    />
+  ))
   .add('with knobs', () => (
```
bernardodiasc: we can add [`array`](https://github.com/storybooks/storybook/tree/master/addons/knobs#array) knob here with the new `toolbarItems`

src/components/ui/RichTextEditor.js
migueliriano: I see we are including an explicit class name when we are using a new style in other components, I did the same on this one because not was included. 

Do you think that this can produce an issue?

nicksp: I personally don't see anything wrong with it, just exclude `cmz part from each of rulesets`, it can be declared once at a parent level like we do in other places.. Also fix the typo: `RishTextEditor` should read `RichTextEditor`

src/components/ui/RichTextEditor.js
migueliriano: Actually, I think we don't need to export this data.

src/components/ui/RichTextEditor.js
migueliriano: That is the divider symbol `|` that is used multiple times, so yes we need it duplicated.

divider_element

src/components/ui/RichTextEditor.js
reyronald: > I see we are including an explicit class name when we are using a new style in other components, I did the same on this one because not was included.

This is achieved when we pass a string as the first argument of  the `cmz` function, like Nick said, but this is not what this is doing.

Here we are just initializing a constant with two statements. In those cases only the the last statement is used to initialize the constant, so here the 'RichTextEditor' string is being ignored.

image

migueliriano: I'll rollback this change because I get an error when I try to apply this request change, seems we are using a different version of `cmz` in `auto` => `2.6.0` and `auto-ui` => `3.12.1`, in the latest version we can't apply this `composition` pattern anymore.

src/components/ui/RichTextEditor.js
```diff
@@ -40,11 +40,39 @@ const cx = {
...
+  'table',
+  'image',
+  'link',
+  'divider',
+  'code',
+  'codeblock'
+]
+
+const defaultToolBarItems = toolbarItems.filter(item => !excludedFromDefaultItems.includes(item))
```
nicksp: @migueliriano Sorry, I didn't get it, but what's the point of having two variables and excluding smth, if we could just pass an array of expected buttons? 🤔

nicksp:
@bernardodiasc Could you please give it a final look since changes have been applied?

src/components/ui/RichTextEditor.js
nicksp: How is that different from directly specifying the required toolbar actions in a specific order? I really think what we do is overhead. IMO it's much easier to specify the list just once, it's sufficient in this particular case. So please consider changing that, less code to support in future :)

migueliriano: Comment addressed on this commit: 3a61e90377606a24aa5446440853f3adf84224e1

nicksp:
Happy to merge that.

PR #329, opened 2019-01-17
 Add bottom padding for rich text editor

galileo:
**Release Type:** *Bug fix* 

Fixes https://x-team-internal.atlassian.net/browse/XP-2720

## Description

As wrote in the ticket in JIRA I am not able to reproduce this error. I followed the suggestion left by Nick to fix the issue but can not confirm that this solves it. 

## Checklist

- [x] **manually tested the component** by running it in the browser and checked nothing is broken and op
erates as expected!

## Steps to Test or Reproduce

Type in any rich text field and write content and break lines. If you will reach the end of the veritcal space of that field the scrollbar should no longer overflow the text put into the text field.

All RichText fields used in the app.

## Screenshots

zrzut ekranu z 2019-01-17 17-07-50

### DIFF
--- a/src/components/ui/RichTextEditor.js
+++ b/src/components/ui/RichTextEditor.js
@@ -23,6 +23,10 @@ const cx = {
     & .CodeMirror-scroll {
       overflow-x: hidden !important;
     }
+
+    & .te-ww-container .tui-editor-contents:first-child {
+      padding-bottom: 16px;
+    }
   `),
 
   disabledOverlay: cmz(`


reyronald:
I couldn't reproduce it either, in my case the Editor grows horizontally or veritcally with the text, I never saw any scrollbars, but this is expected for me because I remember explicitly removing the scrollbar when I implemented this feature initially to prevent this from happening, here: https://github.com/x-team/auto-ui/blob/ce7704977937d5cc4d9259eb09cfc652d62c9da8/src/components/ui/RichTextEditor.js#L23-L25

So this bug seems really strange to me 🤔 , I think the editor should never display horizontal scrollbars so that it works in a word-wrap way so I would've gone with that fix instead. But if this is working and product is OK with it then good 👍

nicksp:
@reyronald The fix existed as you can see in screenshots I shared in a ticket, I personally and Paul also could reproduce that on Staging and locally.

nicksp:
The style you referenced above doesn't fix the scrollbar apparently, cause look, just made a recording from Staging. I can clearly reproduce that even without shrinking the browser window:

reyronald:
@nicksp I'm aware since that's what was reported on the ticket, but it did fix it for me originally when I was implementing it, this is how I see it on my side (no scroll visible):

image

nicksp:
That is strange then. I can clearly see that you disable scrolling. So, my guess here would be platform specific thing. But it's still nice to have a bottom padding so the text editor renders consistently and has same spacing around the text.

galileo:
I guess this can be related to the grammarly plugin itself. I can not reproduce this error on an chrome account without it. And with grammarly the scrollbars are poping up for me.

And the screen shared on the task itself also contains grammarly plugin.

zrzut ekranu z 2019-01-17 20-45-14

reyronald:
Wow @galileo 	, that was an incredible find! I assume that Grammarly is reaching into the DOM and adding/changing some styles 😠 😠 

I don't have any plugins on my Chrome that can affect our app here so that would explain why I couldn't reproduce the issue. @nicksp 	can you confirm if you have any plugins like that as well?

PR #331, opened 2019-01-21
Use Toast UI Editor Viewer component for Notes

reyronald:

Fixes https://x-team-internal.atlassian.net/browse/XP-2716

## Description

Add a `more` prop to the `RichTextEditor` component that will add support for the `viewer` mode.

## Checklist

## Steps to Test or Reproduce

1. Open the new `viewer` story under the `RichTextEditor` component

- `RichTextEditor` component
- `Note` component (rendering)

## Screenshots

`RichTextEditor` component in `viewer` mode.

viewer

There are slight differences in how our old markdown engine rendered content versus how Toast UI does it:

**before (with `markdown-to-jsx`)**

before

**after (with Toast UI)**

after

### DIFF
--- a/src/components/ui/Note.js
+++ b/src/components/ui/Note.js
@@ -2,10 +2,8 @@
 
 import React, { PureComponent } from 'react'
 import differenceInHours from 'date-fns/difference_in_hours'
-import { compiler as markdownCompiler } from 'markdown-to-jsx'
 
 import Avatar from './Avatar'
-import Text from './Text'
 import FileLinks from './FileLinks'
 import PencilButton from './PencilButton'
 import InlineEditor from './InlineEditor'
@@ -14,7 +12,7 @@ import RichTextEditor from './RichTextEditor'
 import typo from '../../styles/typo'
 import theme from '../../styles/theme'
 import elem from '../../utils/elem'
-import { replaceBlankLinesForNewLines, timeSince } from '../../utils/helpers'
+import { timeSince } from '../../utils/helpers'
 
 import type { EditorProps, PresenterProps } from './InlineEditor'
 
@@ -91,6 +89,18 @@ const TextWrapper = elem.div(cmz(
   `
 ))
 
+const RichContentWrapper = elem.div(cmz(
+  `
+  & * {
+    font-family: "Source Sans Pro", "Helvetica Neue", Helvetica, Arial, sans-serif
+    font-weight: 300
+    font-size: 20px
+    color: ${theme.typoParagraph}
+    line-height: 30px
+  }
+  `
+))
+
 const FileLinksWrapper = elem.div()
 
 type Props = {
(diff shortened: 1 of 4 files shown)


src/components/ui/RichTextEditor.stories.js
```diff
@@ -41,3 +41,33 @@ storiesOf('UI Components/RichTextEditor', module)
       toolbarItems={array('Toolbar Items', ['heading', 'bold', 'italic'])}
     />
   ))
+  .add('viewer', () => (
```
nicksp: Let's add all possible use cases to see how rendering engine works. We're missing the following items:

1. strike
1. blockquote
1. ordered list
1. unordered list
1. more heading levels

src/components/ui/Note.js
```diff
@@ -152,17 +162,11 @@ class Note extends PureComponent<Props, State> {
...
-        return markdownCompiler(noteText || '')
-      } catch (err) {
-        return noteText
-      }
-    })()
-
-    return TextWrapper({}, <Text content={content} isPureContent />)
+    return TextWrapper(
+      RichContentWrapper(
+        <RichTextEditor initialValue={text} mode='viewer' />
```
nicksp: new prop looks fine.

src/components/ui/Note.js
```diff
@@ -91,6 +89,18 @@ const TextWrapper = elem.div(cmz(
...
 ))
 
+const RichContentWrapper = elem.div(cmz(
+  `
+  & * {
+    font-family: "Source Sans Pro", "Helvetica Neue", Helvetica, Arial, sans-serif
+    font-weight: 300
+    font-size: 20px
+    color: ${theme.typoParagraph}
+    line-height: 30px
```
nicksp: `line-height` almost always should be a unitless value as this will give you greater control over your app's typography. I believe we have had a great discussion with several links shared a while ago 😄 

Some of the links:

- https://meyerweb.com/eric/thoughts/2006/02/08/unitless-line-heights/
- https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/

src/components/ui/Note.js
```diff
@@ -91,6 +89,18 @@ const TextWrapper = elem.div(cmz(
   `
 ))
 
+const RichContentWrapper = elem.div(cmz(
+  `
+  & * {
+    font-family: "Source Sans Pro", "Helvetica Neue", Helvetica, Arial, sans-serif
+    font-weight: 300
+    font-size: 20px
```
nicksp: shouldn't we use `em` for things like texts? This wrapper as I can just wrap the text content, so `em` should be applicable here.

src/components/ui/Note.js
```diff
@@ -14,7 +12,7 @@ import RichTextEditor from './RichTextEditor'
 import typo from '../../styles/typo'
 import theme from '../../styles/theme'
 import elem from '../../utils/elem'
-import { replaceBlankLinesForNewLines, timeSince } from '../../utils/helpers'
```
nicksp: Seems like `replaceBlankLinesForNewLines` isn't in use anymore, so we can safely get rid of it.

src/components/ui/Note.js
reyronald: I'm aware of that. I actually extracted this styles from our typography file and didn't notice how they were defined, would've fixed it otherwise

Thanks for that catch!

src/components/ui/Note.js
reyronald: Yes we should use `rem`. I copied these styles from our typography file and didn't realize they were defined like that, thanks for the catch!

src/components/ui/Note.js
nicksp: Nice, regardless of whether you know that or not I just wanted to make a useful review note that the whole team can learn.

src/components/ui/Note.js
reyronald: The line you are seeing here is a removed lined, notice the red color of the diff. So `replaceBlankLinesForNewLines` had already been taken out :)

nicksp: In other words: the method usage is removed from the file but not the definiotn itself 😄

reyronald: Ah gotcha, I didn't get that is what you meant initially, will remove.

reyronald: Done

PR #334, opened 2019-01-25
Add EmailFeed UI component

migueliriano:

Fixes https://x-team-internal.atlassian.net/browse/XP-2588

## Description

> A follow-up for the https://x-team-internal.atlassian.net/browse/XP‌-2589 task. Once we have component, representing single conversation item, we should create another UI component to represent a feed of email items similar to existing NotesFeed component.
>
> For the mockup please refer to the epic https://x-team-internal.atlassian.net/browse/XP‌-2625

## Checklist

## Related PRs

branch | PR
------ | ------
XP-2589-single-email-item | [Auto UI](https://github.com/x-team/auto-ui/pull/332)

## Steps to Test or Reproduce
1- Go to http://localhost:9001/?selectedKind=UI%20Components%2FEmailFeed&selectedStory=initial%20expanded%20all&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel&background=%23fff
2- Confirm that match with the design: https://www.figma.com/file/MEqB5f6kDR0BrPKgxK5dOi/X-Team---XP---Email-History---15th-Jan?node-id=768%3A965
3- Confirm all these features are in the component:

- By default, only load the 3 most recent e-mails. Clicking 'View 3 More Email' shows 3 more emails associated to this e-mail.
- ~~After clicking the above to load 3 more emails, the button is changed to "View 10 More Emails" and remains at 10 to allow loading the remaining emails associated to this applicant in batches of 10~~
- The "Expand All" toggle allows the display of emails in full, whereas the default is toggled off and just shows the shorter form: https://d.pr/free/i/TKW6nj incl subject, from, to and timestamp
- Click "Expand All" shows the full content, incl the above and the body text for emails shown and the ones loaded on clicking the "View X More" button
- "Expand Single Email:" Expand single email when is clicked anywhere within this outlined region here: https://d.pr/free/i/itt1kB

## Screenshots

email_list_module


src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,270 @@
...
+                color='silver'
+                className={`${cx.viewMore}`}
+              >
+                GO TO FRONT FOR MORE DETAILS
+              </Button>
+            </a>
+          )}
+          viewMore={(amount, action, isFetching) => {
+            const totalAmount = this.state.numberItemsShowed + amount
+            const amountNumber = totalAmount > emails.length ? (emails.length - this.state.numberItemsShowed) : amount
```
migueliriano: I realized that sometimes we get a wrong `amount` number, so this calculation fixes it, anyway we need to create a cleanup ticket or task to fix this issue in `TruncatedList` component.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,270 @@
...
+          items={emails.map(renderEmail)}
+          endListElement={endButtonUrl && (
+            <a href={endButtonUrl} target='_blank' rel='noreferrer' className={cx.endButtonLink} >
+              <Button
+                wide
+                outlined
+                color='silver'
+                className={`${cx.viewMore}`}
+              >
+                GO TO FRONT FOR MORE DETAILS
```
nicksp: Same as before with having all caps. Use CSS for that purpose.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,270 @@
...
+          visible={this.visibleItems}
+          increment={this.incrementItems}
+          items={emails.map(renderEmail)}
+          endListElement={endButtonUrl && (
+            <a href={endButtonUrl} target='_blank' rel='noreferrer' className={cx.endButtonLink} >
+              <Button
+                wide
+                outlined
+                color='silver'
+                className={`${cx.viewMore}`}
```
nicksp: I believe this can be written as simple as: `className={cx.viewMore}`. Or did you mean to concatenate classes here?

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,270 @@
...
+              </Button>
+            </div>
+          </div>
+        </div>
+        { emails.length > 0 && <TruncatedList
+          visible={this.visibleItems}
+          increment={this.incrementItems}
+          items={emails.map(renderEmail)}
+          endListElement={endButtonUrl && (
+            <a href={endButtonUrl} target='_blank' rel='noreferrer' className={cx.endButtonLink} >
```
nicksp: The trailing space after `className={cx.endButtonLink} ` can be removed, right?

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,270 @@
...
+                disabled={isRefreshing}
+                className={isRefreshing ? cx.disabledButton : ''}
+                onClick={this.onRefreshEmails}
+              >
+                {isRefreshing ? 'Loading...' : 'Refresh'}
+              </Button>
+            </div>
+          </div>
+        </div>
+        { emails.length > 0 && <TruncatedList
```
nicksp: Can we re-format this a little bit so it's a bit easier to perceive?

```js
{emails.length > 0 && (
  <TruncatedList 
    prop1={...}
    prop2={...}
  >
    Text
  />
)}
```

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,270 @@
...
+          viewMore={(amount, action, isFetching) => {
+            const totalAmount = this.state.numberItemsShowed + amount
+            const amountNumber = totalAmount > emails.length ? (emails.length - this.state.numberItemsShowed) : amount
+
+            return (<Button
+              wide
+              outlined
+              color='silver'
+              className={cx.viewMore}
+              onClick={() => {
```
nicksp: Is it possible to extract this handler to a separate methos? Again to improve the final code readability.

src/index.js
```diff
@@ -15,6 +15,7 @@ import ColorPalette from './components/ui/ColorPalette'
 import Dropdown from './components/ui/Dropdown'
 import ErrorBox from './components/ui/ErrorBox'
 import Email from './components/ui/Email'
```
nicksp: Can we have a correct alphabetical order please :)? I just noticed that `Email`/`EmailFeed` should be placed before `ErrorBox`? Not a big deal and not a must-have change request either, just to keep consistency 😏

src/index.js
```diff
@@ -73,6 +74,7 @@ export {
   Dropdown,
   ErrorBox,
   Email,
+  EmailFeed,
```
nicksp: Same here with correct order.

nicksp:
There's also a merge conflict that should be very easy to fix as this relates to the order of imports I fixed on dependent PR.

src/components/ui/EmailFeed.js
migueliriano: After doing some test on `TruncatedList` I found that the issue is with `EmailFeed`, we are setting the number of items indicator in `EmailFeed` setting a state after clicking the button `view more`: https://github.com/x-team/auto-ui/blob/d4dec25a092f6e8e8c1bbad8fa97897b51f57078/src/components/ui/EmailFeed.js#L288-L294

Seems this is affecting the current `state` of `TruncatedList` and returning a wrong `amount` value.

migueliriano: Fixed on this commit: 3bd50d479b87ad1d0a95fc7aef084ccef355b379, also I made a little clean up 😏.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+              checked={expandAll}
+              onChange={this.toggleExpandAll}
+            />
+          </div>
+
+          <div className={`${cx.headerRow} ${cx.headerRefreshRow}`}>
+            <div>
+              {lastSyncRefresh &&
+                <span className={cx.lastestEmailSync}>
+                  Lastest email sync: <span className={cx.syncDateAgo}>{timeSince(lastSyncRefresh, false)}</span>
```
nicksp: Here's a typo: should be `Latest`.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+            items={emails.map(renderEmail)}
+            endListElement={endButtonUrl && (
+              <a href={endButtonUrl} target='_blank' rel='noreferrer' className={cx.endButtonLink}>
+                <Button
+                  wide
+                  outlined
+                  color='silver'
+                  className={cx.viewMore}
+                >
+                  Go to front for more details
```
nicksp: Shouldn't `front` start with a capital `F`. It's a service name and I suppose it should read `Front`. Or was it requested like that? Wdyt?

src/components/ui/TruncatedList.js
```diff
@@ -83,18 +91,37 @@ class TruncatedList extends PureComponent<Props, State> {
     }))
   }
 
-  render () {
-    const { items, visible, increment, inserted, viewMore, isFetching, hasMore, listClass, itemClass } = this.props
-    const { allVisible, hiddenItems, page, itemsLength } = this.state
+  getRealVisible = () => this.props.inserted ? this.props.visible - 1 : this.props.visible
```
nicksp: I might be missing the point what does word `real` mean in the scope of the method name? Is there a way to make it more understandable?

src/components/ui/TruncatedList.js
```diff
@@ -83,18 +91,37 @@ class TruncatedList extends PureComponent<Props, State> {
...
 
-    const realVisible = inserted
-      ? visible - 1
-      : visible
+  getNextRealIncrement = () => {
+    const { increment, inserted } = this.props
+    const { hiddenItems, itemsLength } = this.state
+    const realVisible = this.getRealVisible()
```
nicksp: same here with the meaning of a variable.

src/components/ui/TruncatedList.js
```diff
@@ -83,18 +91,37 @@ class TruncatedList extends PureComponent<Props, State> {
...
     const nextIncrement = increment || itemsLength - realVisible
     const nextRealIncrement = hiddenItems < nextIncrement || (inserted && nextIncrement + 1 === hiddenItems)
       ? hiddenItems
       : nextIncrement
-    const nextView = increment ? realVisible + (page > 1 ? (page - 1) * increment : 0) : realVisible + nextRealIncrement
+
+    return nextRealIncrement
```
nicksp: Please check all occurrences of symbols containing the word `real`. I'm not sure a new developer and future yourself would quickly grasp its meaning and purpose 😛

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+class EmailFeed extends PureComponent<Props, State> {
+  static defaultProps = {
+    emails: [],
+    initialExpandedAll: false,
+    isRefreshing: false,
+    errorMessage: ''
+  }
+
+  state = {
+    expandAll: this.props.initialExpandedAll,
```
nicksp: Is there a chance we can clarify the name? I can see that `expandAll` is supposed to be a boolean, but when I read it among the code down below, I can treat its name as a hint to **expand** all sections (an **action**), rather than an indicator **whether** sections are expanded (a **state**). Do you see that difference? So can we be more explicit with this prop name? For booleans I always suggest to stick to a rule to prepend such prop names with `is/are/has/can/shpuld/etc`, anything that will clearly tell you what's the meaning of the variable itself.

src/components/ui/TruncatedList.js
migueliriano: I only extracted this logic to a function to clean more the code and I left the same name. Seem that each time that this component is rendered we have a new value calculated with the `state` data, for that reason I guess they assigned this variable with `real` in the name.

Any suggestion about what name can be in this case?

migueliriano:
@nicksp - I think you are seeing this old mockup: https://d.pr/free/i/TKW6nj.

The design was updated I followed this `figma` to make this component: https://www.figma.com/file/MEqB5f6kDR0BrPKgxK5dOi/X-Team---XP---Email-History---15th-Jan?node-id=768%3A666.

> 1. In 1-14 of 14 of should not be in bold, and it's bold right now:

This is bold in the Figma design:

number_bold

> 2. Email History title and Expand All toggle needs to be on different lines. Check designs:

They are in the same line in the `figma` design

> 3. Could we please also add one more story for EmailFeed that will show emails containing timestamps as it's shown in the screenshot above? I couldn't find it among existing storries, sorry if I missed that.

Good catch, I also checked and we don't have this case yet.

Thank you for your review, tomorrow morning I'm going to address this comments.

migueliriano:
@nicksp - About the numbers, looks good to me, I applied the same: `font-family`, `color` and `font-weight` as we have in the **Figma** document, so I don't know what can be the difference on this case.

nicksp:
@migueliriano I know they have similar styles in Figma, just meant it could be a rendering inconsistency due to fonts or smth. So, as I said, let's keep it as it is now 👍

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+    numberItemsShowed: getInitialNumberItemsShowed(this.props.emails)
+  }
+
+  onRefreshEmails = () => {
+    const { onRefreshEmails } = this.props
```
galileo: I would suggest to extract here the `emails` value and then use it fro the getinitialNumberItemsShowed function argument.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+    if (onRefreshEmails) {
+      onRefreshEmails()
+      this.setState({ numberItemsShowed: getInitialNumberItemsShowed(this.props.emails) })
+    }
+  }
+
+  toggleExpandAll = () => {
```
galileo: This can be written simplier in one line.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+    this.setState((prevState: State) => ({ expandAll: !prevState.expandAll }))
+  }
+
+  getLoadIndicator = () => `1-${this.state.numberItemsShowed} of ${this.props.emails.length}`
+
+  handleViewMore = (action: Function, amount: number) => {
```
galileo: `Function` type is unsafe and should be avoided.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  render () {
+    const { expandAll } = this.state
+    const { emails, isRefreshing, lastSyncRefresh, endButtonUrl, errorMessage } = this.props
+
+    const htmlErrorMessage = (() => {
+      try {
+        return markdownCompiler(errorMessage)
+      } catch (err) {
```
galileo: Do we need the `err` argument if we are not using it in the catch body?

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+
+          <div className={`${cx.headerRow} ${cx.headerRefreshRow}`}>
+            <div>
+              {lastSyncRefresh &&
+                <span className={cx.lastestEmailSync}>
+                  Lastest email sync: <span className={cx.syncDateAgo}>{timeSince(lastSyncRefresh, false)}</span>
+                </span>}
+            </div>
+            <div>
+              {emails.length > 0 && <span className={cx.loadIndicator}>{this.getLoadIndicator()}</span>}
```
galileo: We are callingthis `emails.length` multiple times. Wouldn't be good to extract this into vairable for example `hasEmails`?

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,309 @@
...
+                disabled={isRefreshing}
+                className={isRefreshing ? cx.disabledButton : ''}
+                onClick={this.onRefreshEmails}
+              >
+                {isRefreshing ? 'Loading...' : 'Refresh'}
+              </Button>
+            </div>
+          </div>
+        </div>
+        { emails.length > 0 && (
```
galileo: The same story `hasEmails` ?

src/components/ui/EmailFeed.stories.js
```diff
@@ -0,0 +1,136 @@
...
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+
+import EmailFeed from './EmailFeed'
+
+const currentData = new Date()
+
+const getEmailObjectStructure = (index) => ({
+  subject: `Welcomen Subject ${++index}`,
```
galileo: I don't like this approach maybe we could just don't focus in here about the incrementing this value. It is nice trick but in overall I feel it make the function a little tricky, and hard to understand. What the index value in each place is. The concept of setting and displaying is mixed in one place. But It can be only my feeling.

nicksp:
Hey @migueliriano Please note that I rebased the feature branch with latest `develop`; and the current branch with a feature branch. So it should be all in sync now 😏

src/components/ui/EmailFeed.js
nicksp: Exactly, also website can be previewed by screenreaders and automated readers that ignore CSS, so we should aim at writing text correctly.

src/components/ui/TruncatedList.js
nicksp: Naming is always tricky. Let's keep it as it is, but let's consider renaming along with potential refactoring in a cleanup ticket. Cheers!

migueliriano: I created this cleanup ticket: https://github.com/x-team/auto/issues/1798

src/components/ui/TruncatedList.js
migueliriano: Created cleanup ticket for this: x-team/auto#1798

src/components/ui/TruncatedList.js
migueliriano: Created cleanup ticket for this: x-team/auto#1798

src/components/ui/EmailFeed.stories.js
migueliriano: I changed it a little bit, I hope that is more clear now, was addressed on this commit: bd53f72a0e27837ebc6938c2cc8e70a08c274fdd

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,327 @@
...
+  initialExpandedAll: boolean,
+  lastSyncRefresh?: Date,
+  isRefreshing: boolean,
+  onRefreshEmails?: () => void,
+  endButtonUrl?: string,
+  errorMessage?: string
+}
+
+type State = {
+  isExpandAll: boolean,
```
nicksp: Going t change that to `isExpanded` to reflect the **state** (expanded), rather than an **action** (expand). That's what this state props is about.

src/components/ui/EmailFeed.js
```diff
@@ -0,0 +1,327 @@
...
+                </Button>
+              </a>
+            )}
+            viewMore={(amount, action, isFetching) => (
+              <Button
+                wide
+                outlined
+                color='silver'
+                className={cx.viewMore}
+                onClick={this.handleViewMore.bind(null, action, amount)}
```
nicksp: Instead of this explicit and long `bind` construction, you could easily use the beauty of ES7: prop initializers: `onClick={this.handleViewMore(action, amount)}` and specify this as: `handleViewMore = (action, amount) => (event) => {}` Much cleaner and easier to read.

I'm gonna apply a fix.

PR #335, opened 2019-01-28
`cmz` CSS styles to file

reyronald:
Fixes https://x-team-internal.atlassian.net/browse/XP-2752

### DIFF
--- a/.jest/setupJSDOM.js
+++ b/.jest/setupJSDOM.js
@@ -1,0 +1,29 @@
+const { JSDOM } = require('jsdom')
+
+const jsdom = new JSDOM('<!doctype html><html><body></body></html>')
+const { window } = jsdom
+
+function copyProps (src, target) {
+  Object.defineProperties(target, {
+    ...Object.getOwnPropertyDescriptors(src),
+    ...Object.getOwnPropertyDescriptors(target)
+  })
+}
+
+// Mock date for tests
+const constantDate = new Date('2018-06-13T04:41:20')
+/* eslint no-global-assign:off */
+Date = class extends Date {
+  constructor (date) {
+    super()
+    return constantDate
+  }
+}
+
+window.matchMedia = () => ({})
+global.navigator = {
+  userAgent: 'node.js',
+  appVersion: []
+}
+
+copyProps(window, global)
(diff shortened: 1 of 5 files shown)


.jest/setupJSDOM.js
```diff
@@ -0,0 +1,29 @@
...
+  Object.defineProperties(target, {
+    ...Object.getOwnPropertyDescriptors(src),
+    ...Object.getOwnPropertyDescriptors(target)
+  })
+}
+
+// Mock date for tests
+const constantDate = new Date('2018-06-13T04:41:20')
+/* eslint no-global-assign:off */
+Date = class extends Date {
```
KamWis: are you convinced that this is save? You are overriding `Date` global. What if this global will be in some way used in the implementation itself?

package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "jest": "jest --roots=src",
     "start": "start-storybook -p 9001 -c .storybook",
-    "build": "cross-env NODE_ENV=production webpack --config webpack.standalone.js",
+    "build": "cross-env NODE_ENV=production webpack --config webpack.standalone.js && node cmz-to-file.js > ./lib/styles.css && cmz-min lib/auto-ui.js > ./lib/auto-ui.min.js && cp lib/auto-ui.min.js lib/auto-ui.js && rm -rf lib/auto-ui.min.js",
```
KamWis: this is hard to read - can we have a bash script for that? Isn't there any way to move this as a plugin to the webpack config?

reyronald: Sure, although I wouldn't use a bash script for it since it's not cross platform.

Don't know about the webpack plugin.

.jest/setupJSDOM.js
reyronald: I didn't write this code, this was just moved from one file to the other.

However, I know that this was done to keep deterministic snapshots for stories that use dates.

Regarding safety, naturally there are components that do use `Date` in their implementations. When writing any automated tests the developer must overwrite and  arrange all the globals as necessary before performing any actions or assertions. So what's on this setup file shouldn't be an issue I think.

Currently we don't have any automated tests in this repo so it's kind of irrelevant at the moment. This can be revisited once that changes.

nicksp:
It's been inactive for a quite long time, so I'm gonna close it for the time being. We can reopen that or cherry-pick commits when needed. The parent task is on hold now.

PR #344, opened 2019-02-11
Include new component TwoColumnsLayout

bernardodiasc:
**Release Type:** *Non-braking feature*

Fixes https://x-team-internal.atlassian.net/browse/XP-2791

## Description

Include new component `TwoColumnsLayout`.

## Checklist

## Steps to Test or Reproduce

1. Run `npm start`
1. Go to http://localhost:9001/?selectedKind=UI%20Components%2FTwoColumnsLayout

## Screenshots

screen shot 2019-02-11 at 16 15 43

### DIFF
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -68,9 +68,7 @@ class AdminScreen extends PureComponent<Props, State> {
   }
 
   componentDidUpdate (prevProps: Props) {
-    const { content: prevContent } = prevProps.modal
-    const { content } = this.props.modal
-    if (prevContent !== content) {
+    if (prevProps.modal.prevContent !== this.props.modal.content) {
       this.setState({ isModalOpen: !!this.props.modal.content })
     }
   }
@@ -84,7 +82,7 @@ class AdminScreen extends PureComponent<Props, State> {
   render () {
     const { header, children, modal } = this.props
     const { isModalOpen } = this.state
-    const adminClassNames = isModalOpen ? [cx.admin, cx.adminWidModal].join(' ') : cx.admin
+    const adminClassNames = isModalOpen ? `${cx.admin} ${cx.adminWidModal}` : cx.admin
     return (
       <div className={adminClassNames}>
         {modal && modal.content && (
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -10,7 +10,6 @@ const cmz = require('cmz')
 
 const cx = {
   admin: cmz(`
-    height: 100%
     min-height: 100vh
     overflow-y: hidden
     position: relative
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -10,9 +10,12 @@ const cmz = require('cmz')
 
 const cx = {
   admin: cmz(`
+    height: 100vh
     min-height: 100vh
     overflow-y: hidden
     position: relative
+    display: flex
+    flex-direction: column
   `),
 
   adminWidModal: cmz(`
@@ -22,12 +25,19 @@ const cx = {
   header: cmz(`
     width: 100%
     height: 56px
-    position: fixed
     z-index: 10000
   `),
 
   content: cmz(`
-    padding-top: 56px
+    & {
+      flex: 1 0 0
+      overflow: auto
+      height: 100%
+    }
+
+    & > div {
+      height: inherit
+    }
   `)
 }
 
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -11,26 +11,41 @@ const cmz = require('cmz')
 const cx = {
   admin: cmz(`
     height: 100%
+    min-height: 100vh
+    overflow-y: hidden
     position: relative
   `),
 
+  adminWidModal: cmz(`
+    height: 100vh
+  `),
+
   header: cmz(`
+    width: 100%
     height: 56px
-    position: relative
+    position: fixed
     z-index: 10000
+  `),
+
+  content: cmz(`
+    padding-top: 56px
   `)
 }
 
 type Props = {
   header?: Element<*>,
   children?: Element<*>,
-  modal?: {
+  modal: {
     onClose: Function,
     content: Element<*>
   }
 }
 
-class AdminScreen extends PureComponent<Props, void> {
+type State = {
+  isModalOpen: boolean
+}
+
+class AdminScreen extends PureComponent<Props, State> {
   static defaultProps = {
     header: null,
     children: null,
@@ -39,17 +54,37 @@ class AdminScreen extends PureComponent<Props, void> {
     }
   }
 
+  state = {
+    isModalOpen: !!this.props.modal.content
+  }
+
+  componentDidUpdate (prevProps: Props) {
+    const { content: prevContent } = prevProps.modal
+    const { content } = this.props.modal
+    if (prevContent !== content) {
+      this.setState({ isModalOpen: !!this.props.modal.content })
+    }
+  }
+
+  handleModalClose = () => {
+    const { modal } = this.props
+    this.setState(() => ({ isModalOpen: false }))
+    modal.onClose && modal.onClose()
+  }
+
   render () {
     const { header, children, modal } = this.props
+    const { isModalOpen } = this.state
+    const adminClassNames = isModalOpen ? [cx.admin, cx.adminWidModal].join(' ') : cx.admin
     return (
-      <div className={cx.admin}>
+      <div className={adminClassNames}>
         {modal && modal.content && (
-          <Modal onClose={modal.onClose}>
+          <Modal onClose={this.handleModalClose} isOpen={isModalOpen}>
             {modal.content}
           </Modal>
         )}
         <div className={cx.header}>{header}</div>
-        {children}
+        <div className={cx.content}>{children}</div>
       </div>
     )
   }
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -68,8 +68,10 @@ class AdminScreen extends PureComponent<Props, State> {
   }
 
   componentDidUpdate (prevProps: Props) {
-    if (prevProps.modal.prevContent !== this.props.modal.content) {
-      this.setState({ isModalOpen: !!this.props.modal.content })
+    const { content: prevContent } = prevProps.modal
+    const { content } = this.props.modal
+    if (prevContent !== content) {
+      this.setState({ isModalOpen: !!content })
     }
   }
 
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -10,51 +10,27 @@ const cmz = require('cmz')
 
 const cx = {
   admin: cmz(`
-    height: 100vh
-    min-height: 100vh
-    overflow-y: hidden
+    height: 100%
     position: relative
-    display: flex
-    flex-direction: column
-  `),
-
-  adminWidModal: cmz(`
-    height: 100vh
   `),
 
   header: cmz(`
-    width: 100%
     height: 56px
+    position: relative
     z-index: 10000
-  `),
-
-  content: cmz(`
-    & {
-      flex: 1 0 0
-      overflow: auto
-      height: 100%
-    }
-
-    & > div {
-      height: inherit
-    }
   `)
 }
 
 type Props = {
   header?: Element<*>,
   children?: Element<*>,
-  modal: {
+  modal?: {
     onClose: Function,
     content: Element<*>
   }
 }
 
-type State = {
-  isModalOpen: boolean
-}
-
-class AdminScreen extends PureComponent<Props, State> {
+class AdminScreen extends PureComponent<Props, void> {
   static defaultProps = {
     header: null,
     children: null,
@@ -63,37 +39,17 @@ class AdminScreen extends PureComponent<Props, State> {
     }
   }
 
-  state = {
-    isModalOpen: !!this.props.modal.content
-  }
-
-  componentDidUpdate (prevProps: Props) {
-    const { content: prevContent } = prevProps.modal
-    const { content } = this.props.modal
-    if (prevContent !== content) {
-      this.setState({ isModalOpen: !!content })
-    }
-  }
-
-  handleModalClose = () => {
-    const { modal } = this.props
-    this.setState(() => ({ isModalOpen: false }))
-    modal.onClose && modal.onClose()
-  }
-
   render () {
     const { header, children, modal } = this.props
-    const { isModalOpen } = this.state
-    const adminClassNames = isModalOpen ? `${cx.admin} ${cx.adminWidModal}` : cx.admin
     return (
-      <div className={adminClassNames}>
+      <div className={cx.admin}>
         {modal && modal.content && (
-          <Modal onClose={this.handleModalClose} isOpen={isModalOpen}>
+          <Modal onClose={modal.onClose}>
             {modal.content}
           </Modal>
         )}
         <div className={cx.header}>{header}</div>
-        <div className={cx.content}>{children}</div>
+        {children}
       </div>
     )
   }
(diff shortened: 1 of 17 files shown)


src/components/ui/SearchForm.stories.js
```diff
@@ -1,60 +1,81 @@
...
+  </div>
+)
+
+const FakeFilterTags = () => (
+  <div style={{ width: '100%', margin: '10px 0 0', textAlign: 'center', fontStyle: 'italic' }}>
+    FilterTags is not available at auto-ui, this is a fake representation
+  </div>
+)
+
+export const StorySearchForm = (props) => (
```
migueliriano: We can remove the `()` fro `props` here.

src/components/ui/SearchForm.stories.js
```diff
@@ -1,60 +1,81 @@
...
+    onChangeKeywords={action('onChangeKeywords')}
+    fields={object('Fields', props.fields || fields, 'SearchForm')}
+    onSelectField={action('onSelectField')}
+    statuses={object('Statuses', props.statuses || statuses, 'SearchForm')}
+    onSelectStatus={action('onSelectStatus')}
+    onSubmit={action('onSubmit')}
+    openListEditorModal={action('openListEditorModal')}
+    renderApplicantsStatusFilter={<FakeApplicantsStatusFilter />}
+    tabularFilterTags={<FakeFilterTags />}
+    switchDisplay={(mode) => console.log('switchDisplay:', mode)}
```
migueliriano: You can use `action('switchDisplay')` here. 

```js
switchDisplay={action('switchDisplay')}
```
Action print all the parameters when is called in a callback function, so in this case, going to print the `switchDisplay` text with `mode` data.

src/components/ui/TwoColumnsLayout.stories.js
```diff
@@ -0,0 +1,44 @@
...
+          contentHeading='Search'
+        />
+      </AdminScreen>
+    </Body>
+  ))
+  .add('missing props (does component explode?)', () => (
+    <Body>
+      <TwoColumnsLayout />
+    </Body>
+  ))
```
migueliriano: Can we add a `stories case` when the `AdminScreen` has the modal open?

KamWis:
Gr8 work @bernardodiasc !

One note before I check this PR end to end: For me there are to much changes. PR is about introducing `TwoColumnLayout` but most of the code doesn't belong to it. Can we divide this PR into two for component itself and then for those tweaks in other views?

The reason behind this request is that it's very hard for me to check this PR properly in the context of the `ToColumnLayout` itself.

nicksp:
That's actually a great and legit point @KamWis That's smth I encourage each of us to follow whenever we create a PR. A PR should only contain relevant changes so it's easier to review and reason about. Any other code not relevant to the actual ticket should be extracted to a separate PR whenever it's possible
cc @bernardodiasc

bernardodiasc:
I totally understand this request and I know some parts doesn't looks related, but they are. I believe the most non related thing in this PR is an extra line added to [Button.stories.js](https://github.com/x-team/auto-ui/pull/344/files#diff-1bf0fe76eaaebe5cb0f420b15b4fde15).

I explained the reason in the PR description already, I had to update `AdminScreen`, and because of that I had to make sure some other screen components had no regressions. To properly QA those I upgraded some stories. It's collateral effect, because the screen components rely on window viewport height, it's part of QA of this PR to check everything that was affected by the `AdminScreen` updates, as `ListsScreens` and `Modal` (these looks out of scope, but they actually require QA for the reasons afore explained). 

What was updated in this PR:

```
// components
src/components/ui/AdminScreen.js
src/components/ui/Modal.js
src/components/ui/SvgIcon.js
src/components/ui/TwoColumnsLayout.js
src/components/ui/SettingsScreen.js

// stories
src/components/ui/AdminScreen.stories.js
src/components/ui/ApplicantScreen.stories.js
src/components/ui/Button.stories.js
src/components/ui/ListsScreen.stories.js
src/components/ui/MilestonesScreen.stories.js
src/components/ui/Modal.stories.js
src/components/ui/SearchForm.stories.js
src/components/ui/SettingsScreen.stories.js
src/components/ui/SvgIcon.stories.js
src/components/ui/TwoColumnsLayout.stories.js
```

What can be dropped from PR:

```
src/components/ui/Button.stories.js
src/components/ui/ApplicantScreen.stories.js
src/components/ui/MilestonesScreen.stories.js
```

**I'll remove the unrelated updates, but please make sure to check the others.**

The `src/components/ui/SearchForm.stories.js` had upgrades that also don't look really necessary, but these updates are used on `src/components/ui/ListsScreen.stories.js`, in which I introduced a new story `full composition with AdminScreen`, and for a better representation and leveraging the showcases I introduced the `StorySearchForm`.

KamWis:
@bernardodiasc while I get what you are trying to convey here - Ofc I've read the description and such and I still can't grasp how is all of this related to the `implementation` part. If `TwoColumnLayout` is a standalone component - then why introducing it would break anything in other parts of the system? Is this ticket covering also usage of `TwoColumnLayout` in pages? If yes, it shouldn't be that way. I genuinely think that we should approach this component like `Tab` or `Collapsible`.

I know it's a hassle to divide the PRs and I don't like doing it myself - but I think it's worth it and in the end you will also be even more proud of the final result :)

bernardodiasc:
I really understand your concern @KamWis and I have no problem with it, I'll try to remove as much as possible from this PR. The reason this might not be possible requires a long and detailed explanation, but can be resumed as a "pain point of CSS". My first try was to make this purely standalone and adjustable height on the context, but in the moment I created the `composed in AdminScreen` story, it was clear that I was going to deliver something broken/incompatible right away on it's creation, which seems quite bad idea to me. Let me see how can I clean up this PR, thanks for the feedback.

bernardodiasc:
@KamWis @migueliriano this is ready to be reviewed again. I removed as much as I could and feel safe to do. And included more use cases for different situations. When testing, please make sure to scroll contents, resize screen, etc and check if everything is alright.

screen shot 2019-02-14 at 12 20 20

It's important to make it clear (again) that we are dealing with CSS for viewport height, which is a pain point of CSS. I could refactor layout approach to use flexbox instead of inherited dimensions, and it's much better now. The updates on `AdminScreen` was really necessary and unfortunately it affected the behaviour of the `Modal`, that's why I had to update that component as well and included more showcases do demonstrate that it's working fine after the changes.

screen shot 2019-02-14 at 12 19 57

And because `ListsScreen` is a non-trivial layout component, I made sure to include a bunch of new use cases in the stories so that we can ensure the updates in the `AdminScreen` didn't caused any regression elsewhere --- for the folks unaware, we have a very different way to provide scrolling in the sidebar of `ListsScreen` because of the popup panels available in the ApplicantBadges, see below:

screen shot 2019-02-14 at 12 42 51

I also made sure to check the implementation in the https://github.com/x-team/auto/pull/1813 PR and it's working as expected on my tests (the steps to test are in the PR description).

😅A bit complicated one, thanks for the feedback!

src/components/ui/AdminScreen.stories.js
```diff
@@ -4,29 +4,84 @@ import { action } from '@storybook/addon-actions'
...
+        header={<FakeXHeader />}
+      >
+        {sampleContent}
+      </AdminScreen>
+    </Body>
+  ))
+  .add('default with short content', () => (
+    <Body>
+      <AdminScreen
+        header={<FakeXHeader />}
```
migueliriano: Following the story case: `default with short content`, I think in this case we should remove the `header` prop.

src/components/ui/AdminScreen.js
```diff
@@ -39,17 +63,37 @@ class AdminScreen extends PureComponent<Props, void> {
...
+  handleModalClose = () => {
+    const { modal } = this.props
+    this.setState(() => ({ isModalOpen: false }))
+    modal.onClose && modal.onClose()
+  }
+
   render () {
     const { header, children, modal } = this.props
+    const { isModalOpen } = this.state
+    const adminClassNames = isModalOpen ? [cx.admin, cx.adminWidModal].join(' ') : cx.admin
```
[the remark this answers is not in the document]
bernardodiasc: No reason... It would have a reason if I used this type of declaration:

```js
const adminClassNames = [cx.admin, isModalOpen ? cx.adminWidModal : ''].join(' ')
```

which is not the case, so I'm going to simplify as you suggested.

src/components/ui/AdminScreen.stories.js
```diff
@@ -4,29 +4,84 @@ import { action } from '@storybook/addon-actions'
...
       >
-        <div>Anything goes in the content</div>
+        {sampleContent}
       </AdminScreen>
-    </div>
+    </Body>
   ))
-  .add('with modal', () => (
-    <div style={{ height: '500px' }}>
+  .add('short content with Modal with short content', () => (
```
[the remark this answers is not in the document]
bernardodiasc: Hmm not sure I get your question.

The difference is in the content, not the header.

short content = `<div>bump</div>`
tall content = `Array(80).fill('Anything goes in the content body').map((each, i) => <div key={`content-${i}`}>{each}</div>)`

does that make it clear? any different suggestion?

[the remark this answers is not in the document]
bernardodiasc: Better explaining, the effect I wanted to simulate here is the vertical scrolling. 

Independently of the header, the short example need to fill the entire viewport height. Which can be better seen with the Modal open.

```jsx
<AdminScreen>
    <div>bump</div>
</AdminScreen>
```

A tall content would crate vertical scrolling only in the content area and keep the header fixed on the top. This also can be experimented with the Modal open which in that case will not cause the content to be scrollable, if you try to scroll, the Modal overlay should be scrollable. An example would look like:


```jsx
<AdminScreen>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
    <div>bump</div>
</AdminScreen>
```

It can be confusing because it's a mix of different situations and I tried to cover all of them. One practical example is the "Edit Lists" modal in the App, if you try in a small screen, you should be able to scroll the Modal overlay without scrolling the AdminScreen content area, and once you close the Modal what will be scrollable is the AdminScreen content area.

More situations can be when AdminScreen content area don't have a dedicated layout, like the TwoColumnsLayout. This happens in the `create-applicant` page. In the Lists and Advanced Search pages it does have a layout component in the content area, and the scrolling are handled on another places, what to observe them is that we don't redundancy clashing of scrollings, like when multiple scrolls can be seen and the scrolling experience looks a bit weird (I had to deal and fix with such case actually).

Hope that helps your debugging, let me know any question. :)

src/components/ui/AdminScreen.js
```diff
@@ -39,17 +63,37 @@ class AdminScreen extends PureComponent<Props, void> {
...
+    isModalOpen: !!this.props.modal.content
+  }
+
+  componentDidUpdate (prevProps: Props) {
+    const { content: prevContent } = prevProps.modal
+    const { content } = this.props.modal
+    if (prevContent !== content) {
+      this.setState({ isModalOpen: !!this.props.modal.content })
+    }
+  }
```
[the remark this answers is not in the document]
bernardodiasc: Well, I think I get it, will update to the second alternative

```
    if (prevProps.modal.content !== this.props.modal.content) {
      this.setState({ isModalOpen: !!this.props.modal.content })
    }
```

bernardodiasc:
@migueliriano thanks for your suggestions! I updated the PR. Just regarding the comments https://github.com/x-team/auto-ui/pull/344#discussion_r256889270 and https://github.com/x-team/auto-ui/pull/344#discussion_r256897186, it looks like that's not the intention. I provided some explanation of the original intention, and given that explanation I believe there's nothing to change. If you have a better description for the story I'm fine on updating, otherwise this PR should be ready again. Thanks!

src/components/ui/AdminScreen.stories.js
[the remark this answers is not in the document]
bernardodiasc: Actually... there's something wrong in the showcase, I'm going to fix now... 😞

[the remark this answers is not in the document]
migueliriano: Ok, thank you for the explanation, now it's clear 👍

src/components/ui/AdminScreen.js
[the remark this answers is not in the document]
migueliriano: This alternative looks good to me, thank you for updating it 👍

bernardodiasc:
@migueliriano @KamWis @nicksp now a proper `Non-braking feature` PR with only required updates.

I kept the showcases with AdminScreen as those can expose the issue that I fixed with the extra changes, I hope those are fine to remain in the PR 😄

src/components/ui/TwoColumnsLayout.js
```diff
@@ -0,0 +1,161 @@
...
+
+import SvgIcon from './SvgIcon'
+
+import theme from '../../styles/theme'
+import { typeface } from '../../styles/typo'
+
+import type { Element } from 'react'
+import type { Icon } from './SvgIcon'
+
+const cmz = require('cmz')
```
KamWis: Why we are not using `import` here? :)

src/components/ui/TwoColumnsLayout.js
```diff
@@ -0,0 +1,161 @@
...
+type Props = {
+  sidebar: Element<*>,
+  sidebarHeading: string,
+  sidebarWidth: number,
+  sidebarIcon: Icon,
+  content: Element<*>,
+  contentHeading: string
+}
+
+class TwoColumnsLayout extends PureComponent<Props, void> {
```
KamWis: I assume this is a class for the future `toggle feature`? :)

src/components/ui/TwoColumnsLayout.js
```diff
@@ -0,0 +1,161 @@
...
+    content: null,
+    contentHeading: ''
+  }
+
+  render () {
+    const { sidebar, sidebarHeading, sidebarWidth, sidebarIcon, content, contentHeading } = this.props
+
+    return (
+      <div className={cx.layout}>
+        <div className={cx.sidebar} style={{ width: `${sidebarWidth}px` }}>
```
KamWis: Can we extract sidebar div to a function? This will improve readability of the render method

src/components/ui/TwoColumnsLayout.js
```diff
@@ -0,0 +1,161 @@
...
+              <div className={cx.sidebarHeadingText}>
+                {sidebarHeading}
+              </div>
+            </div>
+          )}
+          <div className={cx.sidebarBody}>
+            {sidebar}
+          </div>
+        </div>
+        <div className={cx.content}>
```
KamWis: Same here - can we extract it?

src/components/ui/TwoColumnsLayout.js
bernardodiasc: I believe the import here is not compatible, I don't know why. This is a great question, I'm not really how to fix it, maybe something for an `cleanup` task.

src/components/ui/TwoColumnsLayout.js
bernardodiasc: Not sure I get your question. Is it about the `sidebarWidth` prop? I think we decided to use a prop to set the width for the sidebar, for the future as you said. The `PureComponent` class would help on re-rendering check if I'm not mistaken, or is it something that can be accomplished with a pure function component?

bernardodiasc:
@KamWis can you take a look again and see if that's the pattern you expected regarding the abstraction of the render? Thanks!

src/components/ui/TwoColumnsLayout.js
KamWis: well, it's just seems that it could be a pure function component instead of a class since it doesn't do anything with the state - so I thought you've decided  to use class in order to have possibility to use `state` in the future for a `toggle` feature

bernardodiasc: Ah, yeah, that's it. Indeed, there's no use now but will have eventually. Do you think it should be changed now?

KamWis: na, leave it :) I know what's the intent behind it.

nicksp:
Just rebased with a feature branch and resolved merge conflicts.

PR #347, opened 2019-02-14
Create generic Collapsible component

migueliriano:

Fixes https://x-team-internal.atlassian.net/browse/XP-2795

## Description

> The component should hold none design, only the behaviour and minimal CSS for positioning and display
> Add component into Storybook and provide stories for common and edge cases

## Checklist

## Steps to Test or Reproduce
Go to Storybook: http://localhost:9001/?selectedKind=UI%20Components%2FCollapsible&selectedStory=basic%20usage&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel&background=%23fff

## Screenshots

generic_collapsed_component

### DIFF
--- a/src/components/ui/Collapsible/Collapsible.stories.js
+++ b/src/components/ui/Collapsible/Collapsible.stories.js
@@ -41,7 +41,7 @@ storiesOf('UI Components/Collapsible', module)
     </Collapsible.Container>
   ))
   .add('with initial expanded', () => (
-    <Collapsible.Container initialCollapsed>
+    <Collapsible.Container initialExpanded>
       <Collapsible.Header>
         <ExampleHeader text='Collapsed Header' />
       </Collapsible.Header>
@@ -114,6 +114,23 @@ storiesOf('UI Components/Collapsible', module)
       </Collapsible.Header>
       <Collapsible.Body>
         <p>{exampleText}</p>
+      </Collapsible.Body>
+    </Collapsible.Container>
+  ))
+  .add('with nested Collapsibles', () => (
+    <Collapsible.Container>
+      <Collapsible.Header>
+        <h1>First Level</h1>
+      </Collapsible.Header>
+      <Collapsible.Body>
+        <Collapsible.Container>
+          <Collapsible.Header>
+            <h2>Second Level</h2>
+          </Collapsible.Header>
+          <Collapsible.Body>
+            <p>{exampleText}</p>
+          </Collapsible.Body>
+        </Collapsible.Container>
       </Collapsible.Body>
     </Collapsible.Container>
   ))
--- a/src/components/ui/Collapsible/Collapsible.stories.js
+++ b/src/components/ui/Collapsible/Collapsible.stories.js
@@ -4,23 +4,23 @@ import { action } from '@storybook/addon-actions'
 
 import Button from '../Button'
 
-import Collapsable from './'
+import Collapsible from './'
 
 const exampleText = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.'
 
-const ExampleHeader = ({ isCollapsed, text, ...props }) => (
+const ExampleHeader = ({ isExpanded, text, ...props }) => (
   <span
     {...props}
-    className={`${isCollapsed ? 'isCollapsed' : ''}`}
+    className={`${isExpanded ? 'isExpanded' : ''}`}
   >
     {text}
   </span>
 )
 
-const ExampleHeaderButton = ({ isCollapsed, text, ...props }) => (
+const ExampleHeaderButton = ({ isExpanded, text, ...props }) => (
   <Button
     {...props}
-    color={`${isCollapsed ? 'gray' : 'normal'}`}
+    color={`${isExpanded ? 'gray' : 'normal'}`}
   >
     {text}
   </Button>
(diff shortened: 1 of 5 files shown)


src/components/ui/Collapsible/Header.js
```diff
@@ -0,0 +1,27 @@
...
+
+const Header = (props: Props) => {
+  const { children, onClick, isCollapsed } = props
+
+  const handleOnClick = (event: Object) => {
+    event.preventDefault()
+    onClick(!isCollapsed)
+  }
+
+  return React.isValidElement(children) ? React.cloneElement(
```
KamWis: does Header accept only 1 child? If yes, it should be checked via `React.Children.only` method. If you will decide to use it - I suggest to check if children exists before that, like so:
```
if (!children) {
  return null
}

const validatedChildren = React.Children.only(children)
```

Then we can proceed to final return. I don't know if `React.isValidElement(validatedChildren )` would be then necessary in this case or not. WDYT?

src/components/ui/Collapsible/Container.js
```diff
@@ -0,0 +1,74 @@
...
+    headerWrapper: (props: { children: Element<*> }) => <Fragment>{props.children}</Fragment>,
+    bodyWrapper: (props: { children: Element<*> }) => <Fragment>{props.children}</Fragment>,
+    onChange: () => {}
+  }
+
+  state = {
+    isCollapsed: this.props.initialCollapsed
+  }
+
+  componentDidUpdate (prevProps: Props) {
```
KamWis: why do we need `componentDidUpdate` here? :)

src/components/ui/Collapsible/Collapsible.stories.js
```diff
@@ -0,0 +1,102 @@
...
+      <Collapsable.Header>
+        <ExampleHeader text='Collapsed Header' />
+      </Collapsable.Header>
+      <Collapsable.Body>
+        <p>{exampleText}</p>
+      </Collapsable.Body>
+    </Collapsable.Container>
+  ))
+  .add('with onChange event', () => (
+    <Collapsable.Container onChange={action('onChange collapsed')}>
```
KamWis: maybe instead of `onChange collapsed`: `onChange triggered`? since it will also trigger on `show`

src/components/ui/Collapsible/Header.js
migueliriano: Yes, in this case, we accept only 1 child for the header, is not necessary to use `React.Children.only` because I'm using `Array.find` to get the child element, that means that we are going to have only `1` in all cases.

https://github.com/x-team/auto-ui/blob/c2d978fd6dc13040b9b37b93c034757ede17fa4e/src/components/ui/Collapsible/Container.js#L57

I'm using `isValidElement` to avoid an error when we don't have a child element and also to check that is a valid Element.

makes sense for you?

src/components/ui/Collapsible/Container.js
migueliriano: Good question :), as you may know, we are setting the `isCollapsed` with the prop `initialCollapsed`, if for some reason this component is re-render and the `initialCollapsed` changes, we also want to reflect this change for the state `isCollapsed`.

Let me know if is clear or if you have any other question about this.

src/components/ui/Collapsible/Header.js
KamWis: I know you use find, but that's a knowledge outside of this component. What if this Header ill be used outside the Collapsible?

src/components/ui/Collapsible/Collapsible.stories.js
migueliriano: Comment addressed on this commit: c292d15727451da792a69cab6a8a3d403fb76183

src/components/ui/Collapsible/Container.js
KamWis: I'm just concerned about "for some reason" part :) isn't that fall into the YAGNI category? WDYT?

migueliriano: That happened before in a component that we created using this approach of setting the `state` using a `prop`. So I think this is not into the YAGNI category.

KamWis: ok then :)

src/components/ui/Collapsible/Header.js
migueliriano: `isValidElement` can handle this case.

When you have multiple `child` elements like this:

```
    <Collapsable.Header>
      <ExampleHeader text="testing"/>
      <ExampleHeader text="testing 2"/>
    </Collapsable.Header>
```
`isValidElement` going to return `false` because this is not a valid Elements, the value will be an `Array<Element>` and this only accept as valid a `Element`.

migueliriano:
Hey @bernardodiasc - Thank you, I addressed your comments on this commit: d99a2f116c13e9c4481cc695a81ab4ef52306227

>I'm not sure if this is correct, the "initial collapsed" is actually initially expanded, while the default is always collapsed. I guess this should be the opposite, right?

Yes, I think the name is confused, I rename this with the prop `isExpanded`, I think this name is more clear. 

> The other problem is a console error being logged of a prop being injected in an HTML element, which is invalid. I tried to quickly find the reason for this but it wasn't clear to me what's causing that, maybe something we would like to prevent.

The reason for this is that the `HTML element` is inheriting the prop `isCollapsed` as an attribute, and that is not a valid attribute for the HTML tag, this is how looks in the react console: `<p isCollapsed={false}> </p>`, as you can notice `isCollapsed` is not a valid attribute, to fix this I added a validation that the child element should be a `function`.

If you have any comment please let me know.

Thanks.

KamWis:
@migueliriano what if DOM element is used in a Header? It wont render, right? If that's the case - are we sure we want to handle that like this? Will that not be a annoying experience to work with? I would rather have warning then component not showing. WDYT?

migueliriano:
@KamWis -Yes, we can render DOM elements in the header, If we use a **DOM** element in the header we are not passing the `props` to avoid the warning in the console. Please see the stories: 

- https://github.com/x-team/auto-ui/pull/347/files#diff-ae48b6cc2565c528b0d2dff45e1d908aR96
- https://github.com/x-team/auto-ui/pull/347/files#diff-ae48b6cc2565c528b0d2dff45e1d908aR106

Let me know if this answers your question and any concern.

KamWis:
ah, you're right. You changed the implementation from return to props.

nicksp:
Seems like we forgot to expose that component in `index.js` file or did I miss that?

KamWis:
@nicksp can we expose all introduced components during a PR for `XP-2711-advanced-search-sprint-1`?

nicksp:
Since this is merged and closed now, we'll have to do that, otherwise, we won't be able to request components from `auto` 😏 Let's please keep that in mind what we'll need to expose in https://github.com/x-team/auto-ui/blob/master/src/index.js

KamWis:
yeah. I've totally forgot we have to do that. So we need to do that both for `Collapsible` and `Tab` component. My bad. :)

PR #348, opened 2019-02-14
Create Filters component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-2792

## Description

Create new component `Filters`

## Checklist

## Steps to Test or Reproduce

1. Run `npm start`
1. Go to http://localhost:9001/?selectedKind=UI%20Components%2FFilters


src/components/ui/Filters/Filter.js
```diff
@@ -0,0 +1,58 @@
...
+import Label from './Label'
+import Collapsible from '../Collapsible'
+import SvgIcon from '../SvgIcon'
+
+import type { Element } from 'react'
+
+export type Props = {
+  label: Element<*>,
+  children: Element<*>,
+  collapsible: boolean
```
migueliriano: Can we rename this to `iscollapsible`?

I think this more readable because is used as a condition.

src/components/ui/Filters/Filter.js
```diff
@@ -0,0 +1,58 @@
...
+    display: flex
+    cursor: pointer
+  `),
+
+  triangle: cmz(`
+    padding: 0 60px 10px 0
+  `)
+}
+
+const Filter = (props: Props) => props.collapsible ? (
```
migueliriano: If we do the comment above, we need to change the variable `collapsible` for `isCollapsible` here.

src/components/ui/Filters/Heading.js
```diff
@@ -0,0 +1,56 @@
...
+
+import theme from '../../../styles/theme'
+import { typeface } from '../../../styles/typo'
+
+import type { Element } from 'react'
+
+export type Props = {
+  children: Element<*>,
+  isExpanded: boolean,
+  onClick: (event: any) => void,
```
migueliriano: Can we remove the `,` here? Seems this is the last element of the object.

src/components/ui/Filters/Heading.js
```diff
@@ -0,0 +1,56 @@
...
+
+const Heading = (props: Props) => (
+  <div onClick={props.onClick} className={cx.heading}>
+    <div className={cx.text}>
+      {props.children}
+    </div>
+    <SvgIcon
+      icon={props.isExpanded ? 'triangleup' : 'triangledown'}
+      color='grayscarpaflow'
+    />
```
migueliriano: **NOT REQUIRED CHANGE**: I see that we are repeating this behavior in different places about the logic for the triangle, I think that we need to considered move this into the `Collapsed` component later.

src/components/ui/Filters/TabHeads.js
```diff
@@ -0,0 +1,24 @@
...
+const cmz = require('cmz')
+
+const cx = {
+  tabHeads: cmz(`
+    & {
+      padding: 0 60px 20px
+      display: flex
+    }
+  `)
+}
```
migueliriano: Here we can remove the `&` wrapper because the CSS rule only for one element.

```
const cx = {
  tabHeads: cmz(
    `
      padding: 0 60px 20px
      display: flex
  `)
}
```

migueliriano:
@nicksp - Good catch, thank you.

comment addressed in this commit: 0bdf4c2df4a7f5c6083236d3bc10282a183e38ef

nicksp:
Thanks for review guys @KamWis @migueliriano Great help with progressing further! Let's get it merged and tested in a feature branch 🎉

PR #351, opened 2019-02-15
Adjust AdminScreen and Modal for TwoColumnsLayout integration

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-2833

## Description

This PR have adjustments on `AdminScreen` and `Modal` in order to make it compatible with requirements of `TwoColumnsLayout`.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-2833-adjust-adminscreen-and-modal-for-twocolumns-layout-integration` | [Auto](https://github.com/x-team/auto/pull/1822)

## Steps to Test or Reproduce

1. Go to `AdminScreen` stories
   1. Observe that the header is always fixed at the top of the page
   1. Observe that when the content heigh is too tall, it's correct scrollable
   1. Observe that when the content heigh is short it does't display scrollbars in the page
1. Go to `TwoColumnsLayout` stories composed with `AdminScreen`
   1. Observe that it doesn't overflow the page viewport anymore
   1. Observe that the layout occupy full available height space of the screen
   1. Observe that sidebar and content have independent vertical and horizontal scrolling when needed
1. Go to `ListsScreen` stories composed with `AdminScreen`
  1. Observe the this PR doesn't introduced any regression
1. Go to `Modal` stories and observe that is working as expected
   1. Clicking on overlay or in the `X` should close
   1. Observe that the `Modal` showcases also scrolls well and have it's content visible when scrolling
   1. Observe that when the `Modal` is opened the page content is not scrollable
   1. Observe that when closes the `Modal` the page content is scrollable again
   1. On `AdminScreen`, `ListsScreen` and `TwoColumnsLayout` showcases have demonstrations with `Modal` open, observe that the `Modal` behaviour is correct there

- AdminScreen
- Modal

## Screenshots

_A GIF that fully demonstrates the fix is too big to be attached here, so it will be posted in the Jira ticket. The fixes can be viewed in the stories:_

screen shot 2019-02-18 at 10 47 06

### DIFF
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -18,10 +18,6 @@ const cx = {
     flex-direction: column
   `),
 
-  adminWidModal: cmz(`
-    height: 100vh
-  `),
-
   header: cmz(`
     width: 100%
     height: 56px
@@ -29,32 +25,22 @@ const cx = {
   `),
 
   content: cmz(`
-    & {
-      flex: 1 0 0
-      overflow: auto
-      height: 100%
-    }
-
-    & > div {
-      height: inherit
-    }
+    flex: 1 0 0
+    overflow: auto
+    height: 100%
   `)
 }
 
 type Props = {
   header?: Element<*>,
   children?: Element<*>,
-  modal: {
-    onClose: Function,
-    content: Element<*>
+  modal?: {
+    onClose?: Function,
+    content?: Element<*>
   }
 }
 
-type State = {
-  isModalOpen: boolean
-}
-
-class AdminScreen extends PureComponent<Props, State> {
+class AdminScreen extends PureComponent<Props, void> {
   static defaultProps = {
     header: null,
     children: null,
@@ -63,32 +49,17 @@ class AdminScreen extends PureComponent<Props, State> {
     }
   }
 
-  state = {
-    isModalOpen: !!this.props.modal.content
-  }
-
-  componentDidUpdate (prevProps: Props) {
-    const { content: prevContent } = prevProps.modal
-    const { content } = this.props.modal
-    if (prevContent !== content) {
-      this.setState({ isModalOpen: !!content })
-    }
-  }
-
   handleModalClose = () => {
     const { modal } = this.props
-    this.setState(() => ({ isModalOpen: false }))
-    modal.onClose && modal.onClose()
+    modal && modal.onClose && modal.onClose()
   }
 
   render () {
     const { header, children, modal } = this.props
-    const { isModalOpen } = this.state
-    const adminClassNames = isModalOpen ? `${cx.admin} ${cx.adminWidModal}` : cx.admin
     return (
-      <div className={adminClassNames}>
+      <div className={cx.admin}>
         {modal && modal.content && (
-          <Modal onClose={this.handleModalClose} isOpen={isModalOpen}>
+          <Modal onClose={this.handleModalClose}>
             {modal.content}
           </Modal>
         )}
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -68,9 +68,7 @@ class AdminScreen extends PureComponent<Props, State> {
   }
 
   componentDidUpdate (prevProps: Props) {
-    const { content: prevContent } = prevProps.modal
-    const { content } = this.props.modal
-    if (prevContent !== content) {
+    if (prevProps.modal.prevContent !== this.props.modal.content) {
       this.setState({ isModalOpen: !!this.props.modal.content })
     }
   }
@@ -84,7 +82,7 @@ class AdminScreen extends PureComponent<Props, State> {
   render () {
     const { header, children, modal } = this.props
     const { isModalOpen } = this.state
-    const adminClassNames = isModalOpen ? [cx.admin, cx.adminWidModal].join(' ') : cx.admin
+    const adminClassNames = isModalOpen ? `${cx.admin} ${cx.adminWidModal}` : cx.admin
     return (
       <div className={adminClassNames}>
         {modal && modal.content && (
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -10,7 +10,6 @@ const cmz = require('cmz')
 
 const cx = {
   admin: cmz(`
-    height: 100%
     min-height: 100vh
     overflow-y: hidden
     position: relative
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -10,9 +10,12 @@ const cmz = require('cmz')
 
 const cx = {
   admin: cmz(`
+    height: 100vh
     min-height: 100vh
     overflow-y: hidden
     position: relative
+    display: flex
+    flex-direction: column
   `),
 
   adminWidModal: cmz(`
@@ -22,12 +25,19 @@ const cx = {
   header: cmz(`
     width: 100%
     height: 56px
-    position: fixed
     z-index: 10000
   `),
 
   content: cmz(`
-    padding-top: 56px
+    & {
+      flex: 1 0 0
+      overflow: auto
+      height: 100%
+    }
+
+    & > div {
+      height: inherit
+    }
   `)
 }
 
--- a/src/components/ui/AdminScreen.js
+++ b/src/components/ui/AdminScreen.js
@@ -11,26 +11,41 @@ const cmz = require('cmz')
 const cx = {
   admin: cmz(`
     height: 100%
+    min-height: 100vh
+    overflow-y: hidden
     position: relative
   `),
 
+  adminWidModal: cmz(`
+    height: 100vh
+  `),
+
   header: cmz(`
+    width: 100%
     height: 56px
-    position: relative
+    position: fixed
     z-index: 10000
+  `),
+
+  content: cmz(`
+    padding-top: 56px
   `)
 }
 
 type Props = {
   header?: Element<*>,
   children?: Element<*>,
-  modal?: {
+  modal: {
     onClose: Function,
     content: Element<*>
   }
 }
 
-class AdminScreen extends PureComponent<Props, void> {
+type State = {
+  isModalOpen: boolean
+}
+
+class AdminScreen extends PureComponent<Props, State> {
   static defaultProps = {
     header: null,
     children: null,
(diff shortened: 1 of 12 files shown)


nicksp:
@bernardodiasc Please check merge conflicts we got and rebase with latest `develop`

src/components/ui/Modal.js
```diff
@@ -57,13 +58,19 @@ const cx = {
...
+  }
+
   state = {
-    open: false
+    open: this.props.isOpen
   }
 
-  componentDidMount () {
-    this.setState({ open: true })
+  componentDidUpdate (prevProps: Props) {
```
nicksp: As of version React 16.3, the recommended way to update state in response to props changes is with the new static getDerivedStateFromProps lifecycle. (That lifecycle is called when a component is created and each time it receives new props) https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props

src/components/ui/Modal.js
```diff
@@ -57,13 +58,19 @@ const cx = {
...
+    if (prevProps.isOpen !== this.props.isOpen) {
+      this.setState({ open: this.props.isOpen })
```
nicksp: Overall, what's the reason for this logic?

src/components/ui/Modal.js
```diff
@@ -57,13 +58,19 @@ const cx = {
   `)
 }
 
-class Modal extends Component<Props, State> {
+class Modal extends PureComponent<Props, State> {
+  static defaultProps = {
+    isOpen: true
```
nicksp: Why did we make it `true` by default? Shouldn't it be closed initially? I didn't get it.

src/components/ui/AdminScreen.js
```diff
@@ -10,27 +10,51 @@ const cmz = require('cmz')
...
-    height: 100%
+    height: 100vh
+    min-height: 100vh
+    overflow-y: hidden
     position: relative
+    display: flex
+    flex-direction: column
+  `),
+
+  adminWidModal: cmz(`
```
nicksp: could you please adjust the name so it reads properly? Don't understand the meaning. Did you mean `adminWithModal`?

src/components/ui/AdminScreen.js
```diff
@@ -39,17 +63,37 @@ class AdminScreen extends PureComponent<Props, void> {
...
+    const { content } = this.props.modal
+    if (prevContent !== content) {
+      this.setState({ isModalOpen: !!content })
+    }
+  }
+
+  handleModalClose = () => {
+    const { modal } = this.props
+    this.setState(() => ({ isModalOpen: false }))
+    modal.onClose && modal.onClose()
```
nicksp: I don't think you need to check here cause `onClose` prop is marked as required. You'll get Flow warning in this case.

src/components/ui/AdminScreen.js
```diff
@@ -39,17 +63,37 @@ class AdminScreen extends PureComponent<Props, void> {
...
   }
 
+  state = {
+    isModalOpen: !!this.props.modal.content
+  }
+
+  componentDidUpdate (prevProps: Props) {
+    const { content: prevContent } = prevProps.modal
+    const { content } = this.props.modal
+    if (prevContent !== content) {
```
nicksp: As prev commnet: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props

src/components/ui/AdminScreen.js
```diff
@@ -39,17 +63,37 @@ class AdminScreen extends PureComponent<Props, void> {
     }
   }
 
+  state = {
+    isModalOpen: !!this.props.modal.content
```
nicksp: Why do we want to open the modal if the content is present? What if I still want it to be hidden? Shouldn't the modal itself handle the case of when it should be open and when closed? Seems weird that we tight modal and AdminScreen this way. Don't understand the reasons you put here, sorry.

src/components/ui/AdminScreen.js
```diff
@@ -10,27 +10,51 @@ const cmz = require('cmz')
...
+  `),
+
+  content: cmz(`
+    & {
+      flex: 1 0 0
+      overflow: auto
+      height: 100%
+    }
+
+    & > div {
```
nicksp: What if we don't use `div`s inside children and use `span` for example? How do we know that we'll have divs inside? I wouldn't be that optimistic.

bernardodiasc: I admit this is a hack. The only proper solution here is to replace the `<div>` of https://github.com/x-team/auto/blob/develop/src/client/components/WithTitle.js#L10-L15 with `<Fragment>`, but I didn't wanted to bring that in this PR to not affect even more places that it's already affecting. Wdyt?

nicksp: It's clear that having `& > div` is unsafe and assumes we're confident about the content, which isn't. Let;s find a different way.

src/components/ui/AdminScreen.js
bernardodiasc: It's the opposite, we don't want to display the modal if there's no content present. I'll take a second look if this can be done differently.

src/components/ui/AdminScreen.js
nicksp: We can at least create a use case to test what will happen if we use different elements inside. What's the purpose of this style at all? Will we get a broken UI or what? Once the implications are known we can decide what might be the solution.

src/components/ui/Modal.js
bernardodiasc: It's closed by default on UX of the implementation in the app. The standalone component need to be open by default, otherwise there's no way for us to see it in Storybook, for instance.

src/components/ui/AdminScreen.js
nicksp: > It's the opposite, we don't want to display the modal if there's no content present.

How is that possible? You have `!!` in front which will evaluate to `true` if the content is there.

```js
!!"asdasdad" === true
```

src/components/ui/Modal.js
nicksp: Didn't get it. `Storybook` is a different thing, we don't care about that. The default Prop says it's `true`, so if I don't pass it it will be open.

src/components/ui/AdminScreen.js
bernardodiasc: The solution is to use `<Fragment>` in the `WithTitle`, or move this `height: inherit` style to that div.

> What's the purpose of this style at all? Will we get a broken UI or what?

Yes, a broken UI. That's the whole point about CSS problematic stuff. I think I'll write an article about this, it's frustrating and super confusing to explain. 😓

src/components/ui/AdminScreen.js
bernardodiasc: `!!"asdasdad" === true` is expected, as well `!!undefined === false`

bernardodiasc: Do you have suggestion for a better way to declare this? 

Linter complains if I try something like:

```
isModalOpen: this.props.modal.content ? true : false
```

src/components/ui/AdminScreen.js
bernardodiasc: > We can at least create a use case to test what will happen if we use different elements inside.

Whatever element that goes inside and doesn't have set the `height`, it will break the UI.

From CSS2.1 specs:

> The percentage is calculated with respect to the height of the generated box's containing block. If the height of the containing block is not specified explicitly (i.e., it depends on content height), and this element is not absolutely positioned, the value computes to 'auto'.

I would bet this is was one of the reasons React created `<Fragment>`, because using `<div>` as wrappers as we used to do before this feature is not insignificant for the layout painting.

nicksp: Well, whatever is a solution, the current one is unsafe unfortunately as you can see. We'll get a broken UI, so I'd trust our predictions about the nature of elements we'll use inside children.

src/components/ui/AdminScreen.js
nicksp: Well, let me outline the "problem" with what you did. In most cases, this is an antipattern to copy properties to state values. It creates a second source of truth for your data, which usually leads to bugs. One source of truth is best. The second issue is that in order to reflect changes when parent's component prop changes, you'll need to create a boilerplate code to update state. That's is not looking good and leads to confusing code. Imaging you add 5 new state props that are initialized with prop values, you'll need to repeat the condition code for all of them. Not the best approach.

Components will already re-render when their props change, so there’s no need to duplicate the props as state and then try to keep it up to date.

However, it’s not an anti-pattern if you make it clear that the prop is only seed data for the component’s internally-controlled state.

Think of it this way: it’s fine if the state needs a starting value which the component will then control. Ask yourself: does this component “own” the data? Does it just need a jump-start from a prop? Those are good reasons to initialize state from a prop.

For more details about why this is not a recommended approach, check this section of docs: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#common-bugs-when-using-derived-state

src/components/ui/AdminScreen.js
bernardodiasc: The solution will affect even more places, I'm fine with that, just saying, we'll have to QA all the pages that uses `WithTitle`.

bernardodiasc: And again, it will not work by default with anything inside. It's mandatory that the element inside have the `height` set. This is characteristic of CSS, I wouldn't expect to use `& > *` here to deal with such case.

nicksp: Are you safe with the current solution you have in a PR? I don't. Can you ensure we won't get broken UI in case of elements different from `div` that might be used in `children`.  If you can ensure that won't happen, let's keep it as it is.

src/components/ui/AdminScreen.js
bernardodiasc: > you'll need to create a boilerplate code to update state

Got it, you are totally right. I believe it's not the first time I got confused with this, now I realised my mistake. I'm addressing changes to make this component simpler and not loose the benefit of showcase its UX on Storybook. Thanks for explanation.

nicksp: Glad it was helpful 👏

bernardodiasc: Yeah, I remember clearly that this was exactly one of the reasons we switched to Storybook, I guess I had my mind in Styleguidist limitations... 😜

src/components/ui/AdminScreen.js
bernardodiasc: No, I'm not happy either. I fixed in the proper way here https://github.com/x-team/auto/pull/1822

src/components/ui/Modal.js
bernardodiasc: I resolved that by using the State boilerplate in the Storybook. 👍

bernardodiasc:
@nicksp thanks for thorough code review, I appreciate that. I removed all the local states from AdminScreen and Modal, that was a mistake and solved by wrapping with a boilerplate stateful component in Storybook, on the application side its a concern of Redux.

An extra PR was opened in `auto` https://github.com/x-team/auto/pull/1822 so that this fix works as expected in the application.

src/components/ui/Modal.js
```diff
@@ -13,68 +15,51 @@ type Props = {
...
+    overflow-x: auto
+  `),
+
   close: cmz('closemodal', `
     position: absolute
     right: 12px
     top: 12px
     cursor: pointer
+    z-index: 1
+    displa
```
migueliriano: This is a typo: `displa`, we need to remove it. 

it's curious that `cmz` doesn't trigger an error when finding a wrong property like in this case

nicksp: It doesn't as well as some other css-in-js solutions. That was one of the points I suggested to use CSS Modules with css defined in a separate file so we get complete linting fir things like that 😛

bernardodiasc: Good catch Miguel 😄 

Actually some css-in-js solutions already have stylelint support :)

bernardodiasc:
By the way, I updated the stories navigation to look like this:

screen shot 2019-02-19 at 15 33 47

That's what we discussed on past call, to make the application use cases in a separated group.

I wonder if we would get benefit of grouping the stories that we made for debugging edge cases, e.g.:

screen shot 2019-02-19 at 15 36 28

There are situations that we have the regular representations and some "extreme" representations to make sure it doesn't break, if we group inside "Debug" we can make it clear for Paul and designers what's the purpose of these stories. This is not included in this PR, I'm just throwing an idea.

nicksp:
> By the way, I updated the stories navigation to look like this

That is a great improvement and smth I showed we already do for some of the components, so good to stick to this flow in the future. Thanks Bernardo!

bernardodiasc:
@nicksp yeah, much better! I used a `State` component we have available for showcases, that was all I needed to accomplish the desired effects 🎉

PR #356, opened 2019-02-20
Generic tabs fix

galileo:
**Release Type:** *Bug fix* 

Fixes https://x-team-internal.atlassian.net/browse/XP-2846

## Description

I spot the bug while I was trying to implement the Context filers with list/all tabs. The issue was that the pane was not updated while rest the page was. The issue was that it was cached in construction phase.

## Checklist

## Steps to Test or Reproduce

1. Go into storybook
1. Go into `GenericTabs/Debug` section
1. And cliick on the `Increment` button
1. You should see the value incremented in both places
 

All GenericTab instances

## Screenshots

**Before fix**

Only Tab.Header value is changing.

before-tab-fix

**After fix**

Both values are changing.

after-tab-fix

### DIFF
--- a/src/components/ui/GenericTabs/Container.js
+++ b/src/components/ui/GenericTabs/Container.js
@@ -35,8 +35,8 @@ class Container extends Component<Props, State> {
   constructor (props: Props) {
     super(props)
 
-    this.paneChildren = this.getChildrenByType(this.props.children, Pane)
-    const [ firstPane ] = this.paneChildren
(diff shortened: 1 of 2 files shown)


nicksp:
Great point @bernardodiasc did about usage of `notes` addon for Storybook. In cases like that, when we need to explain what was the bug (assuming there won't be a bug after merging changes) about, this addon is a perfect place to put explanation down.
cc @galileo

PR #362, opened 2019-03-04
Create MetaGroup component

bernardodiasc:
**Release Type:** *Non-breaking feature*

Fixes https://x-team-internal.atlassian.net/browse/XP-2892

## Description

Create MetaGroup component and add some Storybooks stories to bullet proof the component.

## Checklist

## Steps to Test or Reproduce

1. run `npm start`
1. go to http://localhost:9001/?selectedKind=UI%20Components%2FMetaGroup
1. go to http://localhost:9001/?selectedKind=UI%20Components%2FMetaGroup%2FDebug&selectedStory=responsive%20demonstration%20lots%20elements for a more debuggable showcase (check Notes informations)

- `MetaGroup` component

### DIFF
--- a/src/components/ui/MetaGroup.js
+++ b/src/components/ui/MetaGroup.js
@@ -53,8 +53,8 @@ class MetaGroup extends PureComponent<Props, void> {
           {leftSideElement}
         </div>
         <div className={cx.rightSideColumn}>
-          {rightSideElements.map((element, i) => (
-            <div className={cx.rightSideColumns} key={i}>
+          {rightSideElements.map(element => (
+            <div className={cx.rightSideColumns} key={element.key}>
               {element}
             </div>
           ))}
--- a/src/components/ui/MetaGroup.js
+++ b/src/components/ui/MetaGroup.js
@@ -1,0 +1,67 @@
+// @flow
+
+import React, { PureComponent } from 'react'
+
+import type { Element } from 'react'
+
+const cmz = require('cmz')
+
+const cx = {
+  layout: cmz(`
+    position: relative
+    display: flex
+    height: inherit
+  `),
+
+  leftSideColumn: cmz(`
+    box-sizing: border-box
+    flex: 1
+    min-width: 60%
+  `),
+
+  rightSideColumn: cmz(`
+    display: flex
+    flex-wrap: wrap
+    justify-content: flex-end
+    align-items: flex-start
+    height: 100%
+    width: 300px
+  `),
+
+  rightSideColumns: cmz(`
+    box-sizing: border-box
+    flex-shrink: 0
+  `)
+}
+
+type Props = {
+  leftSideElement: Element<*>,
+  rightSideElements: Array<Element<*>>,
+}
+
+class MetaGroup extends PureComponent<Props, void> {
+  static defaultProps = {
+    leftSideElement: null,
+    rightSideElements: []
+  }
+
+  render () {
+    const { leftSideElement, rightSideElements } = this.props
+    return (
+      <div className={cx.layout}>
+        <div className={cx.leftSideColumn}>
+          {leftSideElement}
+        </div>
+        <div className={cx.rightSideColumn}>
+          {rightSideElements.map((element, i) => (
+            <div className={cx.rightSideColumns} key={i}>
+              {element}
+            </div>
+          ))}
+        </div>
+      </div>
+    )
+  }
+}
+
+export default MetaGroup
--- a/src/components/ui/MetaGroup.js
+++ b/src/components/ui/MetaGroup.js
@@ -13,13 +13,13 @@ const cx = {
     height: inherit
   `),
 
-  leftSideColumn: cmz(`
+  mainBodyColumn: cmz(`
     box-sizing: border-box
     flex: 1
     min-width: 60%
   `),
 
-  rightSideColumn: cmz(`
+  secondaryColumns: cmz(`
     display: flex
     flex-wrap: wrap
     justify-content: flex-end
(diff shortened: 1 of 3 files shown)


src/components/ui/MetaGroup.js
```diff
@@ -0,0 +1,67 @@
...
+
+type Props = {
+  leftSideElement: Element<*>,
+  rightSideElements: Array<Element<*>>,
+}
+
+class MetaGroup extends PureComponent<Props, void> {
+  static defaultProps = {
+    leftSideElement: null,
+    rightSideElements: []
```
migueliriano: Why `rightSideElements` is `Array<Element<*>>` and `leftSideElement` not?

src/components/ui/MetaGroup.stories.js
```diff
@@ -0,0 +1,72 @@
...
+  <div style={{ margin: '60px 20px', border: '1px dashed silver' }}>
+    {children}
+  </div>
+)
+
+storiesOf('UI Components/MetaGroup', module)
+  .add('standalone default usage', () => (
+    <Sandbox>
+      <MetaGroup
+        leftSideElement={<LeftSideElement>Here would goes a bunch of filters definitions</LeftSideElement>}
```
migueliriano: Why we don't handle the filters definition as an array the same approach that we are doing for `rightSideElements`?

bernardodiasc:
@migueliriano answering your questions, the paddings should belong the the children components. After a closer look the the designs I notice that there's not a single value for the grid spacings, therefore it doesn't make sense to apply as a responsibility of MetaGroup.

Regarding your question about `leftSideElement` and `rightSideElements`, it's expected that the first will get only 1 value, while the second will get many. The structure and responsiveness pattern are based on this.

wireframe

If you check the "standalone default usage" this might become clearer, the component that should be put inside the left-side column is the filters summary, it's a component that should hold a single responsibility. On the left-side columns it's expected to have 3 different components, each one with own responsibility.

Does that answer your questions? Do you still think that need to be changed?

bernardodiasc:
hey @migueliriano, that was good feedback, I updated the showcase with a more accurate demonstration whats going on:

migueliriano:
@bernardodiasc - FYI, I updated the PR to export the `MetaGroup` component, this is the commit: 004515b0026a215ffe304306e885a3192190ca95

src/components/ui/MetaGroup.js
```diff
@@ -0,0 +1,67 @@
...
+  render () {
+    const { leftSideElement, rightSideElements } = this.props
+    return (
+      <div className={cx.layout}>
+        <div className={cx.leftSideColumn}>
+          {leftSideElement}
+        </div>
+        <div className={cx.rightSideColumn}>
+          {rightSideElements.map((element, i) => (
+            <div className={cx.rightSideColumns} key={i}>
```
KamWis: can we do `key` as the other value than `index`? `index` should be our last resort: https://reactjs.org/docs/lists-and-keys.html#keys

bernardodiasc: I'm aware of that, but I can't see much options here. do you have any suggestion?

bernardodiasc: in the storybook demos I did provided `key` in the `element` (as in `element.key`), that can be used. but not sure if that will be reliable for during the implementation, wdyt?

bernardodiasc: well, a warning will appear in the console if someone skip key prop for 2 or more elements. it will also require key prop in HTML elements. this increased the complexity of the component usage a bit in comparison to the index, but I understand where this come from. I have divided opinion on this because the index seems safe to be used on this place. I updated the code, please take a look when possible, thanks!

bernardodiasc:
@KamWis can you take a second look at this PR when you got some time? thanks!

src/components/ui/MetaGroup.js
KamWis: maybe some combination of `element` property? like `name` + `key`? Anyways - I will approve it since it's a non blocker :) Thanks m8!

bernardodiasc: the `element` can be anything, a component or a DOM element, it's unpredictable from the MetaGroup what will be the props available. i wanted to use `DisplayName`, but that doesn't seem available without by default, right?

src/components/ui/MetaGroup.js
```diff
@@ -0,0 +1,67 @@
...
+const cmz = require('cmz')
+
+const cx = {
+  layout: cmz(`
+    position: relative
+    display: flex
+    height: inherit
+  `),
+
+  leftSideColumn: cmz(`
```
nicksp: I'd suggest to not use stylistic representation in a name, since what is gonna be "left" today, might easily be "right" tomorrow, or "top" section :) In this case, we'd name to change the name and update markup with a changed class name. Not a blocker, but just wanted to mention that. I'd call smth based on the expected content for example: `filterColumn`/`actionsColumn`, or simply, `mainBodyColumn`/`secondaryColumn`

src/components/ui/MetaGroup.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  render () {
+    const { leftSideElement, rightSideElements } = this.props
+    return (
+      <div className={cx.layout}>
+        <div className={cx.leftSideColumn}>
+          {leftSideElement}
+        </div>
+        <div className={cx.rightSideColumn}>
```
nicksp: Shouldn't this be the opposite way? Since this is a parent element, it should probably be `cx.rightSideColumns` and each child should have `cx.rightSideColumn`. Am I correct?

bernardodiasc:
Good catch @nicksp, I updated the component with more meaningful prop names, thanks for the suggestions!

PR #364, opened 2019-03-05
Create ui component to show result number of element

migueliriano:

Fixes https://x-team-internal.atlassian.net/browse/XP-2883

## Do not merge
We need to wait that this PR merged into `develop` branch: https://github.com/x-team/xp-ui/pull/362

## Description
Create resultCount component to show the number of items in a result for search or any other situation that you need to show this information

## Checklist

## Related PRs

branch | PR
------ | ------
XP-2883-show-total-number-search-results| [XP](https://github.com/x-team/xp/pull/1836)

## Steps to Test or Reproduce
1- Go to http://localhost:9001/?selectedKind=UI%20Components%2FResultCount&selectedStory=basic%20usage&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel&background=%23fff

## Screenshots
resultcomponentui


bernardodiasc:
@migueliriano by the way, please make sure to consider the singular representation when it's `1 item`.

migueliriano:
@bernardodiasc - In the ticket description you are going to see this information:

> See Figma in Epic for latest designs however at time of writing it has a pending update. The correct text is eg. "52 results"

Seems the figma it's outdate with this information.

migueliriano:
@bernardodiasc - About this comment:

> @migueliriano by the way, please make sure to consider the singular representation when it's 1 item.

I think in this case the word `Results` works for both cases, singular, and plural.

Wdyt?

src/components/ui/ResultCount.js
```diff
@@ -0,0 +1,32 @@
...
+  `),
+
+  resultNumber: cmz(`
+    font-weight: 600;
+    margin-right: 5px;
+  `)
+}
+
+type Props = {
+  items?: number | Array<*> | Object
```
KamWis: what's the reason behind the `Object` type? Why would we pass an object to count it as result count?

migueliriano: Yeah, I think we don't have a case that we want to count an object, I'm going to remove it.

migueliriano: Comment addressed on this commit: 5b474ae2f25870aefbe37ddd803205ecfbc7277e

src/components/ui/ResultCount.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+type Props = {
+  items?: number | Array<*>
+}
+
+const ResultCount = ({ items }: Props) => (
+  <span className={cx.result}>
+    <span className={cx.resultNumber}>{isNumber(items) ? items : size(items)}</span>
+    Results
```
nicksp: I have same concerns as Bernardo had:

1. as per the comment
> See Figma in Epic for latest designs however at time of writing it has a pending update. The correct text is eg. "52 results"

the text should be `N results` (small `r`)

2. In case of 1 item, it should be in singular form: `1 result`
Good news that we already have `pluralize` helper function that we can use to do that easily: https://github.com/x-team/xp-ui/blob/1a3bf344e33c32b3d534405ff97974653bce7834/src/utils/helpers.js#L37-L39

migueliriano: I addressed your comment on this commit: fd51156b3e0be017d923016ea0b25358a7588854

About point number 2:

> In case of 1 item, it should be in the singular form: 1 result

I didn't use the function `pluralize` because the function returns the number and we need the number wrapper in another element to add a specific style, so to avoid increase more the responsibility of that function I added the logic directly in the `ResultCount` component.

migueliriano:
@nicksp - Comments addressed on this commit: 4ed515c2ef28551cae3bfb146874dc44c4d022f1

nicksp:
Changed base branch for the PR to a `develop` and rebased to get actual changes only.

src/components/ui/ResultCount.js
```diff
@@ -0,0 +1,35 @@
...
+  `)
+}
+
+type Props = {
+  items?: number
+}
+
+const ResultCount = ({ items }: Props) => {
+  const countItems = isNumber(items) ? items : 0
+  const resultText = parseInt(countItems, 10) > 1 ? 'results' : 'result'
```
nicksp: I suggest to change the condition here because in case of `0` it should be plural as well, not singular as it is now:
```suggestion
  const resultText = parseInt(countItems, 10) !== 1 ? 'results' : 'result'
```

image

src/components/ui/ResultCount.js
```diff
@@ -0,0 +1,35 @@
...
+    margin-right: 5px;
+  `)
+}
+
+type Props = {
+  items?: number
+}
+
+const ResultCount = ({ items }: Props) => {
+  const countItems = isNumber(items) ? items : 0
```
nicksp: I suggest to slightly change the order of words in the name, because when a variable starts with a verb, it naturally means an action rather than a container for a variable.
```suggestion
  const itemsCount = isNumber(items) ? items : 0
```

src/components/ui/ResultCount.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+
+import React from 'react'
+import isNumber from 'lodash.isnumber'
+
+const cmz = require('cmz')
+
+const cx = {
+  result: cmz(`
```
nicksp: suggest to rename to more generic class names like `wrapper` and `content` for example, since we're already in the scope of the component with name `ResultCount` so it's clear what it's all about :) Also it's easier to scale in case we decide to render smth different in that container. We won't need to rename classes in this case and the component looks much more generic and abstracted away from the knowledge of its content. That a good rule of thumb to stick to when selecting names. Sometimes having generic names is really helpful. LMK if any concern!

PR #366, opened 2019-03-13
Create new UI to select the columns of the datagrid and be displayed above the datagrid

marcoacfilho:
**Release Type:** *non breaking changes* 

Fixes https://x-team-internal.atlassian.net/browse/XP-2800

## Description

Created a new UI component based on the already existing SelectBox/Dropdown component, styled to be used above the DataGrid. This component is used to choose what columns of the DataGrid need to be displayed.

## Checklist

## Steps to Test or Reproduce

In the Storybook search field type ColumnsCustomizer, you will see the option to visualize the created new UI, click on the sub menu "basic usage", you will se the UI component being shown on the viewport.

- `ColumnsCustomizer.js`
- `ColumnsCustomizer.stories.js`

## Screenshots

image

image

### DIFF
--- a/src/components/ui/ColumnsCustomizer.js
+++ b/src/components/ui/ColumnsCustomizer.js
@@ -17,7 +17,7 @@ const styles = {
   columnsDropdown: cmz(
     typo.formText,
     `
-    & { 
+    & {
       background-color: ${theme.baseBright};
       border: 1px solid ${theme.lineSilver2};
       border-radius: 2px;
@@ -41,9 +41,7 @@ type Props = {
   label?: Element<*> | string
 }
 
-const ColumnsCustomizer = (props: Props) => {
-  const { items, onSelect, width, visibleItems, label } = props
-
+const ColumnsCustomizer = ({ items, onSelect, width, visibleItems, label }: Props) => {
   return (
     <Dropdown className={styles.columnsDropdown} targetXOrigin='right' label={label} indicator padded>
       <SelectBox
--- a/src/components/ui/ColumnsCustomizer.js
+++ b/src/components/ui/ColumnsCustomizer.js
@@ -9,6 +9,7 @@ import theme from '../../styles/theme'
 import typo from '../../styles/typo'
 
 import type { Element } from 'react'
+import type { Item } from './SelectBox'
 
 const cmz = require('cmz')
 
@@ -37,18 +38,6 @@ const styles = {
   )
 }
 
-type Status = '' | 'selecting' | 'editing' | 'saving' | 'edited' | 'creating' | 'created' | 'confirm' | 'deleting' | 'deleted' | 'dismissed' | 'archiving' | 'archived' | 'unarchiving' | 'unarchived'
-
-type Item = {
-  id: number,
-  value: string,
-  selected?: boolean,
-  archived?: boolean,
-  editing?: string | null,
-  hidden?: boolean,
-  status?: ?Status
-};
-
 type Props = {
   items?: Array<Item>,
   onSelect?: Function,
@@ -59,8 +48,10 @@ type Props = {
 
 class ColumnsCustomizer extends PureComponent<Props> {
   static defaultProps = {
-    label: '',
-    items: []
+    label: 'Columns',
+    items: [],
+    width: 250,
+    visibleItems: 8
   };
 
   render () {
--- a/src/components/ui/ColumnsCustomizer.js
+++ b/src/components/ui/ColumnsCustomizer.js
@@ -27,6 +27,7 @@ const styles = {
       border-radius: 2px;
       padding-left: 10px;
       padding-right: 8px;
+      margin-left: 22px;
     }
 
     & > :nth-child(2n) {
--- a/src/components/ui/ColumnsCustomizer.js
+++ b/src/components/ui/ColumnsCustomizer.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { PureComponent } from 'react'
+import React from 'react'
 
 import Dropdown from './Dropdown'
 import SelectBox from './SelectBox'
@@ -20,10 +20,6 @@ const styles = {
     & { 
       background-color: ${theme.baseBright};
       border: 1px solid ${theme.lineSilver2};
-      color: ${theme.typoHighlightOnDarkBackground};
-      font-weight: normal;
-      text-rendering: optimizeLegibility;
-      -webkit-font-smoothing: subpixel-antialiased;
       border-radius: 2px;
       padding-left: 10px;
       padding-right: 8px;
@@ -43,34 +39,32 @@ type Props = {
   width?: number,
   visibleItems?: number,
   label?: Element<*> | string
-};
+}
 
-class ColumnsCustomizer extends PureComponent<Props> {
-  static defaultProps = {
-    label: 'Columns',
-    items: [],
-    width: 250,
-    visibleItems: 8
-  };
+const ColumnsCustomizer = (props: Props) => {
+  const { items, onSelect, width, visibleItems, label } = props
 
-  render () {
-    const { items, onSelect, width, visibleItems, label } = this.props
+  return (
+    <Dropdown className={styles.columnsDropdown} targetXOrigin='right' label={label} indicator padded>
+      <SelectBox
+        items={items}
+        hasSearch={false}
+        size='small'
+        expanded
+        width={width}
+        visibleItems={visibleItems}
+        shouldSortItems={false}
+        onSelect={onSelect}
+      />
+    </Dropdown>
+  )
+}
 
-    return (
-      <Dropdown className={styles.columnsDropdown} targetXOrigin='right' label={label} indicator padded>
-        <SelectBox
-          items={items}
-          hasSearch={false}
-          size='small'
-          expanded
-          width={width}
-          visibleItems={visibleItems}
-          shouldSortItems={false}
-          onSelect={onSelect}
-        />
-      </Dropdown>
-    )
-  }
+ColumnsCustomizer.defaultProps = {
+  label: 'Columns',
+  items: [],
+  width: 250,
+  visibleItems: 8
 }
 
 export default ColumnsCustomizer
--- a/src/components/ui/ColumnsCustomizer.js
+++ b/src/components/ui/ColumnsCustomizer.js
@@ -1,16 +1,16 @@
 // @flow
 
-import React, { PureComponent } from 'react';
+import React, { PureComponent } from 'react'
 
-import Dropdown from './Dropdown';
-import SelectBox from './SelectBox';
+import Dropdown from './Dropdown'
+import SelectBox from './SelectBox'
 
-import theme from '../../styles/theme';
-import typo from '../../styles/typo';
+import theme from '../../styles/theme'
+import typo from '../../styles/typo'
 
-import type { Element } from 'react';
+import type { Element } from 'react'
 
-const cmz = require('cmz');
+const cmz = require('cmz')
 
 const styles = {
   columnsDropdown: cmz(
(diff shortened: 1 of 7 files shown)


src/components/ui/ResultCount.js
```diff
@@ -22,7 +22,7 @@ type Props = {
 
 const ResultCount = ({ items }: Props) => {
   const itemsCount = isNumber(items) ? items : 0
-  const resultText = parseInt(itemsCount, 10) !== 1 ? 'results' : 'result'
```
bernardodiasc: I think this is out of the current scope. It was correct with `results` instead `items`. A little confusion with outdated mocks led to this, I suggest revert the change and leave as it was. :)

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,86 @@
...
+
+    & > :nth-child(2n) {
+      right: 140px;
+      top: 45px;
+    }
+  `
+  )
+}
+
+type Status = '' | 'selecting' | 'editing' | 'saving' | 'edited' | 'creating' | 'created' | 'confirm' | 'deleting' | 'deleted' | 'dismissed' | 'archiving' | 'archived' | 'unarchiving' | 'unarchived'
```
bernardodiasc: Maybe it would work better if we export the Status type from the original component (https://github.com/x-team/xp-ui/blob/develop/src/components/ui/SelectBox.js#L452) and import in this component, wdyt?

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,86 @@
...
+
+type Item = {
```
bernardodiasc: Same here, we could export the type declaration from original source and reuse it here (https://github.com/x-team/xp-ui/blob/develop/src/components/ui/SelectBox.js#L454-L462).

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,86 @@
...
+type Props = {
+  items?: Array<Item>,
+  onSelect?: Function,
+  width?: number,
+  visibleItems?: number,
+  label?: Element<*> | string
+};
+
+class ColumnsCustomizer extends PureComponent<Props> {
+  static defaultProps = {
```
bernardodiasc: I left a suggestion here https://github.com/x-team/xp/pull/1841#discussion_r265073122 to declare `width`, `visibleItems`,  and `label` props with default values. These concerns are related to UI and maybe don't really need to be a prop at all. But regardless of being a prop or not, we could use these values as default ones.

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,86 @@
...
+    label: '',
+    items: []
+  };
+
+  render () {
+    const { items, onSelect, width, visibleItems, label } = this.props
+
+    return (
+      <Dropdown className={styles.columnsDropdown} label={label} indicator padded>
```
bernardodiasc: Also, I noticed that you included some custom styles for `Dropdown`, maybe that wasn't neede. Dropdown component already have a prop that you can set `targetXOrigin='right'` and I guess it will result as expected for this component. Please take a look if that helps :)

src/components/ui/ResultCount.js
marcoacfilho: Done

src/components/ui/ColumnsCustomizer.js
marcoacfilho: Done

src/components/ui/ColumnsCustomizer.js
marcoacfilho: targetXOrigin='right' doesn't work for what we need to style here unfortunately so the styling for Dropdown as coded is needed

src/components/ui/ColumnsCustomizer.js
marcoacfilho: Done

src/components/ui/ColumnsCustomizer.js
marcoacfilho: Done

nicksp:
@bernardodiasc I saw you requested changes the first time. More updates have been added since then. Could you please give it a final review?

src/components/ui/SelectBox.js
```diff
@@ -449,9 +449,9 @@ const cx = {
   `)
 }
 
-type Status = '' | 'selecting' | 'editing' | 'saving' | 'edited' | 'creating' | 'created' | 'confirm' | 'deleting' | 'deleted' | 'dismissed' | 'archiving' | 'archived' | 'unarchiving' | 'unarchived'
+export type Status = '' | 'selecting' | 'editing' | 'saving' | 'edited' | 'creating' | 'created' | 'confirm' | 'deleting' | 'deleted' | 'dismissed' | 'archiving' | 'archived' | 'unarchiving' | 'unarchived'
```
nicksp: That's a good point to reuse the type instead of re-declaring it in other components 👍

src/components/ui/MetaGroup.stories.js
```diff
@@ -19,8 +20,12 @@ const SecondaryElement = ({ children }) => (
   </div>
 )
 
-const sampleSecondaryElements = Array(20).fill('Right-side').map((each, i) => (
-  <SecondaryElement key={`rightside-${i}`}>{each}</SecondaryElement>
+const sampleSecondaryElements = Array(10).fill('').map((each, i) => (
+  <SecondaryElement key={`rightside-${i}`}>
+    {Array(faker.random.number(2) + 1).fill('').map((e, ind) => (
```
nicksp: Was it done on purpose to restructure the test cases for that component? How does it relate to the current ticket?

src/components/ui/ColumnsCustomizer.stories.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+
+import ColumnsCustomizer from './ColumnsCustomizer'
+
+storiesOf('UI Components/ColumnsCustomizer', module).add('basic usage', () => {
```
nicksp: Let's move `.add` on the next line to keep it consistent with other stories and make it easier to read the code. Besides, we also tend to add to each of story cases a story `missing props (does component explode?)` to see whether component behaves correctly when used with no props. Please add that as well 😏

src/components/ui/ColumnsCustomizer.stories.js
```diff
@@ -0,0 +1,24 @@
...
+  const itemsArray = [
+    { id: 1, value: 'item-1', selected: true },
+    { id: 2, value: 'item-2', selected: true },
+    { id: 3, value: 'item-3', selected: true },
+    { id: 4, value: 'item-4', selected: true },
+    { id: 5, value: 'item-5', selected: true },
+    { id: 6, value: 'item-6', selected: true }
+  ]
+  return (
+    <div style={{ right: 0, position: 'absolute' }}>
```
nicksp: Could you please position the component in the top left instead of top right? It's really hard to play around with that as it's covered by the badge and a scrollbar:

image

It's better to make it easily accessible for testing purposes.

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,76 @@
...
+
+type Props = {
+  items?: Array<Item>,
+  onSelect?: Function,
+  width?: number,
+  visibleItems?: number,
+  label?: Element<*> | string
+};
+
+class ColumnsCustomizer extends PureComponent<Props> {
```
nicksp: Since this is now stateless and doesn't rely upon lifecycle hooks, I suggest we convert that to a functional component instead. The main benefit among other small nice things is that the final transpiled code takes less bytes. Lots of boilerplate go into creating a class. Compare https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECCAO9oG8BQ0BOBTAdgEywwAoBKFadaTLAFwFcMdoc6QR0BfVDoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.3.3 (class transpilation) to https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAggBwTAvDBAnECKoHwwAUAUDADwAmAlgG55gCuANk2QPTV3ECUQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.3.3 (function representation)

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,76 @@
...
+  static defaultProps = {
+    label: 'Columns',
+    items: [],
+    width: 250,
+    visibleItems: 8
+  };
```
nicksp: Not sure why linter didn't catch it up 🤔  , but we don't use semicolons `;`. Please remove them from the file.

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,76 @@
...
+    & > :nth-child(2n) {
+      top: calc(100% + 10px)
+    }
+  `
+  )
+}
+
+type Props = {
+  items?: Array<Item>,
+  onSelect?: Function,
```
nicksp: Can we be more explicit with expected type here? Can we use a function signature instead?

According to [Flow docs](https://flow.org/en/docs/types/functions/#toc-function-type):
> However, if you need to opt-out of the type checker, and don’t want to go all the way to any, you can instead use Function. Function is unsafe and should be avoided.

src/components/ui/ColumnsCustomizer.js
```diff
@@ -0,0 +1,76 @@
...
+
+const styles = {
+  columnsDropdown: cmz(
+    typo.formText,
+    `
+    & { 
+      background-color: ${theme.baseBright};
+      border: 1px solid ${theme.lineSilver2};
+      color: ${theme.typoHighlightOnDarkBackground};
+      font-weight: normal;
```
nicksp: These 4 props (`color`, `font-weight`, `text-rendering`, and `font-smoothing`) can be removed here as they've been declared in outer components:

image

bernardodiasc:
@nicksp regarding this comment

> Also, I've noticed that the cursor isn't as expected when you hover over the checkbox and the label. We should fix that:

I believe this is not what expected for this task, that effect on mouse hover is expected for the https://x-team-internal.atlassian.net/browse/XP-2889 feature.

nicksp:
@bernardodiasc 

> I believe this is not what expected for this task, that effect on mouse hover
You didn't get me. I said the cursor is wrong, not the style effect for the element underneath.

src/components/ui/ColumnsCustomizer.stories.js
bernardodiasc: I think it would be better to add a `top: 50px` in this case.

nicksp: i'd expect to see that in the clean space of the screen: top right. So please add there.

src/components/ui/MetaGroup.stories.js
marcoacfilho: Yes exactly that was done to break the lines so we could have the 3 elements on the right with different heights, but even with different heights they should be vertically aligned like they are now due to my changes in the CSS

src/components/ui/ColumnsCustomizer.stories.js
marcoacfilho: Done

src/components/ui/ColumnsCustomizer.stories.js
marcoacfilho: Done

src/components/ui/ColumnsCustomizer.js
marcoacfilho: Done

src/components/ui/ColumnsCustomizer.js
marcoacfilho: Done

src/components/ui/ColumnsCustomizer.js
marcoacfilho: This is the same as in the SelectBox Props, I think if I change it there I have to change it in the SelectBox component as well

nicksp:
Sure. In this case let’s keep it like that.



Sent from myMail for iOS


Friday, 15 March 2019, 23:28 +0300 from notifications@github.com  <notifications@github.com>:

src/components/ui/ColumnsCustomizer.js
marcoacfilho: Done

PR #367, opened 2019-03-14
included size props to InputFields to be used in the SideBar Advancer…

marcoacfilho:
… Search as per design and fixed date input background

Fixes https://x-team-internal.atlassian.net/browse/XP-2916

## Description

Included new props to InputField. The size props has so far 2 options 'small' or 'normal' (default), the small option has now a new styling to be used in the SideBar menu of the Advanced Search and the normal option is the already existing styling.
The background-color of the date input was transparent which was not how it is supposed to be, now the background-color was changed to white.

## Checklist

## Steps to Test or Reproduce

1. Open Storybook
2. Search for InputField
3. Open the option 'standard use (text input element small)'
4. Open the option 'Use date field (small)'

- `forms/InputField.js`
- `forms/InputField.stories.js`

## Screenshots

image

image

### DIFF
--- a/src/components/forms/InputField.js
+++ b/src/components/forms/InputField.js
@@ -151,22 +151,22 @@ const errorInput = cmz(`
 
 const dateInput = cmz(`
   & {
-    position: relative;
+    position: relative
   }
 
   & input {
-    height: 50px;
+    height: 50px
     background-color: ${theme.baseBrighter} !important
   }
 
   & input::-webkit-clear-button {
-    margin-right: 15px;
-    z-index: 5;
-    cursor: pointer;
+    margin-right: 15px
+    z-index: 5
+    cursor: pointer
   }
 
   & input::-webkit-inner-spin-button {
-    display: none;
+    display: none
   }
 
   & input::-webkit-calendar-picker-indicator {
@@ -176,39 +176,39 @@ const dateInput = cmz(`
     height: 100%
     color: transparent
     background: transparent
-    z-index: 2;
+    z-index: 2
   }
 
   & > svg {
-    position: absolute;
-    transform: translateY(-50%);
-    height: 15px;
-    width: 15px;
-    right: 10px;
-    top: 50%;
-    z-index: 1;
+    position: absolute
+    transform: translateY(-50%)
+    height: 15px
+    width: 15px
+    right: 10px
+    top: 50%
+    z-index: 1
   }
 `)
 
 const dateInputSmall = cmz(`
   & {
-    position: relative;
+    position: relative
   }
 
   & input {
-    height: 40px; !important
+    height: 40px !important
     font-size: 1rem !important
     background-color: ${theme.baseBrighter} !important
   }
 
   & input::-webkit-clear-button {
-    margin-right: 15px;
-    z-index: 5;
-    cursor: pointer;
+    margin-right: 15px
+    z-index: 5
+    cursor: pointer
   }
 
   & input::-webkit-inner-spin-button {
-    display: none;
+    display: none
   }
 
   & input::-webkit-calendar-picker-indicator {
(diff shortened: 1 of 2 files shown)


src/components/forms/InputField.js
```diff
@@ -110,6 +111,33 @@ const inputStyles = [
...
+const inputStylesSmall = [
+  typo.formText,
+  cmz(`
+    & {
+      position: relative
+      display: table-cell
+      margin: 0
+      outline: none
+      width: 100%
+      height: 40px !important
```
nicksp: Is it really important to use `!important` here?

src/components/forms/InputField.js
```diff
@@ -128,7 +156,49 @@ const dateInput = cmz(`
 
   & input {
     height: 50px;
-    background: transparent;
+    background-color: ${theme.baseBrighter} !important
```
nicksp: Could we be more consistent please with the usage of `;`? I can see that in some cases you use that and in others don't. Try to stick to what is being used in that particular file for other styles declarations.

src/components/forms/InputField.js
```diff
@@ -128,7 +156,49 @@ const dateInput = cmz(`
...
+  }
+`)
+
+const dateInputSmall = cmz(`
+  & {
+    position: relative;
+  }
+
+  & input {
+    height: 40px; !important
```
nicksp: This style is broken due to misplaced `;`

image

src/components/forms/InputField.js
```diff
@@ -128,7 +156,49 @@ const dateInput = cmz(`
...
+    font-size: 1rem !important
```
nicksp: this can be removed as per screenshot above

src/components/forms/InputField.js
```diff
@@ -274,9 +344,9 @@ const customTypesDefinitions: Object = {
 
 const getFinalType = type => customTypesDefinitions[type] || type
 
-const inputFactory = type => {
+const inputFactory = (type, size) => {
```
nicksp: we can provide a default value for the `size` right in the signature for the arguments: `(type, size = 'normal')` to ensure it's `normal` in case it's not passed.

src/components/forms/InputField.js
```diff
@@ -356,7 +428,7 @@ class InputField extends PureComponent<Props> {
     if (type === 'textarea') {
       const props = {
         ...baseProps,
-        className: `${inputStyles.join(' ')} ${textareaStyles} ${errorClassName} ${spacingClassName}`,
+        className: `${size === 'normal' ? inputStyles.join(' ') : inputStylesSmall.join(' ')} ${textareaStyles} ${errorClassName} ${spacingClassName}`,
```
nicksp: I can see we repeat this logic quite a lot to check whether it's a `normal` size. Could we abstract this away into a getter method? So in case we need to change the value, we do it once in one place only.

src/components/forms/InputField.js
```diff
@@ -128,7 +156,49 @@ const dateInput = cmz(`
...
+
+const dateInputSmall = cmz(`
+  & {
+    position: relative;
+  }
+
+  & input {
+    height: 40px; !important
+    font-size: 1rem !important
+    background-color: ${theme.baseBrighter} !important
```
nicksp: Is it really important to use !important here? Seems like while is the default background color for inputs, and we don't even need this declaration. Please clarify that.

src/components/forms/InputField.js
marcoacfilho: Yes, unfortunately this is needed, I discussed that with Bernardo already

src/components/forms/InputField.js
marcoacfilho: Sure. We might want to review some of the components though because this inconsistency was also in some of the CSSs that were already in this code from past PRs. But I fixed it for this component

src/components/forms/InputField.js
marcoacfilho: Fixed

src/components/forms/InputField.js
marcoacfilho: Done

src/components/forms/InputField.js
marcoacfilho: As said it was a con-sense between Bernardo and I as I showed him the code while doing this CSS

src/components/forms/InputField.js
marcoacfilho: done

src/components/forms/InputField.js
bernardodiasc: We still have some global styles in `xp` for input elements. This is needed to override the global styles.

PR #368, opened 2019-03-18
Create ListExclusionFormPopup component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-2928

## Description

Creates the `ListExclusionFormPopup` component which is a clone from `xp/ListExclusionForm` with some improvements and the popup effect.

## Checklist

## Steps to Test or Reproduce

1. run `npm start`
1. go to http://localhost:9001/?selectedKind=UI%20Components%2FListExclusionFormPopup

- A new UI component `ListExclusionFormPopup` was created
- Included new prop `scrollableSidebar` at `TwoColumnsLayout`

### Standalone example

> - Applicant name in the header is fixed
> - The form is scrollable
> - The popup have max height of 400px
> - Clicking on radio buttons will fill up the comment box
> - Changing comment box to something else will uncheck any radio
> - Send button is disabled by default and will be enabled once any comment exists

> - Flexible heigh for big applicant names

### Small viewport height

> - Adjusts popup max height to fit in the viewport
> - The triangle aligns with the middle of the click target
> - If the click target is out of the viewport top boundary (including the popup top margin), align the triangle at the top of the popup
> - If the click target is out of the viewport bottom boundary (including the popup bottom margin), align the triangle at the bottom of the popup

> - Cancel button and clicking outside hides the popup (the state that handles this effect belongs to parent component or redux)
> - The scrolling of the element bellow is prevents when the popup is opened (this behaviour relies on the parent/global state and the capabilities of the scrollable component)
> - Custom top and bottom margins can be provided for the popup (in the demo below its 70px in the top and 10px in the bottom)

### Bigger viewport height

> - By default, the top of the popup will be aligned to the top of click target and the triangle will be aligned to the middle of the click target
> - If the click target is out of the viewport top boundary (including the popup top margin), align the triangle at the top of the popup
> - If the click target is out of the viewport bottom boundary (including the popup bottom margin), align the triangle at the bottom of the popup
> - In the case that the popup can't be aligned with the top of the click target without going out the viewport bottom boundary, align the popup to the bottom of the viewport (popup bottom margin included) and align the triangle in the middle of the click target


src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,284 @@
...
+    window.removeEventListener('resize', this.updateDimensions)
+  }
+
+  render () {
+    const { applicant, reasons } = this.props
+    const { reasonIndex, reasonLabel, anchor, style } = this.state
+    return (
+      <ClickOutside onClickOutside={this.handleCancel}>
+        <form className={cx.wrapper} style={style}>
+          <style dangerouslySetInnerHTML={{ __html: `
```
marcoacfilho: Do we have to use `dangerouslySetInnerHTML`? If we could maybe find a way to use something else I think would be better.

bernardodiasc: @marcoacfilho do you have some suggestion here? I'm not aware of an alternative approach but would love to learn. :)

bernardodiasc:
> Maybe we can trim some of the margin-top from the Popup Title, see that the margin-top is bigger than the margin-bottom.

Fixed. The margins was equal but that heading has `1.2` line-height, which adds a bit more space. 

> The send icon is not centralized, is this how it is supposed to look like ?

Nice catch, that's fixed as well.

> I clicked on the very bottom "Click Here #11" that is just barely showing, and the Popup arrow seems to be pointing to the "Click Here #10", I understand the size constrains so maybe we should prevent the user from clicking on a Button that is not showing at least half of it.

About this one, I didn't changed current behaviour. I think what you said makes sense but it's not clear what would be the best solution. I don't like to disable the click event as it could feel like a bug or something strange to happen from the user perspective, but thats just my opinion.

The current logic uses a simple principle, where the triangle is aligned in the edge of the popup when the top of the click target locates above the top of the popup element (the same for the bottom edges if the cases where popup would appear at the bottom boundaries of viewport):

Screen Shot 2019-03-21 at 12 06 29
Screen Shot 2019-03-21 at 12 06 49
Screen Shot 2019-03-21 at 12 07 07
Screen Shot 2019-03-21 at 12 07 18
Screen Shot 2019-03-21 at 12 07 26
Screen Shot 2019-03-21 at 12 08 13

The implementation allows to customise the margins top and bottom of the popup, that margin relates to the viewport, not the scrollable place where the click target is located. Bellow I reduced the top margin from 70px to 50px, the issue you spot happens differently. I mean, it makes to be not an issue anymore, but on other hand it doesn't look so good.

Screen Shot 2019-03-21 at 12 09 14
Screen Shot 2019-03-21 at 12 09 24
Screen Shot 2019-03-21 at 12 09 30

I'm open to suggestions. If everyone feels that disabling the click makes sense, then we need to find a more specific situation where to disable the click: 

e.g. when the bottom edge of click target is above of the middle of triangle when the triangle is aligned on the top edge of the popup element, and the same principle for the bottom alignment...

yeah, that can be pretty tricky to accomplish, do we really want to handle such use case?

src/components/ui/ListExclusionFormPopup.js
marcoacfilho: @bernardodiasc there is a library called `html-react-parser` see the link https://medium.com/@remarkablemark/an-alternative-to-dangerously-set-innerhtml-64a12a7c1f96 but if it is too much trouble just leave it as it is

nicksp: I think it's valid to use `dangerouslySetInnerHTML` in this case, so we can keep it as is.

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  state = {
+    reasonIndex: noneIndex,
+    reasonLabel: '',
+    anchor: 4,
+    style: {}
+  }
+
+  handleCancel = (event: Object) => {
```
nicksp: Could we add a better typing to `event` objects please, smth like: `SyntheticEvent<HTMLInputElement>` should work.

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+    this.setState(
+      {
+        reasonIndex: noneIndex,
+        reasonLabel: ''
+      },
+      () => onCancel && onCancel()
+    )
+  }
+
+  handleSubmit = (event: Object) => {
```
nicksp: same here with a type

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+    event.preventDefault()
+    const { onSubmit } = this.props
+    const { reasonLabel } = this.state
+    onSubmit && onSubmit(reasonLabel)
+  }
+
+  handleReasonChange = (event: Object) => {
```
nicksp: Please check all types across the change set

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+    const margins = marginTop + marginBottom
+    const triangleHeight = 20
+    const initialAnchor = (height / 2) - (triangleHeight / 2)
+
+    let anchor = initialAnchor > 0 ? initialAnchor : 0
+
+    const style = {
+      height: maxHeight,
+      maxHeight: `calc(100vh - ${margins}px)`,
+      top: top,
```
nicksp: can be just `top`, since both the prop name and value are the same

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+      left: left + width + 12
+    }
+
+    // Small screens than popup max height
+    if (innerHeight <= maxHeight + margins) {
+      style.top = marginTop
+      anchor = bottom > innerHeight - marginBottom
+        ? innerHeight - margins - triangleHeight
+        : (top < marginTop ? 0 : top + initialAnchor - marginTop)
+
```
nicksp: seems like redundant line, we can remove that one

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+    this.setState({
+      anchor,
+      style
+    })
+  }
+
+  componentDidMount () {
+    this.updateDimensions()
+    window.addEventListener('resize', this.updateDimensions)
```
nicksp: `resize` is very high-rate event. In short, when resizing a browser window, it can emit many resize events while dragging the resize handle. We should improve performance by debouncing such events, because we are only interested on the final value, after we stops resizing the browser.

src/components/ui/ListExclusionFormPopup.stories.js
```diff
@@ -0,0 +1,122 @@
...
+      {({ setState, state }) => (
+        <Body>
+          {state.isOpen ? (
+            <StoryListExclusionFormPopup
+              applicant={state.applicant}
+              positioning={state.positioning}
+              onCancel={() => setState({ isOpen: false })}
+              onSubmit={() => setState({ isOpen: false })}
+            />
+          ) : null}
```
nicksp: I think, this can be shorted to just: 

```js
{state.isOpen && (
  <div>...</div>
)}
```

`null` isn't required in this case

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+          <h1 className={cx.applicant}>{applicant}</h1>
+          <div className={cx.container}>
+            {reasons.map((value, index) => (
+              <div className={cx.input} key={index} data-testid='exclusion-reasons'>
+                <InputField
+                  name='reasons'
+                  type='radio'
+                  label={value}
+                  value={index}
+                  checked={Number(reasonIndex) === index}
```
nicksp: why do you need to use: `Number(reasonIndex)`, `reasonsIndex` is number as I can see from Flow type, so you'll be warned by Flow if that's not the case.

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+              type='textarea'
+              name='reason'
+              label='Add a comment'
+              placeholder='Provide a reason'
+              value={reasonLabel}
+              onChange={this.handleCommentChange}
+              data-testid='exclusion-comment'
+            />
+            <div className={cx.buttons}>
+              <Button size='normal' pseudolink onClick={this.handleCancel} data-testid='exclusion-cancel'>
```
nicksp: Is that important to provide `size` prop as `normal`? This is a default value so should be used if the prop isn't provided. Or am I missing anything?

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+                Cancel
+              </Button>
+              <Button size='normal' onClick={this.handleSubmit} data-testid='exclusion-submit' disabled={reasonLabel === ''}>
```
nicksp: Same here with the `size` prop.

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,290 @@
...
+
+    const style = {
+      height: maxHeight,
+      maxHeight: `calc(100vh - ${margins}px)`,
+      top: top,
+      left: left + width + 12
+    }
+
+    // Small screens than popup max height
+    if (innerHeight <= maxHeight + margins) {
```
nicksp: 🤔 Don't you think there's lots of conditions and nesting and the whole logic is a little bit vague, kinda hard to read? Could we try to improve so it's easier to read and reason about?

What if we extract the top level condition into a variable with the meaningful name and then extract both top level conditions' bodies to 2 separate methods? So the idea is to have smth like that in the end:

```js
if (isBigScreen) {
  recalculateDimensiononBigScreen(...)
} else {
  recalculateDimensiononBigScreen(...)
}
```

names are just examples, feel free to come up with better versions. The final `updateDimensions` method will be much simpler to read and we can ignore those low-level methods that will do calculations. It's a separate concern so should be extracted.

Wdyt? 😄

bernardodiasc:
Yeah, that edge case is quite tricky to workaround. There's always the fact that someone can click on a visible element and resize the screen with the popup opened until the click target go out of the screen. The current implementation differs the two cases:

1. when the click target is somehow in a good place, the triangle is positioned the the half height of the click target.
2. when the click target is somehow out of boundaries, the triangle is positioned on the popup edges.

By the way, I was recording the UX mentioned above and I found an issue: 

Observe how the triangle jumps when the popup docks to the bottom:



I fixed already, this is how it should be, smooth:



Will send that fix together with the code review changes.

nicksp:
> Observe how the popup jumps when it docks to the bottom
Great, that you fixed that, looked much smoother now 👍 What I wanted to say before I read it, that it's very unlikely that somebody will be resizing while working on the app. We don't have many users and it's a very rare case I'd say. Also wanted to suggest to fix in case it was an easy one to deal with.

But, as it's fixed now, great catch 🎉 Looking forward to reviewing the code again once all review notes are addressed and we'll be great to move this forward 😄

bernardodiasc:
> What I wanted to say before I read it, that it's very unlikely that somebody will be resizing while working on the app.

Yeah, of course. It's makes more sense if you think about this resizing check as a fast way to test all possible positions and how the calculation results of each case. The bug exposed a wrong variable used in a condition :)

bernardodiasc:
hey @nicksp and @marcoacfilho, I addressed all the code review requests. Please take a look when you have some time. Thanks!

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -0,0 +1,319 @@
...
+      return 0
+    }
+    return this.getAnchorPositionedAtMiddleOfClickedElement({ top, initialAnchor, marginTop })
+  }
+
+  getPopupTopWhenDockedToTheScreenBottom = ({ innerHeight, maxHeight, marginBottom }: { innerHeight: number, maxHeight: number, marginBottom: number }) => innerHeight - maxHeight - marginBottom
+
+  getAnchorPositionedAtBottomEdgeOfThePopupWhenPopupIsDockedToTheScreenBottom = ({ maxHeight, triangleHeight }: { maxHeight: number, triangleHeight: number }) => maxHeight - triangleHeight
+
+  getAnchorPositionedAtMiddleOfClickedElementWhenPopupIsDockedToTheScreenBottom = ({ innerHeight, maxHeight, marginBottom, top, initialAnchor }: { innerHeight: number, maxHeight: number, marginBottom: number, top: number, initialAnchor: number }) => top - this.getPopupTopWhenDockedToTheScreenBottom({ innerHeight, maxHeight, marginBottom }) + initialAnchor
```
nicksp: I like the composition we made, but usually, such "long" names are bad practice and hard to read. Let's keep it as it is for now, but if you can come up with shorter more generic names, please do as a separate commit.

PR #370, opened 2019-03-26
Fix SelectBox visibleItems height calculation

bernardodiasc:
**Release Type:** *Bug fix*

Fixes https://x-team-internal.atlassian.net/browse/XP-2940

## Description

Fix SelectBox visible items height calculation.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-2940-fix-selectbox-visibleitems-prop` | [XP](https://github.com/x-team/xp/pull/1850)

## Steps to Test or Reproduce

- `visitbleItems` value should works as expected both for normal size as for small.

`SelectBox`

## Screenshots

This PR fixed a regression, so many places now behaves as expected:

Screen Shot 2019-03-26 at 15 10 30
Screen Shot 2019-03-26 at 15 11 20
Screen Shot 2019-03-26 at 15 11 35

### DIFF
--- a/src/components/ui/SelectBox.js
+++ b/src/components/ui/SelectBox.js
@@ -834,7 +834,7 @@ class SelectBox extends Component<Props, State> {
 
     const getItemClasses = (item) => ([
       cx.item,
-      (size === 'small') ? cx.itemSmall : '',
+      isSmall() ? cx.itemSmall : '',
       (lined || !expanded) ? cx.lined : '',
       ((item.editing || item.editing === '') && item.editing !== item.value) ? cx.editing : ''
     ].join(' '))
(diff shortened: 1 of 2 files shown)


src/components/ui/SelectBox.js
```diff
@@ -1126,25 +1126,31 @@ class SelectBox extends Component<Props, State> {
       return data
     }
 
+    const isSmall = () => size === 'small'
+
+    const getHeightFromVisibleItems = () => isSmall()
```
nicksp: Since we reverted the condition, we've accidentally affected the following items' heights on the List view search form. This it correct since the condition was wrong actually but just smth to be aware of:
- Select List
- Select Field
- Status

We actually show less items now. So I wonder if we want to change `SELECTBOX_HEIGTH` to `4` by default instead of `3` to retain the way it looks now. Wdyt?

image

bernardodiasc: Yeah, I agree with you. Will updated the code.

bernardodiasc:
@nicksp I agree with you with this update. The height calculation break at some point, specially for small size, and it seems no one realised that, and I noticed that the requests for visible amount of items was adjusted on top of the wrong calculation. So I guess it should be fine to update this to reflect what had before.

PR #374, opened 2019-03-26
Wire up the tooltip UI component

marcoacfilho:

Fixes https://x-team-internal.atlassian.net/browse/XP-2929

## Description

Removed DropDown from ApplicantBadge, refactored calculations to arrow and names.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-2929-wire-up-the-tooltip-ui-component-with-redux-in-the-application` | [XP](https://github.com/x-team/xp/pull/1849)

## Steps to Test or Reproduce

1. Filter component ListExclusionFormPopup
2. Check if arrows are in place when clicking on different buttons, check if name changes in all Popups

- `ui/ListExclusionFormPopup.js`

## Screenshots

image

### DIFF
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -633,10 +633,15 @@ class ApplicantBadge extends PureComponent<Props> {
       />
     )
 
-    const handleClick = (event: Object) => {
+    const handleClick = (event: SyntheticEvent) => {
       event.stopPropagation()
       const { id, onClick } = this.props
       onClick && onClick(id)
+    }
+
+    const actionEvent = (event: SyntheticEvent, actionOnClick: Function) => {
+      event.stopPropagation()
+      actionOnClick && actionOnClick(event.currentTarget.getBoundingClientRect())
     }
 
     if (!id) {
@@ -673,15 +678,12 @@ class ApplicantBadge extends PureComponent<Props> {
           )}
         </div>
         <div className={cx.controls}>
-          {!isTabular && actions.map(({ key, icon: Icon, onClick = null }) => (
+          {!isTabular && actions.map(({ key, icon: Icon, onClick: actionOnClick }) => (
             Icon && (
               <span
                 key={key}
                 className={cx.control}
-                onClick={event => {
-                  event.stopPropagation()
-                  onClick && onClick(event.currentTarget.getBoundingClientRect())
-                }}
+                onClick={event => actionEvent(event, actionOnClick)}
               >
                 <Icon />
               </span>
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -1,3 +1,5 @@
+/* global SyntheticEvent */
+
 import React, { PureComponent } from 'react'
 import md5 from 'crypto-js/md5'
 
@@ -567,6 +569,17 @@ class ApplicantBadge extends PureComponent<Props> {
     handleRankingChange && handleRankingChange(ranking)
   }
 
+  handleClick = (event: SyntheticEvent<>) => {
+    event.stopPropagation()
+    const { id, onClick } = this.props
+    onClick && onClick(id)
+  }
+
+  handleActionClick = (onClick: () => mixed) => (event: SyntheticEvent<>) => {
+    event.stopPropagation()
+    onClick && onClick(event.currentTarget.getBoundingClientRect())
+  }
+
   render () {
     const {
       id,
@@ -633,17 +646,6 @@ class ApplicantBadge extends PureComponent<Props> {
       />
     )
 
-    const handleClick = (event: SyntheticEvent) => {
-      event.stopPropagation()
-      const { id, onClick } = this.props
-      onClick && onClick(id)
-    }
-
-    const actionEvent = (event: SyntheticEvent, actionOnClick: Function) => {
-      event.stopPropagation()
-      actionOnClick && actionOnClick(event.currentTarget.getBoundingClientRect())
-    }
-
     if (!id) {
       return null
     }
@@ -663,7 +665,7 @@ class ApplicantBadge extends PureComponent<Props> {
     ].map(item => ({ ...item, selected: item.id === ranking }))
 
     return (
-      <div onClick={handleClick} className={[cx.mode, cx.displayControlsOnHover, active ? cx.active : ''].join(' ')}>
+      <div onClick={this.handleClick} className={[cx.mode, cx.displayControlsOnHover, active ? cx.active : ''].join(' ')}>
         <div className={cx.name}>
           <div className={cx.nameInner} title={name || email}>{name || email}</div>
           {status && this.renderStatusIndicator()}
@@ -678,12 +680,12 @@ class ApplicantBadge extends PureComponent<Props> {
           )}
         </div>
         <div className={cx.controls}>
-          {!isTabular && actions.map(({ key, icon: Icon, onClick: actionOnClick }) => (
+          {!isTabular && actions.map(({ key, icon: Icon, onClick }) => (
             Icon && (
               <span
                 key={key}
                 className={cx.control}
-                onClick={event => actionEvent(event, actionOnClick)}
+                onClick={this.handleActionClick(onClick)}
               >
                 <Icon />
               </span>
--- a/src/components/ui/ApplicantBadge.js
+++ b/src/components/ui/ApplicantBadge.js
@@ -575,7 +575,7 @@ class ApplicantBadge extends PureComponent<Props> {
     onClick && onClick(id)
   }
 
-  handleActionClick = (onClick: () => mixed) => (event: Element) => {
+  handleActionClick = (onClick: () => mixed) => (event: SyntheticEvent<>) => {
     event.stopPropagation()
     onClick && onClick(event.currentTarget)
   }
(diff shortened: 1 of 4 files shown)


nicksp:
Just updated the PR title to be in line with our standards. It's quite helpful to have a consistent and clean title so that it will be used when merging that.
cc @marcoacfilho

bernardodiasc:
I updated the `ApplicantBadge` stories to reflect what we need:

Screen Shot 2019-03-28 at 12 38 55

This was a quick cleanup in the stories, but the big work for cleanup of the `ApplicantBadge` will be done on another task dedicated for that https://x-team-internal.atlassian.net/browse/XP-2879

src/components/ui/ApplicantBadge.js
```diff
@@ -675,22 +673,18 @@ class ApplicantBadge extends PureComponent<Props> {
...
-                className={dropdownClassName}
-                tooltipClassName={tooltipClassName}
-                onClick={onClick}
-                label={(
-                  <span className={cx.control}>
-                    <Icon />
-                  </span>
-                )}
+                className={cx.control}
+                onClick={event => {
```
nicksp: Not an urgent request, but would be good to have. For performance reasons, I'd replace this inline function, with a separate one to prevent creation of new anonymous functions in memory each time component re-renders.

nicksp:
Hey @bernardodiasc I approved that but would be nice to see some a change applied. LMK if I should elaborate further or help with anything.

src/components/ui/ApplicantBadge.js
bernardodiasc: Can you please take a look at this commit https://github.com/x-team/xp-ui/pull/374/commits/9ad0f93adacc65b8956d33465c06c3dcf6bcb42c if it solves?

nicksp: Replied here https://github.com/x-team/xp-ui/commit/9ad0f93adacc65b8956d33465c06c3dcf6bcb42c#r32955553

bernardodiasc: Alright, another round, please take a look https://github.com/x-team/xp-ui/pull/374/commits/9007e837c43468345520be696ed2a993460dc0b5

nicksp: Looks perfect mate! Just one more suggestion regarding the related handler and we're good.

src/components/ui/ApplicantBadge.js
```diff
@@ -569,6 +569,17 @@ class ApplicantBadge extends PureComponent<Props> {
     handleRankingChange && handleRankingChange(ranking)
   }
 
+  handleBadgeClick = (event: SyntheticEvent<>) => {
+    event.stopPropagation()
+    const { id, onClick } = this.props
+    onClick && onClick(id)
+  }
+
+  handleActionClick = (onClick: () => mixed) => (event: Element) => {
```
nicksp: why did we change `event`'s type to an `Event` instead of a proper `SyntheticEvent<T>` kind of [types](https://flow.org/en/docs/react/events/). `Element` is there more to annotate React elements, rather than events imo.

bernardodiasc: Indeed the param name `event` is misused here, we should expect an element, not the event.

nicksp: really?! How come, since this is a native event that React passes to us.

bernardodiasc: Ah, never mind. The confusion is on my side, this method still get the event, but will pass down in the onClick an element instead. Will fix the type.

nicksp:
@bernardodiasc Could you please also approve changes? Seems like you're still requesting them 😆

image

nicksp:
@marcoacfilho @bernardodiasc Please note, that I've merged #375 because it wasn't in sync with this branch and I thought we're better to address all the notes in one scope here instead of having several PRs. Please address these things: https://github.com/x-team/xp-ui/pull/375#pullrequestreview-220895885

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -267,7 +263,19 @@ class ListExclusionFormPopup extends PureComponent<Props, State> {
...
       style
     })
-  }, 1000)
+  }
+
+  updateDimensions = throttle(this.setUpdatedDimensions, 1000)
+
+  componentDidUpdate (prevProps: Props, prevState: State) {
+    const positioning = this.getActionElementPosition()
+    if (JSON.stringify(prevState.positioning) !== JSON.stringify(positioning)) {
```
nicksp: This method of comparing object are brittle and limited:

1. It only works with simple JSON-style objects without methods and DOM nodes inside (shouldn't be the case for us but just to keep in mind)
1. Props are expected to be in the SAME order. So this method will return false for the following objects:
```js
x = { a: 1, b: 2 }
y = { b: 2, a: 1 }
```

A more reliable solution is to use [native method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is), isn't supported in IE though or, since we have `lodash.isequal` pulled in, let's use this flexible and safe method.

src/components/ui/ListExclusionFormPopup.js
```diff
@@ -169,18 +164,19 @@ class ListExclusionFormPopup extends PureComponent<Props, State> {
...
+    }))
   }
 
   handleCommentChange = (event: SyntheticInputEvent<HTMLInputElement>) => {
-    const { reasons } = this.props
-    this.setState({
-      reasonIndex: reasons.length,
-      reasonLabel: event.target.value
-    })
+    const value = event.target.value
```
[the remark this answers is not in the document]
bernardodiasc: this is a good point, I can't understand exactly why, but usigng `reasonLabel: event.target.value` leads to "cannot get value from undefined"

src/components/ui/ListExclusionFormPopup.js
bernardodiasc: I don't know why, but using lodash.isEqual returned false all the times, even when they were equal. I guess that positioning is not a regular object but a DOMRect. I'll try that again, but I'm afraid that's not possible (because I did tried that before using JOSN stringify). If you have another suggestion please let me know.

src/components/ui/ListExclusionFormPopup.js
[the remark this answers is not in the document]
bernardodiasc: I removed the closure in the setState and it seems to work

src/components/ui/ListExclusionFormPopup.js
bernardodiasc: Well, I tried again, is it's like I said, but I found out a native solution `el.getBoundingClientRect().toJSON()`, with that I can use isEqual 👍

src/components/ui/ListExclusionFormPopup.js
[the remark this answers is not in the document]
nicksp: @bernardodiasc That's strange. I've just cheched that myself and it seems linting fine to me:

```js
reasonLabel: event.target.value
``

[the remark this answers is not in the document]
Comment (nicksp) [reply src/components/ui/ListExclusionFormPopup.js]: Could you please re-test again? Doesn't seem that it should fail 'cause `event` is a mandatory param here, so Flow shouldn't complain.

File: src/components/ui/ListExclusionFormPopup.js (hunk 3 above)
Comment (nicksp) [reply src/components/ui/ListExclusionFormPopup.js]: That's a good finding and a nice option to proceed with. FYI: The reason you can't compare returned objects from `getBoundingClientRect` is that properties in the returned `DOMRect` object are not own properties, and thus, some APIs won't return properties. As a result, you always get an empty object back and any attempt to compare two empty objects will fail. Hope that's clear.

File: src/components/ui/ListExclusionFormPopup.js (hunk 4 above)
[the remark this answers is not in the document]
Comment (bernardodiasc) [reply src/components/ui/ListExclusionFormPopup.js]: it wasn't on flow, it was throwing error on browser. I know this is strange. but no worries, I removed the closure of the setState and this is working again.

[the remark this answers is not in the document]
Comment (nicksp) [reply src/components/ui/ListExclusionFormPopup.js]: Sorry didn't get you correctly, nut just wanted to double check on my machine

**Comment by bernardodiasc** [comment]
hey @nicksp this is ready for another review, I addressed your suggestions at https://github.com/x-team/xp-ui/pull/374/commits/c21fdec068049411ba3469649ec00dffc3acec2b, thanks!

File: src/components/ui/ListExclusionFormPopup.js [hunk 5]
```diff
@@ -169,18 +168,11 @@ class ListExclusionFormPopup extends PureComponent<Props, State> {
...
-      reasonIndex,
-      reasonLabel
-    })
+    this.setState({ reasonIndex, reasonLabel })
   }
 
   handleCommentChange = (event: SyntheticInputEvent<HTMLInputElement>) => {
-    const { reasons } = this.props
-    this.setState({
-      reasonIndex: reasons.length,
```
nicksp: did we exclude this prop on reason?

bernardodiasc: This was used to deselect the radio buttons in case the user inputs a custom comment. Since this is not the expected UX for the component, by removing it, it will persist the last selected option even if the user updates the comment. As we discussed, the only value sent to back-end is a string (the `reasonLabel` in this case).

nicksp: That's correct, agree.

PR #383, opened 2019-04-10
Copy email to clipboard on click

bernardorubin:

Fixes https://x-team-internal.atlassian.net/browse/XP-2749

## Checklist

## Related PRs

XP-2749-copy-email-to-clipboard-on-click | [XP](https://github.com/x-team/xp/pull/1863)

## Steps to Test or Reproduce

Download Branch
Install new package and flow definitions by running `npm install`
run npm`start`
Under UI components GenericCopyToClipboard component should appear

Another new component

## Screenshots

demo2


bernardorubin:
Locally I have no errors 
Screen Shot 2019-04-09 at 11 37 29 PM

but Travis is failing to install a new `node_module` `react-copy-to-clipboard`
Screen Shot 2019-04-09 at 11 38 31 PM

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -0,0 +1,48 @@
+// @flow
+
+import React, { Component } from 'react'
+import { CopyToClipboard } from 'react-copy-to-clipboard'
+import Tooltip from './Tooltip'
+import type { Element } from 'react'
+const cmz = require('cmz')
+
+const container = cmz(`
+cursor: pointer
```
bernardodiasc: looks its missing some indentation here:

```suggestion
  cursor: pointer
```

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -0,0 +1,48 @@
+// @flow
+
+import React, { Component } from 'react'
+import { CopyToClipboard } from 'react-copy-to-clipboard'
+import Tooltip from './Tooltip'
+import type { Element } from 'react'
+const cmz = require('cmz')
```
bernardodiasc: There are a pattern in place that separates the kinds of import  with a blank line:

```js
// External libraries
import React, { Component } from 'react'
import { CopyToClipboard } from 'react-copy-to-clipboard'

// Internal components
import Tooltip from './Tooltip'

// Type defs
import type { Element } from 'react'

// The cmz, that for some reason I'm not exactly sure, needs to be declared using require and as a `const`, it also needs to be later than the imports
const cmz = require('cmz')
```

As simple as it looks like, it's good to stick to the existing patterns :)

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,55 @@
+// @flow
+
+import React from 'react'
```
bernardodiasc: As mentioned in `GenericCopyToClipboard/index.js` feedback, there's a pattern in place that separates the kinds of imports with a blank line.

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+const cmz = require('cmz')
+
+const container = cmz(`
+  position: relative;
+`)
+
+const bubble = cmz(`
```
bernardodiasc: Please fix indentation in this styles declarations, it should be `2 spaces` as the other places.

Also please remove `;` as we are not using it in XP-UI.

One last thing for this part is separate selectors with a blank like, e.g.:

```
  & {
    ...
  }

  &:after {
    ...
  }
```

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+  line-height: 1.4;
+  padding: .75em;
+  text-align: center;
+`)
+
+type Props = {
+  showTooltip: boolean
+}
+
+export default function Tooltip ({ showTooltip }: Props) {
```
bernardodiasc: Can we keep with current pattern for the declaration? I know this looks one line simpler, but by having consistency with patterns we leverage the mental effort to reason about the code :)

`const function Tooltip (...) { ... }`

and then

`export default Tooltip`

src/components/ui/GenericCopyToClipboard/GenericCopyToClipboard.stories.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+
+import GenericCopyToClipboard from '.'
+
+storiesOf('UI Components/GenericCopyToClipboard', module)
```
bernardodiasc: I would like to suggest adding 2 more stories here:

Firstly these can be added in a new story group, as:

`storiesOf('UI Components/GenericCopyToClipboard/Debug', module)`

Then adding a story like `standalone Tooltip` in a way that we can check how that component looks in isolation.

And one another for the `(does component explode?)` of the `Tooltip` as well. This story help us to make sure the minimal render of the component doesn't crash the application nor looks visually broken in anyway.

Also, please make sure to move the existing `missing children (does component explode?)` to the `Debug` category as well and maybe make more explicit that is about `GenericCopyToClipboard`, given that we'll have another story of this type for `Tooltip`.

Please let me know if you have any question or concern :)

src/components/ui/GenericCopyToClipboard/index.js
bernardorubin: Ok, will do the imports in that way. Thanks for the suggestion. 
When I declared the normal import of `cmz` it would return an error and I saw the other files imported the library this way so I stuck to the previous conventions.

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardorubin: I saw this pattern on other files aswell that's why I used it but will gladly change it.
```
export default function ActivityBar ({ text, datetime }: Props) {
  return <div className={cx.container}>
    <SvgIcon icon='time' color='grayscale' />
    <div className={cx.activity} title={text}>{text}</div>
    <div className={cx.time}>{timeSince(datetime, false)}</div>
  </div>
}
```

```
export default function FilterTag ({ children, onClickRemove }: Props) {
  return (
    <Button className={cx.tabularFilter} color='silver' component='span' selected outlined rounded raised>
      <span>{children}</span>
      <span className={cx.removeTabularFilter} onClick={onClickRemove}>
        ×
      </span>
    </Button>
  )
}
```

src/components/ui/GenericCopyToClipboard/index.js
bernardodiasc: yeah, `cmz` is not compatible with `import`, we need to stick with `const cmz = require('cmz')` 👍

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardodiasc: Interesting, so maybe this is allowed variation. In that case this should fine as it is and I was not aware, apologies for the confusion :)

bernardorubin: I already changed it the way you suggested, I think the norm is the syntax you suggested instead of what I wrote which was a bit simpler.

nicksp:
Rebased with `develop` branch

flow-typed/npm/@types/react-copy-to-clipboard_vx.x.x.js
```diff
@@ -0,0 +1,18 @@
+// flow-typed signature: 85744d2c4b76904dcf2eba12e66acec0
```
nicksp: This is a duplicated Flow libdef file and we can remove that safely.

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -0,0 +1,51 @@
...
+import type { Element } from 'react'
+
+const cmz = require('cmz')
+
+const container = cmz(`
+  cursor: pointer
+`)
+
+type Props = {
+  children?: Element<*>
```
nicksp: It's more preferrable to use the `Node` type for the `children` prop, as it includes everything React can render.
```suggestion
  children: Node
```

And don't forget to import `Node` type: `import type { Node } from 'react'`

Check this react Flow definitions for more details: https://github.com/facebook/flow/blob/5958ffb8dd173c86dc5fc9abbfd2a87eea2d4b4a/lib/react.js#L12-L23

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -0,0 +1,51 @@
...
+  showTooltip: boolean
+}
+
+class GenericCopyToClipboard extends Component<Props, State> {
+  state = {
+    showTooltip: false
+  }
+
+  handleCopy = (value: string) => () => {
+    setTimeout(() => this.setState({ showTooltip: false }), 2500)
```
nicksp: I totally agree with Bernardo, that we should clear timeout otherwise you'll keep functions run by the timer in memory over time, which might impact performance. Right? It's always a great pattern to clean up after yourself, this includes timers you set, events you subscribe to, etc.

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  render () {
+    const { showTooltip } = this.state
+    const { children } = this.props
+    return children ? (
+      <div className={container}>
+        <Tooltip showTooltip={showTooltip} />
+        <CopyToClipboard
+          text={children.props.children}
```
nicksp: Could we pass a text as a prop, otherwise it looks confusing how you obtain that via `children.props.children`. Looks not clear if you check the code first time. What if we just pass the text we need to copy as a prop? It would be much easier to understand and re-use component, right? What if `children.props.children` isn't available and it's not obvious it contains text. So let's switch to using a prop instead.

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+
+  &:after {
+    content: ' '
+    position: absolute
+    top: 100%
+    left: 50%
+    margin-left: -10%
+    border-width: 5px
+    border-style: solid
+    border-color: rgba(0,0,0,0.8) transparent transparent transparent
```
nicksp: can we just say: `border-color-top: rgba(0, 0, 0, .8)`? Will it be enough?

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+`)
+
+const message = cmz(`
+  background: rgba(0,0,0,0.8)
+  border-radius: 3px
+  color: white
```
nicksp: It's ways better not to hard code any colors or typography related things, but rather to re-use them. so it's easier to adjust things in the future. Please check how we specify the color for example: https://github.com/x-team/xp-ui/blob/master/src/components/ui/Button.js#L106

Please check which constants we have defined and thus can re-use: https://github.com/x-team/xp-ui/blob/master/src/styles/theme.js

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+  line-height: 1.4
+  padding: .75em
+  text-align: center
+`)
+
+type Props = {
+  showTooltip: boolean
+}
+
+export default function Tooltip ({ showTooltip }: Props) {
```
nicksp: Since this is now stateless and doesn't rely upon lifecycle hooks, I suggest we convert that to a functional component instead. The main benefit among other small nice things is that the final transpiled code takes less bytes. Lots of boilerplate go into creating a class. Compare https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECCAO9oG8BQ0BOBTAdgEywwAoBKFadaTLAFwFcMdoc6QR0BfVDoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.3.3 (class transpilation) to https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAggBwTAvDBAnECKoHwwAUAUDADwAmAlgG55gCuANk2QPTV3ECUQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.3.3 (function representation)

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+
+import React from 'react'
+
+const cmz = require('cmz')
+
+const container = cmz(`
+  position: relative
+`)
+
+const bubble = cmz(`
```
nicksp: I'd also suggest using sans-serif kind of font for the tooltip text to make it look cleaner. Currently, I think it looks a bit messy. Check the font we use in `xp-ui` to reuse smth that suit us:

image

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardorubin: It is a functional component

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardorubin: Doesn't work by itself `border-color-top: rgba(0,0,0,0.8)`. Results in this:
Screen Shot 2019-04-11 at 1 16 56 PM
also this doesn't work:
```
border: rgba(0,0,0,0)
border-color-top: rgba(0,0,0,0.8)
```
neither this 
```
border: transparent
border-color-top: rgba(0,0,0,0.8) !important
```
or this
```
border-color-top: rgba(0,0,0,0.8)
border-color-bottom: transparent
border-color-right: transparent
border-color-left: transparent
```
I'll just leave it as it is.

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardorubin: I didn't touch the font styles to allow the end user to customize the font as it happens right now on the xp repo, but I'll change the font

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardorubin: Alright, changed to baseDark

nicksp: what do you mean to baseDark? You need a white, right?  So you need `baseBrighter` cc @bernardorubin

src/components/ui/GenericCopyToClipboard/Tooltip.js
nicksp: You're right, sorry. I thought I removed this comment, but seems like it made its way into the final review. Sure, you're doing a triangle so you need to keep it as is.

bernardorubin:
Also updated xp branch to use the new `text` props

nicksp:
@migueliriano you seem to fall into a trap 😄 It should be @bernardorubin

bernardorubin:
I just found a bug that makes the edit icon go to the next line
Screen Shot 2019-04-11 at 5 47 55 PM

nicksp:
> I just found a bug that makes the edit icon go to the next line

Can we get it back as it was?

bernardorubin:
I'm fixing it.

bernardorubin:
It is now working correctly:
Screen Shot 2019-04-11 at 6 22 26 PM

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -0,0 +1,58 @@
...
+    border-style: solid
+    border-color: ${theme.baseDark} transparent transparent transparent
+  }
+`)
+
+const message = cmz(`
+  font-family: 'Open Sans', 'Helvetica Neue', sans-serif
+  background: ${theme.baseDark}
+  border-radius: 3px
+  color: white
```
nicksp: We can also change this color.

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  render () {
+    const { showTooltip } = this.state
+    const { children, text } = this.props
+    return text ? (
+      <div className={container}>
+        <Tooltip showTooltip={showTooltip} />
+        <CopyToClipboard text={text} onCopy={this.handleCopy()}>
+          {children && children}
```
nicksp: Why do we need to check here? If `children` is null it won't be rendered, so we can safely just say: `children`

nicksp:
Let's wait for @bernardodiasc 's review as he'd requested changes before we merge.

PR #400, opened 2019-05-31
Add read more functionality

bernardorubin:

Fixes https://x-team-internal.atlassian.net/browse/XP-2182

## Description

Truncate the display of Notes that are over 250 characters as seen in this image and provide a "Read More" link which expands to see the full note. Attachments should still be displayed at the botton of the teaser so they are downloadable without requiring the "Read More" to be clicked first.

## Checklist

- [x] fill out `Is a breaking change` and `Release Type Reason` (if required) fields for the ticket.

## Steps to Test or Reproduce

Go to http://localhost:9001/?selectedKind=UI%20Components%2FNote&selectedStory=showing%20read%20more%20and%20show%20less&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel&background=%23fff

Notes component


bernardorubin:
AC mentions truncation should be done by word count. I spoke with Bernardo and the library does not seem to support truncation by number of words. The functionality of the library works by using lines as a parameter and when the window is resized, the content stays within those lines and text gets truncated depending on screen width. You can try to test this by resizing the window to understand how it works.

Right now functionality uses the library `react-truncate-markup` as it is intended to be used.

nicksp:
@bernardorubin @bernardodiasc What was the reason guys you chose `react-truncate-markup`? Did you do an investigation of available options and this was a winner? Which pros/cons did it have? 
I just did a quick investigation myself and here's what I found: https://www.npmtrends.com/react-truncate-vs-react-truncate-html-vs-react-truncate-markup

Seems like [react-truncate](https://github.com/One-com/react-truncate) is more popular, more mature and 2 times smaller. By reading its documentation it seems to perform same functionality, so might be a good option to be used instead of the current solution. Wdyt?

src/components/ui/Note.js
```diff
@@ -164,12 +193,47 @@ class Note extends PureComponent<Props, State> {
...
+    }))
+  }
+
   renderPresenter = ({ activateEditingMode }: PresenterProps) => {
     this.activateEditingMode = activateEditingMode
-    const { newValue } = this.state
-    return TextWrapper(
+    const { newValue, shouldTruncate } = this.state
+    const readMore = (
+      <span>
```
nicksp: Do we need this wrapper? Is it a critical one for a layout or styling purposes? If not, we can get rid of that by using React's Fragment.

src/components/ui/Note.js
```diff
@@ -164,12 +193,47 @@ class Note extends PureComponent<Props, State> {
...
+        {shouldTruncate && '...'}
+        {ReadMoreWrapper(
+          <div onClick={this.toggleTruncate}>
+            {shouldTruncate ? (
+              <span>
```
nicksp: if this span isn't critical for rendering the component, we can consider using `Fragment` to get rid of useless markup. Please let me know if that's the case.

src/components/ui/Note.js
```diff
@@ -164,12 +193,47 @@ class Note extends PureComponent<Props, State> {
...
+                <span>READ MORE</span>
+                <SvgIcon icon='triangledown' />
+              </span>
+            ) : (
+              <span>
+                <span>SHOW LESS</span>
```
nicksp: To change UI representation of the text it's encouraged to use CSS, that's what it was created for, and have an initial text written in a user-friendly form like that: `Show less`, like how'd you write that in plain English. It's only then when the markup is completed, we apply CSS to transform things into a desirable state.  In other words, capitalization is a text effect, and effects ideally should be controlled via respective technology (in this particular case, via CSS), so that once the technology is turned off the text is represented in a natural and user-friendly way.

Another form of representing the idea: **CSS** should be used for styling purposes.
- If you change your mind later on, then removing that CSS property will be enough. Updating all HTML would be a nightmare.
- Accessibility-wise, you should write the text as it is supposed to be written for a vocal reader to read it (for blind people).

Hope the arguments above are clear enough and explain the reasons behind the suggestion I made 😄

src/components/ui/Note.js
```diff
@@ -164,12 +193,47 @@ class Note extends PureComponent<Props, State> {
...
+
+    return shouldTruncate ? TextWrapper(
+      RichContentWrapper(
+        <TruncateMarkup lines={3} ellipsis={readMore}>
+          <div>{newValue}</div>
+        </TruncateMarkup>
+      )) : TextWrapper(
       RichContentWrapper(
-        <RichTextEditor initialValue={newValue} mode='viewer' />
+        <div>
```
nicksp: Do we need this wrapper? Is it a critical one for a layout or styling purposes? If not, we can get rid of that by using React's Fragment.

src/components/ui/Note.js
```diff
@@ -164,12 +193,47 @@ class Note extends PureComponent<Props, State> {
...
+        {shouldTruncate && '...'}
+        {ReadMoreWrapper(
+          <div onClick={this.toggleTruncate}>
+            {shouldTruncate ? (
+              <span>
+                <span>READ MORE</span>
+                <SvgIcon icon='triangledown' />
+              </span>
+            ) : (
+              <span>
```
nicksp: Do we need this wrapper? Is it a critical one for a layout or styling purposes? If not, we can get rid of that by using React's Fragment.

bernardorubin: Fixed here https://github.com/x-team/xp-ui/pull/400/commits/95b189da14add1bb2538d9eac4cd9914e6309ee4

src/components/ui/Note.js
bernardorubin: Fixed here https://github.com/x-team/xp-ui/pull/400/commits/95b189da14add1bb2538d9eac4cd9914e6309ee4

src/components/ui/Note.js
bernardorubin: Fixed here https://github.com/x-team/xp-ui/pull/400/commits/95b189da14add1bb2538d9eac4cd9914e6309ee4

src/components/ui/Note.js
bernardorubin: Fixed here https://github.com/x-team/xp-ui/pull/400/commits/95b189da14add1bb2538d9eac4cd9914e6309ee4

bernardorubin:
Demo of latest working changes:

bernardorubin:
New Demo:

src/components/ui/Note.stories.js
```diff
@@ -64,7 +64,7 @@ storiesOf('UI Components/Note', module)
         date={noteDate}
         showNoteType
         name='Samwise Gamgee'
-        text={`Do you remember the Shire, Mr. Frodo? It'll be spring soon. And the orchards will be in blossom. And the birds will be nesting in the hazel thicket. And they'll be sowing the summer barley in the lower fields... and eating the first of the strawberries with cream. Do you remember the taste of strawberries?`}
+        text={`Do you **remember** the Shire, Mr. Frodo? It'll be spring soon. And the orchards will be in blossom. And the birds will be nesting in the hazel thicket. And they'll be sowing the summer barley in the lower fields... and eating the first of the strawberries with cream. Do you remember the taste of strawberries?`}
```
bernardorubin: Left the **bold** here because it might be helpful for other developers to see here how markdown is rendered instead of plain text but can remove it. I know it is outside of the ticket's scope.

src/components/ui/Note.js
bernardorubin: Fixed here https://github.com/x-team/xp-ui/pull/400/commits/9c1a63bac5f4cb34f238f9e5d406f54e2c7a3012

nicksp:
It's been inactive for a quite long time, so I'm gonna close it for the time being. We can reopen that or cherry-pick commits when needed.

PR #406, opened 2019-06-24
Improve GenericCollapsible to become more performant and prevent unnecessary re-render of the body content

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3142

## Description

This PR changes the way GenericCollapsible hides the body, instead removing from DOM it just hides using CSS. With that the body state are persisted during visibility toggles and it becomes more performant since it doesn't cause the body to re-render every time it's expanded.

## Checklist

## Steps to Test or Reproduce

Please check **Screenshots** section below.

- GenericCollapsible

### Storybook

- Go to GenericCollapsible or Filters (which uses GenericCollapsible) and observe the DOM. The rendered element are always there and it just update the CSS classes to hide the body.

### XP App

- Observe that the expand/collapse experience is improved:
    - The exact state of the panel body is preserved (e.g. when scrolling the body of a panel, even after collapsing and expanding again, the visible area will be persisted).
    - Despite some other known performance issues, the lag of the expanding action was drastically reduced.

### DIFF
--- a/src/components/ui/GenericCollapsible/Body.js
+++ b/src/components/ui/GenericCollapsible/Body.js
@@ -1,12 +1,37 @@
 // @flow
 
+import React from 'react'
+
 import type { Element } from 'react'
+
+const cmz = require('cmz')
+
+const cx = {
+  collapsed: cmz(`
+    display: none
+  `),
+
+  accordion: cmz(`
+    overflow-y: auto
+  `)
+}
 
 export type Props = {
   isExpanded: boolean,
+  isAccordion: boolean,
   children: Element<*>,
 }
 
-const Body = (props: Props) => props.isExpanded ? props.children : null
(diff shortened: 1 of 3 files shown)


src/components/ui/GenericCollapsible/Body.js
```diff
@@ -1,12 +1,37 @@
...
 }
 
-const Body = (props: Props) => props.isExpanded ? props.children : null
+const Body = (props: Props) => {
+  const classNames = [
+    props.isExpanded ? '' : cx.collapsed,
+    props.isAccordion ? cx.accordion : ''
+  ].join(' ')
+  return (
+    <div className={classNames}>
```
nicksp: Good. Now we always render the content and not **nullify** it.

PR #408, opened 2019-07-04
Implement lazy-load for Avatar img requests

kararade:

Fixes https://x-team-internal.atlassian.net/browse/XP-3144

## Description

Implemented lazy-load for Avatar to load img files only when the Avatar is visible on viewport the first time

## Checklist

### XP-UI App
1. Run the app with the latest changes from this branch using the `npm run watch`

### XP App
1. Download the latest changes from `master`
2. Run the app linking the local `XP-UI` app following the fallback https://github.com/x-team/xp/wiki/Workflow-with-XP-and-XP-UI#fallback-solution
3. Go to http://localhost:8000/admin and open the webmaster tool to check the `network tab` and just filter by `img`
4. To see the lazyload effect start scrolling down the profile-list and meanwhile the avatars are in the viewport check how the image is downloaded.
5. Confirm that the images that aren't in the viewport don't have any image yet

Avatar UI component


src/components/ui/Avatar.js
```diff
@@ -17,19 +18,32 @@ type Props = {
...
-    round
-    {...rest}
-  />
+  const { alt, src, size } = props
+  let imgSrc = null
+
+  return (
+    <VisibilitySensor>
+      {({ isVisible }) => {
+        if (isVisible) imgSrc = src
```
nicksp: What if we define a variable right in here: `const imgSrc = isVisible ? src : null`, so that we shave off one line of code? Wdyt?

nicksp:
Hey @carlosalmonter Remember to assign you to PRs as well.

src/components/ui/Avatar.js
kararade: The `isVisible` variable is a boolean that is updated every time the component is inside (true) or outside (false) the viewport.

In this case to achieve the effect of the expected lazy-load we are not loading the image by default, but, the first time the component passes through the viewport we are loading the expected image by assigning it this way `if (isVisible) imgSrc = src`, so we also avoid removing the image or changing it again when the component is outside the viewport. 

What do you think if I add a conditional that checks some kind of counter, `if (isVisible && counter === 0) { imgSrc = src; counter++; }` , so that in this way we only do this assignment once and avoid that there is logic running in this component all the time?

src/components/ui/Avatar.js
```diff
@@ -17,19 +18,36 @@ type Props = {
...
-    {...rest}
-  />
+  const { alt, src, size } = props
+  let imgSrc = null
+  let visibleCounter = 0
+
+  return (
+    <VisibilitySensor>
+      {({ isVisible }) => {
+        if (isVisible && visibleCounter === 0) {
```
nicksp: Yeah, this is a good approach to not offload images when we scroll then beyond the viewport 👍

src/components/ui/Avatar.js
```diff
@@ -17,19 +18,36 @@ type Props = {
   alt?: string,
   src: string,
   size: number
-}
+};
```
nicksp: This is not required. We use [Standard](https://standardjs.com/) styleguide to write code. I reverted that.

nicksp:
Changed the PR base as it should be a feature branch

PR #419, opened 2019-07-24
Add sidebar heading component for quick search feature

Moharu:

Fixes https://x-team-internal.atlassian.net/browse/XP-2905

## Description

Adds the sidebar heading component that will enable the quick search functionality to be added to XP. The component has two states, defined by if the quick search input is toggled or not. It can be fully-controlled or self-controlled via `props` (check use cases in storybook).

## Checklist

## Related PRs

branch | PR
------ | ------
XP-2905-add-quick-search-to-sidebar | [XP API](https://github.com/x-team/xp-api/pull/690)
XP-2905-add-quick-search-to-sidebar | [XP](https://github.com/x-team/xp/pull/1962)

## Steps to Test or Reproduce

Check Storybook for Filters/Sidebar Heading cases.

Sidebar heading for TwoColumnLayout.

## Screenshots

xp2905components


Moharu:
If we have the time we can maybe refactor to make the component more generic / usable in instances other than just the sidebar heading.

nicksp:
@Moharu Please resolve the conflict by running `npm run build` and committing those changes.

Moharu:
> @Moharu Please resolve the conflict by running `npm run build` and committing those changes.

Just commited that along with a small change @bernardodiasc pointed out earlier, seems to still be conflicting 😕

nicksp:
> Just commited that along with a small change @bernardodiasc pointed out earlier, seems to still be conflicting

That's because, you should also remember to rebase with `develop` first, as what is bringing those conflicts 😏

nicksp:
Hey @Moharu Just resolved that, by rebasing with `develop` branch

src/components/ui/Filters/SidebarHeading.js
```diff
@@ -0,0 +1,147 @@
...
+          <a onClick={sidebarHeadingLink} className={[cx.sidebarHeadingIcon, cx.sidebarHeadingLink].join(' ')}>
+            <SvgIcon icon={sidebarIcon} color='frenchGrayDarker' />
+          </a>
+        ) : (
+          <div className={cx.sidebarHeadingIcon}>
+            <SvgIcon icon={sidebarIcon} color='frenchGrayDarker' />
+          </div>
+        ))}
+        { text }
+        <QuickSearchButton onClick={() => onToggleInput(true)} />
```
nicksp: Try to skip inline handlers as they create new function references all the time the component is being rendered which means `QuickSearchButton` is gonna be re-rendered even though it shouldn't be. Consider using class fields to eliminate that case. Please check other cases in the changeset as well.

src/components/ui/Filters/SidebarHeading.js
```diff
@@ -0,0 +1,147 @@
...
+          />
+        </form>
+        <div className={cx.sidebarHeadingIconRight} onClick={() => onToggleInput(false)} >
+          <SvgIcon icon={'x'} color='grayscarpaflow' />
+        </div>
+      </div>
+    ) : (
+      <div className={cx.container}>
+        {sidebarIcon && (sidebarHeadingLink ? (
+          <a onClick={sidebarHeadingLink} className={[cx.sidebarHeadingIcon, cx.sidebarHeadingLink].join(' ')}>
```
nicksp: 🤔 How is that we pass a string into an event handler? Am I missing anything? `sidebarHeadingLink` seems to be typed as a string instead of a function...

src/components/ui/Filters/SidebarHeading.js
```diff
@@ -0,0 +1,147 @@
...
+  render () {
+    const { isSearching, inputValue } = this.state
+    const { isQuickSearching, onToggleQuickSearch, quickSearchValue, sidebarIcon, sidebarHeadingLink, text } = this.props
+
+    // component can be either fully controlled or self-controlled
+    const shouldRenderInput = isQuickSearching || isSearching
+    const onToggleInput = onToggleQuickSearch || this.toggleQuickSearch
+    const value = quickSearchValue || inputValue || ''
+
+    return shouldRenderInput ? (
```
nicksp: Not sure about the rest of the team, but the whole rendering seems to do too much so it's hard to follow along. What if we separate rendering pieces into separate smaller functional components, so we have an additional benefit of rendering optimization? Also, the final `render` function will be much simpler to reason about as it'll be just delegating rendering parts and become more declarative.

Moharu: Hey @nicksp, can you check on the latest commit if readability of the code got better?

src/components/ui/QuickSearchButton.stories.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+
+import QuickSearchButton from './QuickSearchButton'
+
+const Body = ({ children }) => (
+  <div style={{ width: '200px', padding: '10px 0', border: '1px solid black' }}>
```
bernardodiasc: Since this is a default showcase (not a debug one), lets remove extra visual parts to not cause confusion. Under the Debug category this would be fine, but other than that, some Storybook users (as the designers) will be confused if they see something that's not expected in the design, like the border in this case.

```suggestion
  <div style={{ width: '64px', padding: '10px 0' }}>
```

src/components/ui/QuickSearchButton.stories.js
```diff
@@ -0,0 +1,15 @@
...
+    {children}
+  </div>
+)
+
+storiesOf('UI Components/QuickSearchButton', module)
+  .add('basic usage', () => <Body><QuickSearchButton onClick={action('onClick')} /></Body>)
+  .add('missing props (does component explode?)', () => <Body><QuickSearchButton /></Body>)
```
bernardodiasc: This suggestion is because ideally we want to make it clear the showcases that are intended for Debug. For a context, consider that we have different types of users in Storybook, we the developers are interested on all showcases, the default ones and the debug ones, but there's also designers and product users that will be using the tool, and for them the Debug type of showcases are not useful at all, so we keep it under a category for that reason.

Also, the `missing props (does component explode?)` should declare only the minimal render, no props and no wrappers.

```suggestion
storiesOf('UI Components/QuickSearchButton/Debug', module)
  .add('missing props (does component explode?)', () => <QuickSearchButton />)
```

src/components/ui/HeadingWithQuickSearch.stories.js
```diff
@@ -0,0 +1,55 @@
...
+storiesOf('UI Components/HeadingWithQuickSearch/Debug', module)
+  .add('with action handler', () => (
+    <Heading>
+      <HeadingWithQuickSearch
+        text='Click this arrow'
+        leftIcon='arrowleft'
+        leftIconAction={action('onClick')} />
+    </Heading>
+  ))
+  .add('missing props (component does not explode)', () => (
```
bernardodiasc: Lets remove the wrapper in this particular debug case, we don't need to worry about visual context, just need to make sure that the component doesn't explode if missing props and if the "minimal render" doesn't look like a broken UI piece (many times will be fine to be blank or even null - no markup at all).

```suggestion
  .add('missing props (does component explode?)', () => (
    <HeadingWithQuickSearch />
  ))
```

src/components/ui/HeadingWithQuickSearch.stories.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+
+import State from '../../utils/State'
+
+import HeadingWithQuickSearch from './HeadingWithQuickSearch'
+
+const Heading = ({ children }) => (
```
bernardodiasc: This wrapper is nice, I can see the benefit of it. Can we add one more Debug showcase just to show how the component looks like without it? This is good for debug purposes, we can see the how the component actually behaves in total isolation from context, and as being a Debug showcase sometimes it's fine if doesn't look so good (as in this case I know that the sizing relies on the context, which is a nice thing, and even better when this technical fact can be expressed in a debug showcase).

src/components/ui/Filters/SidebarHeading.js
```diff
@@ -0,0 +1,154 @@
...
+  onQuickSearchSubmit?: Function,
+  sidebarHeadingLink?: string
+}
+
+type State = {
+  isSearching: boolean,
+  inputValue: string
+}
+
+export default class SidebarHeading extends PureComponent<Props, State> {
```
bernardodiasc: As a standard, we do the export in the end of the file. This might be too subtle but have some benefits. The first is the consistency with the rest of the code. This consistency leverage the effort to read the code since we can take for granted how things are made and don't need to look out for exceptions. Having the export in the end of the file make it easy to check for HOCs and some especial declaration if existing, I mean, some components are really big, the class declaration can be in the middle of hundred lines, on such cases the benefit is more visible than other simpler cases, so we choose to be consistent on every case. :)

Also, please add showcases for this component in the `Filters.stories.js`. You can create 2 showcases in the Debug category: a `standalone SidebarHeading` (similar to the `standalone Filter`) and the minimal render (missing props). If you find opportunity to expose some edge case or variations, which can be the case, please include them also. :)

src/components/ui/Filters/SidebarHeading.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+  `),
+  inputContainer: cmz(`
+    background-color: ${theme.baseBrighter}
+  `),
+  input: cmz(`
+    align-self: stretch
+    width: 100%
+    height: 100%
+    border: 0px !important
```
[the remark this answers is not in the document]
bernardodiasc: Looks like this is still unresolved.

src/components/ui/Filters/SidebarHeading.js
```diff
@@ -0,0 +1,147 @@
...
+    padding-left: 4px
+    margin: 0 0 0 0
```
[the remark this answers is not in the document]
bernardodiasc: Looks like this is still unresolved.

src/components/ui/TwoColumnsLayout.stories.js
```diff
@@ -104,6 +106,16 @@ storiesOf('UI Components/TwoColumnsLayout/Debug', module)
       </StoryAdminScreen>
     </Body>
   ))
+  .add('with composed headings in AdminScreen', () => (
```
bernardodiasc: Sorry I didn't really get the benefit of this story. It looks a bit strange but I know it's a debug, maybe there's some chance to include "Notes" describing the goal here (see an example at https://github.com/x-team/xp-ui/blob/2c1a9e09d9e8ded7d0e6c73b40609dd37b8f1543/src/components/ui/MetaGroup.stories.js#L68-L76). I mean, as a developer, when I'm editing this component or just checking this story, what should I expect?

src/components/ui/Filters/SidebarHeading.js
Moharu: My bad, this whole component was moved to the new `HeadingWithQuickSearch`, just forgot to delete it :fearful:

nicksp:
FYI: Performed another build to include missing `xp-ui.js.map` file.

bernardodiasc:
@nicksp I fixed the bug with the heading link in this commit https://github.com/x-team/xp-ui/pull/419/commits/285cfff748b559c5af3d633d908843a0cd47d1ac

PR #398, opened 2019-05-22
Separate bundle for TextareaEditor

joshwnj:
This change achieves 2 things:

- avoids issues with gatsby's SSR (there are things in `medium-editor` that break due to DOM dependency)
- smaller bundle size, as `medium-editor` is quite chunky, and only needs to be loaded on the Portfolio screen.

*NOTE*: this requires some corresponding changes on the consumer side so hang tight before merging this PR. We should review them together.

PR #426, opened 2019-08-15
Multiple bundles

joshwnj:
Second attempt at #398. The goal is to produce 3 bundles:

- `all`: Contains everything. This remains the default, so that we preserve backwards compatibility.
- `editor`: Contains only the rich-text editor, so it can be lazy-loaded.
- `fonts`: Contains only the `webfontloader` part (which is incompatible with gatsby ssr)
- `registration`: Contains only the components used by the registration site.

Consumers can get the `all` bundle by doing `import { Thing } from "xp-ui"` as normal.  But they can also do `import { Thing } from "xp-ui/lib/registration"` if we only need components related to the registration site.

When we merge this PR, we can:

- close #398 without merging
- delete the `#lite` branch

as they will no longer be needed.

## Description

- updates webpack config to produce multiple bundles
- includes some commits from #398 

## Checklist

## Related PRs

- [x] need to make a corresponding PR in `xp-registration`
  - https://github.com/x-team/xp-registration/pull/58


lib/.babelrc
```diff
@@ -0,0 +1 @@
+{}
```
nicksp: do we need to keep this empty file as well as empty `eslintrc`? Is it required?

joshwnj: I added this because under some conditions, I was seeing babel errors from babel trying to parse the files in `lib/`.  I didn't explore far enough to see exactly what causes it, but I think it's related to using `npm link` with certain versions of npm.

Anyway, the intent is to show that nothing in `lib/` should be babel-ified.

nicksp: Got it @joshwnj. Also, as an option, to disable babel compilation of specific files, we could use `ignore` option as stated here: https://babeljs.io/docs/en/options#ignore

nicksp:
Hi @joshwnj Please notice that I rebased with `develop` branch properly to exclude unrelated changes that are already inside the branch. Also, I rebuilt the app due to that. Now testing to make sure there're no regressions 🤞

PR #429, opened 2019-08-29
Update Search Sidebar when working with Quick Search

bernardodiasc:
**Release Type:** *No breaking change*

Fixes:

- https://x-team-internal.atlassian.net/browse/XP-3283
- https://x-team-internal.atlassian.net/browse/XP-3282

## Description

- Created new component QuickSearchCloseButton
- Included a way to activate focus for InputField, so that when clearing HeadingWithQuickSearch the focus goes back to the input
- Fixed height behaviour of Filter Container and GenericCollapsible Body

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-3283-update-search-sidebar-layout-when-working-with-quick-search` | [XP](https://github.com/x-team/xp/pull/1984)

## Steps to Test or Reproduce

1. Check the new component at http://localhost:9001/?selectedKind=UI%20Components%2FQuickSearchCloseButton
1. Existing components should not have any regression:
    - http://localhost:9001/?selectedKind=UI%20Components%2FHeadingWithQuickSearch
    - http://localhost:9001/?selectedKind=UI%20Components%2FFilters
    - http://localhost:9001/?selectedKind=UI%20Components%2FGenericCollapsible
    - http://localhost:9001/?selectedKind=Form%20Components%2FInputField

- forms/InputField.js
- ui/Filters/Container.js
- ui/GenericCollapsible/Body.js
- ui/HeadingWithQuickSearch.js
- ui/QuickSearchCloseButton.js
- ui/QuickSearchCloseButton.stories.js

## Screenshots

Please refer to https://github.com/x-team/xp/pull/1984 Screenshot section


src/components/ui/QuickSearchCloseButton.js
```diff
@@ -0,0 +1,58 @@
...
+
+import Button from './Button'
+
+import type { Element } from 'react'
+
+const cmz = require('cmz')
+
+type Props = {
+  onClick: (event: SyntheticEvent<HTMLButtonElement>) => void,
+  children: Element<*>,
```
nicksp: Bernardo, the `React$Node` type is the type for the `children` props of React components, which includes all of the possible cases. Did you avoid using it for a specific reason?

https://github.com/facebook/flow/blob/c7062e9ae0ac9bce43e8592e56186c58709513bd/lib/react.js#L8-L20

https://github.com/facebook/flow/blob/c7062e9ae0ac9bce43e8592e56186c58709513bd/lib/react.js#L262

src/components/forms/InputField.js
```diff
@@ -455,10 +459,17 @@ class InputField extends PureComponent<Props> {
       ...baseProps,
       className: `${errorClassName} ${spacingClassName}`,
       type,
-      ...rest
+      ...rest,
+      ref: input => (this.inputElement = input)
```
nicksp: This way of creating refs is still supported, but to give you some more perspective, there's a new way of working with [Refs](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs). There's even an example of [focusing on the input](https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element). Not a change request, but more of pointing out to other possible options. So up to you which one you prefer 😏

nicksp:
@bernardodiasc Please also rebase with latest `develop`, packages will need to be reinstalled since we upgraded some of them. And then rebuild.

src/components/ui/QuickSearchCloseButton.js
```diff
@@ -0,0 +1,58 @@
...
+        onClick={onClick}
+      >
+        {children}
+      </Button>
+    </div>
+  )
+}
+
+QuickSearchCloseButton.defaultProps = {
+  onClick: () => {},
```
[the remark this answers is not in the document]
bernardodiasc: I wonder if running type check on XP does look up on XP-UI type notations, because if not, then it's probably not safe to define props as required, the flow check will not spot wrong usages in the XP, therefore this creates an opening for breakable code. I'd rather leave the button click doing nothing than let it crash the entire app.

src/components/ui/QuickSearchCloseButton.js
bernardodiasc: no particular reason, just kept consistent with how we have been using previously. will update this, thanks for the info

src/components/ui/QuickSearchCloseButton.js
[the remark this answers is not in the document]
nicksp: We run type checks on XP-UI, so should expectations from there. If the function prop isn't required, then there's no need to set it's value to noop function (`() => {}`) which means nothing. I'd rather do a simple `if (function) { function () }` which is more logical.

src/components/ui/HeadingWithQuickSearch.js
```diff
@@ -107,6 +116,15 @@ class HeadingWithQuickSearch extends PureComponent<Props, State> {
     }
   }
 
+  handleClearValue = (): void => {
+    if (this.props.onQuickSearchChangeValue) {
+      this.props.onQuickSearchChangeValue('')
+    } else {
+      this.setState({ inputValue: '' })
+    }
+    this.focusOnSearchInput()
```
nicksp: You still have focus as part of clearValue method, so mixed concerns here. Instead, you should call both `handleClearValue` and `focusOnSearchInput` from the consumer side, so each one does one thing only.

bernardodiasc: you are suggesting creating a 3rd method?

nicksp: For example, so the method defines a set of actions to perform to handle that. But please look into other possibilities.

bernardodiasc:
@nicksp the code review requests was updated in the commits https://github.com/x-team/xp-ui/pull/429/commits/d428adaf3d358dab6d9f0b3d256be39e57544af6 and https://github.com/x-team/xp-ui/pull/429/commits/c544fb78b271e28a83ce8cf9c20db472d44f5dc6, this PR is ready for another round. Thanks for the review&qa!

nicksp:
@bernardodiasc When you have a moment, could you please resolve build issues?

image

bernardodiasc:
> @bernardodiasc When you have a moment, could you please resolve build issues?

That's exactly the issue I had during the day... I don't know what's going on but I can propose a fix that bypass the issue.

nicksp:
> That's exactly the issue I had during the day... I don't know what's going on but I can propose a fix that bypass the issue.

Right, I thought you said you could resolve it with clearing npm cache. Why didn't I get it locally? Sorry if I'm not that deep into the problem.

bernardodiasc:
> Right, I thought you said you could resolve it with clearing npm cache. 

Looks like it solved locally, I'm not having that issue anymore o my machine.

nicksp:
> Looks like it solved locally, I'm not having that issue anymore o my machine.

Any local changes that you will probably need to push to remote? I'd suggest you clear local npm cache, ensure you have Node `v.9.3.0`, then `rm -rf node_modules`, then `npm i`, then `npm run build` and then `npm t`. After that, if there're any changes, you do `git add .` and commit that, so Travis rebuilds that.

bernardodiasc:
Nice, thanks @nicksp. I made the last adjustments on https://github.com/x-team/xp/pull/1984, so both PRs should be ready.

PR #433, opened 2019-09-03
Organise Styleguides in the Storybook

bernardodiasc:
**Release Type:** *Non breaking changes*

Fixes https://x-team-internal.atlassian.net/browse/XP-3284

## Description

Organise Styleguides in the Storybook

## Checklist

## Steps to Test or Reproduce

Go to:

- ColorPalette
  - Showcase: http://localhost:9001/?path=/story/styleguides--colorpalette
  - Notes: http://localhost:9001/?path=/info/styleguides--colorpalette
- Typography
  - Showcase: http://localhost:9001/?path=/story/styleguides--typography
  - Notes: http://localhost:9001/?path=/info/styleguides--typography
- Layout
  - Showcase: http://localhost:9001/?path=/story/styleguides--layout
  - Notes: http://localhost:9001/?path=/info/styleguides--layout

#### Before

previous-color-palette

#### Before

screenshot-x-team github io-2019 09 04-13_39_24

#### Before

Doesn't have.


.storybook/config.js
```diff
@@ -93,9 +83,13 @@ addParameters({
 })
 
 // Load stories dynamically
-const req = require.context('../src/components', true, /\.stories\.js$/)
+const styleguides = require.context('../src/components/styleguides', true, /\.stories\.js$/)
+const components = require.context('../src/components/ui', true, /\.stories\.js$/)
+const forms = require.context('../src/components/forms', true, /\.stories\.js$/)
 function loadStories () {
-  req.keys().forEach(filename => req(filename))
+  styleguides.keys().forEach(filename => styleguides(filename))
```
nicksp: Is there any specific reason we introduced repetitive code here and right above? The problem with such approach is that it easyly gets out of control when we introduce more component groups. Having a generic solution, if possible, is preferred. I tried this and it works great:

```
const req = require.context('../src/components', true, /\.stories\.js$/)
function loadStories () {
  req.keys().forEach(filename => req(filename))
}
```

This way, any time we introduce new groups, we shouldn't remember to make changes in this config, otherwise we run into risks of forgetting to include that.

If your intention was to show `Styleguides` as a first top group, then there're other approaches to changes visible order. I resolved that by renaming `styleguides` to `_styleguides` and it serves the purpose:

image

Here's how it looks:

[Screenshot from Gyazo](https://gyazo.com/25b23fa1c1b22b2212ef76d951edc64d)

bernardodiasc:
> BTW, could you see updated favicon? I tried hard reloading and in incognito mode and it still shows a default one for me

I could not see in the dev instance as well, but you can confirm it works with the build.

.storybook/config.js
bernardodiasc: I don't like much this approach of creating an exception with the folder name. I tried to make the loading more dynamic, but the goal here is have full control of the order that the categories appear, so it will be hardcoded anyway.

nicksp: To be honest, I don't think we need to specify an order here, so let Storybook order alphabetically. This way ew have a dynamic and reliable way of loading. If there's a real need to introduce custom order, we can name folders with some prefixes like `1_`, `2_`, etc. Otherwise, unless it's required let's strive for cleaner code,

nicksp:
> I could not see in the dev instance as well, but you can confirm it works with the build.

Ah I see, seems like an issue referencing a file path. Seems like changing assets path to `./public` tends to work 😄 Please check on your side and update respectively.

bernardodiasc:
> Ah I see, seems like an issue referencing a file path. Seems like changing assets path to ./public tends to work 😄 Please check on your side and update respectively.

There's no `public` folder in the build, the storybook copies the files from that folder and put in the static root. The problem with the dev instance is that it uses a storybook favicon, you can check the source code of the `<head>` and observe that the favicon.ico is replaced but the favicon.png is there.

nicksp:
> There's no public folder in the build

? I can see this folder and it seems working fine for me with changes I commented about. Please test locally.

bernardodiasc:
It doesn't work for me:

Static files from build:

Screen Shot 2019-09-05 at 11 12 29

Dev instance:

Screen Shot 2019-09-05 at 11 13 50

PR #436, opened 2019-09-11
Refactor and cleanup react components related to the old search

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3221

## Description

Moved and refactored XP components over to XP-UI.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-3221-refactor-and-cleanup-react-components-related-to-the-old-search` | [XP](https://github.com/x-team/xp/pull/1993)

### StatusFilter

1. Check and interact default state with http://localhost:9001/?path=/story/ui-components-statusfilter--default
1. Check disabled state http://localhost:9001/?path=/story/ui-components-statusfilter--disabled

### StatusMarker

- Check on knobs http://localhost:9001/?path=/story/ui-components-statusmarker--basic-usage-see-knobs

### ApplicantProfileContentDisplay

> This story looks empty right now because this component is not yet optimised. As agreed we'll need a second iteration on it to get a final version.

1. http://localhost:9001/?path=/story/ui-components-applicantprofile-applicanttopprofiledisplay--basic-usage

### ApplicantTopProfileDisplay

> This story looks empty right now because this component is not yet optimised. As agreed we'll need a second iteration on it to get a final version.

1. http://localhost:9001/?path=/story/ui-components-applicantprofile-applicantprofilecontentdisplay--basic-usage


nicksp:
Rebased with latest `develop`.

bernardodiasc:
@nicksp that's strange, I repeated twice the very same steps and it's working on my machine.

nicksp:
@bernardodiasc NVM, I resolved the problem. Had to reinstall XP's dependencies from scratch 😄

src/components/ui/StatusFilter.js
```diff
@@ -0,0 +1,88 @@
...
+
+const STATUS_ACTIONS = {
+  accepted: 'Accept',
+  excluded: 'Exclude'
+}
+
+class StatusFilter extends PureComponent<Props, void> {
+  static defaultProps = {
+    disabled: true,
+    checked: {
```
nicksp: is it important to provide defaults, since we required `checked` to be provided anyways? Otherwise the compile-time Flow error will pop up. Same in other cases, when you provide faults for required props. Or is it done on purpose?

src/components/ui/StatusFilter.js
```diff
@@ -0,0 +1,88 @@
...
+      accepted: false,
+      excluded: false
+    }
+  }
+
+  handleClick = (status: ApplicantStatusType) => () => {
+    const { handleCheck, disabled } = this.props
+    if (!disabled && handleCheck) {
```
nicksp: `handleCheck` also seems to be required props, so the condition can be simplified. That's how it worked being in XP as I can see. Please check other props that are required in your changeset but you still do provide defaults for them. Those are good candidates to be simplified.

nicksp: Shouldn't `handleCheck` be required as it was before? Do we really expect the app to have a use case when it's missing? I think the previous implementation we had was correct and we just needed to simplify the condition. Wondering what was your reasoning to change the props API.

bernardodiasc: The flow type checking is not that safe when it comes to enforce props, because we can't check the implementation over the repos. E.g. someone forgot to implement a required prop on XP, running flow-type checks there will help nothing. Therefore I also wouldn't expect to crash the app by a missing prop.

nicksp: I'm wondering what was the reason to change the previous implementation? Was it in the scope of changes? I'd avoid doing any changes that aren't expected or tested properly. The scope is to do refactor without changing behaviour. So, I'd suggest leaving doing any type changes for a follow-up proposal. Otherwise, we're forced to check all the change you did which is too much. Please revert those.

bernardodiasc: There wasn't any change in the behaviour. Also all E2E tests that touch this component is passing. Not sure I get your concern.

nicksp:
My concern isn’t to touch internal implementation details if the work you’re on is to move components and do some clean up. Let’s keep types and conditions as they were before and raise bug tickets if you’d like to refine types. Otherwise, we’re introducing potential points of failure.



Sent from myMail for iOS


Monday, 16 September 2019, 15:24 +0300 from notifications@github.com  <notifications@github.com>:

bernardodiasc:
We are not just moving components, many of the cases that's not possible, we need to refactor and adapt to the 2-repos reality.

bernardodiasc:
@nicksp the changes you request doesn't make sense to me. It's wrong to think that we can simply copy&paste from one repo to another, thats not going to work and we'll see ourselves discussing the very same thing on every XP-to-XPUI refactor we are going to make. But feels pointless to spend more time on this PR, then I applied the changes to the type definitions you asked.

- The flow type for props on XP-UI doesn't help anything if something causes wrong usage of the component API on XP.
- By moving components over to XP-UI we are defining an UI API, which means user stuff, so it's wrong to not deal possible undefined props, because we don't want to cause white screen of death because a code mistake.

So it's not useful to not allow and check for undefined props, every prop can possibly be undefined. The mindset for functional programming is not exactly the same as for React UI components.

PR #439, opened 2019-09-18
Improve GenericCopyToClipboard component

bernardodiasc:
**Release Type:** *Non breaking changes*

Fixes https://x-team-internal.atlassian.net/browse/XP-3332

## Description

Improve GenericCopyToClipboard component

## Checklist

## Steps to Test or Reproduce

1. Check the component at http://localhost:9001/?path=/story/ui-components-genericcopytoclipboard--basic-usage

- GenericCopyToClipboard


src/components/ui/GenericCopyToClipboard/GenericCopyToClipboard.stories.js
```diff
@@ -4,28 +4,37 @@ import { storiesOf } from '@storybook/react'
 import GenericCopyToClipboard from '.'
```
nicksp: Since we started adding flow annotations to story files, I suggest we do the same here as well as it reveals some good cases of incorrect component usage and things we should improve in props handling.

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -1,58 +1,82 @@
...
+    position: relative
+    color: ${theme.typoHighlightOnDarkBackground}
+  `),
+
+  bubble: cmz(`
+    & {
+      position: absolute
+      white-space: nowrap
+      bottom: 100%
+      box-shadow: 0px 2px 17px rgba(171, 183, 193, 0.25)
```
nicksp: `0px` could be unitless, but not a strict change requirement, just to make it consistent.

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -1,58 +1,82 @@
...
+export default function Tooltip ({ copied }: Props) {
+  return (
+    <span className={cx.container}>
+      <span className={cx.bubble}>
+        <span className={cx.message}>
+          {copied
+            ? (
+              <Fragment>
+                <SvgIcon icon='check' />
+                <span className={cx.copied}>Copied to Clipboard</span>
```
nicksp: As I can see in AC, the tooltip should read: `Copied to clipboard`

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -37,19 +39,31 @@ class GenericCopyToClipboard extends Component<Props, State> {
...
     return text ? (
-      <div className={container}>
-        <Tooltip showTooltip={showTooltip} />
+      <div
+        className={container}
+        onMouseEnter={this.handleMouseEnter}
+        onMouseLeave={this.handleMouseLeave}
+      >
+        {isHover || copied ? <Tooltip copied={copied} /> : null}
         <CopyToClipboard text={text} onCopy={this.handleCopy()}>
```
nicksp: I might be not understanding smth, but is there any point of assigning a callback like that? Wouldn't it work if we just declare it like: `onCopy={this.handleCopy}` and update handler definition to be: `handleCopy = () => {...}`?

Please let me know if I'm missing an implicit goal here.

bernardodiasc: I believe there's no point in using that way. I haven't touched this part and didn't paid attention. But will make the adjustment, good catch 👍

nicksp: I know, it was a previous caught. But, it's just now, when I realized this bit looks confusing to me 😆

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -1,58 +1,82 @@
...
+  `),
+
+  copied: cmz(`
+    padding: 0 0 0 5px
+  `)
+}
 
 type Props = {
-  showTooltip: boolean
+  copied?: boolean
```
nicksp: seems like we're missing defaultProp value for optional `copied` prop.

src/components/ui/GenericCopyToClipboard/index.js
```diff
@@ -15,19 +15,21 @@ const container = cmz(`
 `)
 
 type Props = {
-  children: Node,
-  text: string
+  children?: Node,
```
nicksp: These are confusing. I think I got the point of types now after reading through several resources.

Node [type]() already assumes null values. Instead of specifying props as optional, we should provide defaults. You should not type any prop as optional if you are going to include a default value for it. If you declare it as optional, Flow will alert of undefined, but this is **impossible**, because the prop has a default value ;). That's why that shouldn't be done.

From the component's usage perspective, Flow is smart enough to understand that a prop is optional when you include the default value for it, so that you can render it without passing that prop.

Check it out:

https://flow.org/try/#0PTAEAEDMBsHsHcBQiCWBbADrATgF1AFSgCGAzqAEoCmxAxvpNrGqAOTY32sDcyuAnhiqgACkwzkAvKADeiUKEixYALlAA7AK5oARlWyhEAX2S1oZcgFl+AYWZZ1VdfioAPXE4Am5anVwA6O0xYR2cAHjFYCQA+WXlQUlxiXBRaUE8qSGJNaFxIiVBpOQUFJVVQAEZ4kxNEEFAAQVyAC1hNAHNm0AADMu7QFHJ1WHwolJDiaAH1UFxm4W780n6BIXTMlHUUcfU6sDnB9NgqUnVWBmIUKb1abNJhADE4eAHyUjRiPFAnNs7Z2FAmnUGWwiWIwL2s2ayVA-DaRzO+EcVE8-1AGAsPT6oHmHFANzuwjm+iorHIxB6GSyOTy4mWihwA3w3AAlIgwtYgg4nPhgNFkEA

So, make props "required", but do provide default prop values for crucial ones:

```js
type Props = {
  children: Node,
  text: string
}

class CustomComponent extends Component {
  static defaultProps = {
    ... defaults here
  }
}
```

bernardodiasc: Glad that we are on same page now :)

bernardodiasc:
@nicksp indeed the hiding behaviour after copied was wrong, it should be fixed now. Thanks for checking. I also fixed the vertical alignment of the icon+text, it was a bit off.

PR #440, opened 2019-09-18
Add new copy and checkmark svg icons

rayepps:
**Release Type:** *Non-Breaking Feature*

Fixes https://x-team-internal.atlassian.net/browse/XP-3331

## Description

Add new svgs to SvgIcon component for the new  `copy` and `checkmark` icons that have been added to the latest designs.

## Checklist

- [x] set yourself as an assignee
- [x] set appropriate labels for a PR (`In Review` or `In Progress` depending on its status)
- [x] move respective JIRA issue to the `IN REVIEW` column
- [x] make sure your code lints (`npm run lint`)
- [x] Flow typechecks passed (`npm run flow`)
- [x] Snapshots tests passed (`npm run jest`)
- [x] check cleanup tasks (https://github.com/x-team/xp/labels/cleanup) and take a suitable small one (if exists) in a related area of the current changes
- [x] component's documentation (`.stories.js` file) is changed or added accordingly to reflect any new or updated use cases or variants usage
- [x] if any snapshots have been changed, verify that component still works and looks as expected and update the changed snapshot
- [x] **manually tested the app** by running it in several different browsers (Firefox, Chrome, Opera, Safari, MS IE/Edge, etc.) and checked nothing is broken and operates as expected!

## Steps to Test or Reproduce

Check the component at https://x-team.github.io/xp-ui/?path=/story/ui-components-svgicon--basic-color

## Impacted Areas in Application

`SvgIcon`

## Screenshots

Screen Shot 2019-09-18 at 11 39 42 PM
^^ See new green color above with updated check mark and new copy icon at end

PR #443, opened 2019-09-19
Update Shortcuts

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-2894

## Description

Upgrade Shortcuts and related components.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-2894-upgrade-shortcuts-with-updated-genericcopytoclipboard` | [XP](https://github.com/x-team/xp/pull/1999)

## Steps to Test or Reproduce

- http://localhost:9001/?path=/story/ui-components-shortcuts--basic-usage (it has dynamic sample data, so when reloading the page you can check with text of different sizes and observe it doesn't break line and for any case will look strange)
- http://localhost:9001/?path=/story/ui-components-genericcopytoclipboard--with-custom-x-offset (this is not very dynamic solution but works for our case, the small target elements need the Tooltip to be positioned with a bit negative X offset)
- http://localhost:9001/?path=/story/ui-components-svgicon--basic-color (had to update `redirect` and `check` SVGs)

- src/components/ui/GenericCopyToClipboard/GenericCopyToClipboard.stories.js
- src/components/ui/GenericCopyToClipboard/Tooltip.js
- src/components/ui/GenericCopyToClipboard/index.js
- src/components/ui/Shortcuts.js
- src/components/ui/Shortcuts.stories.js
- src/components/ui/SvgIcon.js

## Screenshots

> Something that we seem to skip during the review of https://github.com/x-team/xp-ui/pull/440, the SVG of the `copy` doesn't had the inside coloured correctly:

Screen Shot 2019-09-19 at 09 25 41

> So I get the SVG from Figma again: /cc @rayepps 

Screen Shot 2019-09-19 at 09 24 44

> Resulted in:

Screen Shot 2019-09-19 at 09 26 05

> Also the `redirect` link seemed wrong, we didn't discussed about changing this, but I believe its a good move and worth to check with Paul about it, so far it was used only in 1 place:

Screen Shot 2019-09-19 at 11 31 59

> But that doesn't match with the expectation for Shortcuts:

Screen Shot 2019-09-19 at 11 32 47

> So this is how it looks with the new SVG for `redirect` in **Storybook**:

> And this is how it looks everything in **XP app**:


src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -83,7 +87,8 @@ const Tooltip = ({ copied }: Props) => {
 }
 
 Tooltip.defaultProps = {
-  copied: false
+  copied: false,
+  tooltipXOffset: 0
```
nicksp: Just a quick note here, we remember we shouldn't provide defaults to every prop, only to important ones, right 😏 ?

src/components/ui/Shortcuts.js
```diff
@@ -54,84 +50,106 @@ const cx = {
...
+          <SvgIcon icon='redirect' color='grayscale' className={cx.external} />
+        )}
+        {label}
+      </a>
+      {!external && (
+        <GenericCopyToClipboard
+          key={value}
+          text={value}
+          className={cx.showCopy}
+          tooltipXOffset={-30}
```
nicksp: seems like we're simply hardcoding that here. Can we consider using that directly in a component, so we don't pass extra props until needed by other use cases? So we remove `tooltipXOffset` altogether as there seem to not much value out of using that in our current case 👍

src/components/ui/SvgIcon.js
```diff
@@ -655,9 +655,8 @@ export const getIcons = ({ color, hover, ...rest }: { [x: string]: any } = {}) =
     ),
 
     redirect: (
-      <svg width='16' height='15' viewBox='0 0 16 15' fill='none' {...rest}>
-        <path className={strokeClassName} fillRule='evenodd' clipRule='evenodd' d='M13 13H2.5V2.5H7.75V1H2.5C1.6675 1 1 1.675 1 2.5V13C1 13.825 1.6675 14.5 2.5 14.5H13C13.825 14.5 14.5 13.825 14.5 13V7.75H13V13Z' />
-        <path className={strokeClassName} fillRule='evenodd' clipRule='evenodd' d='M9.68 1V2.5H12.3725L5 9.8725L6.0575 10.93L13.43 3.5575V6.25H14.93V1H9.68Z' />
+      <svg width='13' height='14' viewBox='0 0 13 14' fill='none' {...rest}>
```
nicksp: what is the reason we introduce that out of scope change here and for other icon? We had a dedicated PR fr that yesterday and things were approved. But now we're doing changes without having a clear reason.  I'm afraid if you're trying to account for Shortcuts use case, we might end up doing the same for other use cases. So, let's provide reasoning fordoing that and explain whether it's gonna break other places. Ideally, I'd prefer raising a bug ticket to address this change and work in a separate PR.

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -60,12 +63,13 @@ const cx = {
...
-  copied: boolean
+  copied: boolean,
+  tooltipXOffset: number
 }
 
-const Tooltip = ({ copied }: Props) => {
+const Tooltip = ({ copied, tooltipXOffset }: Props) => {
   return (
-    <span className={cx.container}>
+    <span className={cx.container} style={{ left: `${tooltipXOffset}px` }}>
```
nicksp: Consider moving style declaration to its own variable as we usually do, so we don't create a new object on each render and thus don't re-render children.

src/components/ui/SvgIcon.js
bernardodiasc: I explained that in the PR description, the other SVG didn't worked as expected.

src/components/ui/Shortcuts.js
bernardodiasc: This value is not default, it's precisely the value needed for this implementation in Shortcuts. On other places it uses different values (when not the default `0`).

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardodiasc: This is important, why wouldn't be? :)

nicksp: Didn't mean this one, was just a message to ensure we understand the purpose of defaultProps.

src/components/ui/Shortcuts.js
nicksp: Cool 👍 Just saw it's the only one in this PR.

src/components/ui/SvgIcon.js
nicksp: Makes sense 👍

src/components/ui/Shortcuts.js
bernardodiasc: Yeah, I mentioned that in the PR description, this is not super ideal solution. But I wouldn't have time to create a fancy dynamically positioned tooltip, so instead I had to use this approach. You can see on another PR how it's used with different values, e.g. https://github.com/x-team/xp-ui/pull/444/files#diff-4dca92abdfa972f8975233e168d0b7ccR54

nicksp:
Seems like we can re-use this component that we created in another PR: CopyToClipboardButton.js, otherwise why would we have it abstracted 😏

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardodiasc: Got it. Well, these default values are the default values. They seem correct because it's the values the component need if don't get any. Maybe the `copied` could be removed, since `undefined` also results false, but that's a bit strange because the component actually expects `false` as the default. Wdyt?

nicksp: How other libs use defaultProps and optional/required params?

bernardodiasc:
> Seems like we can re-use this component that we created in another PR: CopyToClipboardButton.js, otherwise why would we have it abstracted

Not really, I also considered that, but they have different styles. This one was made accordingly the PencilButton, it's made to fit those ProfileForm places. The Shortcuts implementation is kinda unique for this place.

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardodiasc: I don't know for sure. Default props is not complicated concept. Basically the default values if none are provided. My understanding is that whenever `undefined` is not expected, we should set defaults. Because the `undefined` can be used in the logic, but sometimes can't.

As an example, for a component that receives async data, if the data doesn't exist yet the prop could receive `undefined` and the logic could understand that. But for a place like the `tooltipXOffset` where `undefined` would cause broken results, it seems to me important to have a default value, in this particular case the `0` is used as a number, not as "falsy" value.

bernardodiasc:
@nicksp good catch with the copy effect, I send the fixes in the commit https://github.com/x-team/xp-ui/pull/443/commits/1e7c563851cc5a13827e2d0cb3c4c9dc2ba3ff87, please take a look when you have a chance. Thanks!

bernardodiasc:
@nicksp I fixed the styles to match with previous state (not the current from figma) in this commit https://github.com/x-team/xp-ui/pull/443/commits/e8751a045baae714b58af4892c67a4d02fc58b38

PR #444, opened 2019-09-19
Add CopyToClipboardButton component

bernardodiasc:
**Release Type:** *Non-breaking feature*

Fixes https://x-team-internal.atlassian.net/browse/XP-3335

## Description

Add CopyToClipboardButton component.

**Note:** Please merge https://github.com/x-team/xp-ui/pull/443 before this one.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-2894-upgrade-shortcuts-with-updated-genericcopytoclipboard` | [XP UI](https://github.com/x-team/xp-ui/pull/443)
`XP-3335-upgrade-fullname-with-updated-genericcopytoclipboard` | [XP](https://github.com/x-team/xp/pull/2000)

## Steps to Test or Reproduce

1. Check the component at http://localhost:9001/?path=/story/ui-components-copytoclipboardbutton--basic-usage

New `CopyToClipboardButton` included.


src/components/ui/CopyToClipboardButton.js
```diff
@@ -0,0 +1,62 @@
...
+    hover: 'default'
+  }
+
+  render () {
+    const { text, visible, color, hover } = this.props
+    const visibleClassName = visible ? cx.copyButtonVisible : ''
+
+    return text ? (
+      <span role='button' className={`${cx.copyButton} ${visibleClassName}`}>
+        <GenericCopyToClipboard text={text} tooltipXOffset={-43}>
```
nicksp: Let's make it component reusable so we can integrate it in other places as well. This will mean dynamically specifying `tooltipXOffset` per use case.

src/components/ui/GenericCopyToClipboard/Tooltip.js
```diff
@@ -60,12 +68,13 @@ const cx = {
...
-  copied: boolean
+  copied: boolean,
+  tooltipXOffset: number
 }
 
-const Tooltip = ({ copied }: Props) => {
+const Tooltip = ({ copied, tooltipXOffset }: Props) => {
   return (
-    <span className={cx.container}>
+    <span className={cx.container} style={{ left: `${tooltipXOffset}px` }}>
```
nicksp: Let's extract `style` prop into a variable due to the same reason I explained in another PR.

src/components/ui/CopyToClipboardButton.js
bernardodiasc: Sorry, I didn't get what you meant. Can you be more specific about what changes you expect?

nicksp: Let's use this new component in other places as well: shortcuts, email.

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardodiasc: I'll change this in the scope of https://github.com/x-team/xp-ui/pull/443, notice that only this commit https://github.com/x-team/xp-ui/pull/444/commits/9b90e7ae296411e235d0c9640d541b40537f9b23 is for this PR, the others is from the https://github.com/x-team/xp-ui/pull/443 :)

nicksp: whatever you prefer, just mentioned for the sake of review ;-)

src/components/ui/CopyToClipboardButton.js
bernardodiasc: I commented about that here https://github.com/x-team/xp-ui/pull/443#issuecomment-533485390, Shortcuts needs a unique implementation because of styles. For email, that's already being used, you chan see here https://github.com/x-team/xp/pull/2001

src/components/ui/GenericCopyToClipboard/Tooltip.js
bernardodiasc: No problem, just wanted to make sure that you are aware of the branches overlapping :)

src/components/ui/CopyToClipboardButton.js
bernardodiasc: That's the problem when design thinking doesn't work together with development team. This extra abstraction would only work if designers guarantee both places are in fact the same. And we can tell they are not just looking that how they are implemented on each places.

We already have `GenericCopyToClipboard` and `SvgIcon` abstractions in place, creating an abstraction on top of that which is flexible enough to handle different implementations need to be made carefully.

> Will providing styles when calling this component solve the problem you mentioned?

It's possible to do, of course, but this doesn't seem beneficial approach. The only thing they have in common is `GenericCopyToClipboard` and `SvgIcon`, which are already abstractions, but not even the values used on these components are the same in `Shortcuts` and `CopyToClipboardButton`. As far as I read in many different sources, scalable development doesn't necessarily means abstract all the things, because sometimes this approach leads to entangled codebase.

I'll make another try, but as far as I can tell it will make `CopyToClipboardButton` unnecessarily complex. Or maybe this component should be renamed to `ProfileFormCopyToClipboardButton` to make it explicit that is not a generic thing. `Shortcuts` is not a `ProfileForm` element.

Well, this idea doesn't sound good to me for all the reasons above but let me try again, will get back with something.

nicksp: I understand. Why did we introduce this new abstraction then if we can't reuse it everywhere? Consider removing if you wish, so we use copy to clipboard consistently.

bernardodiasc: > Why did we introduce this new abstraction then if we can't reuse it everywhere?

It supposed to be reused on all ProfileForm elements, along with PencilButton. It is in fact as you can see in this PR https://github.com/x-team/xp/pull/2001. The Shortcuts is the only exception that can't reuse that.

bernardodiasc:
Hey @nicksp, if you agree with this comment https://github.com/x-team/xp-ui/pull/444#discussion_r326574598, then this PR should be ready. I gave a second try on how Shortcuts could use CopyToClipboardButton. Unfortunately the implementation on these places are different. Making CopyToClipboardButton flexible enough for Shortcuts will not help much and it will raise unnecessary complexity. I know the name "CopyToClipboardButton" make it look generic, but as the "PencilButton", they are intended for ProfileForm elements and doesn't seem appropriate to force it to be generic for every CopyToClipboard usages. Please let me know what you think about.

bernardodiasc:
I resolved conflicts on this PR, should be ready for review.

src/components/ui/CopyToClipboardButton.stories.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
```
nicksp: As agreed before, let's remember to add Flow typings check to story files as well.

src/components/ui/PencilButton.stories.js
```diff
@@ -3,27 +3,32 @@ import { storiesOf } from '@storybook/react'
 
 import PencilButton from './PencilButton'
 
+const Body = ({ children, background }) => (
```
nicksp: Let's add `// @flow` here as well.

src/components/ui/Shortcuts.stories.js
```diff
@@ -18,7 +18,7 @@ const Body = ({ children }) => (
 
 const seedCount = 7
 const links = Array(seedCount).fill({}).reduce((acc, cur) => {
```
nicksp: `// @flow`

bernardodiasc: this one already have flow 👍

PR #453, opened 2019-10-08
Update JobsGrid component

bernardodiasc:
**Release Type:** *Non-breaking change*

Fixes https://x-team-internal.atlassian.net/browse/XP-3365

## Description

Update JobsGrid component

## Checklist

## Steps to Test or Reproduce

- Run `npm start`
- Go to http://localhost:9001/?path=/story/ui-components-xp-registration-jobsgrid--default
- Also check http://localhost:9001/?path=/story/ui-components-xp-registration-jobspagelayout--default-usage-full-composition

`JobsGrid`

## Screenshots

Screen Shot 2019-10-08 at 14 50 29

Screen Shot 2019-10-08 at 14 54 12


src/components/ui/ApplicantScreen.js
```diff
@@ -1,30 +1,44 @@
...
+    <div>
+      <HeaderBar />
+      {contentWrapper ? (
+        <div className={cx.content}>{children}</div>
+      ) : children}
+    </div>
+  )
+}
+
+ApplicantScreen.defaultProps = {
```
rayepps: `defaultProps` is on track to be deprecated. I recommend we make a habit of setting default props without depending on this api and instead use the recommended es6 default value syntax. Something like:
```
const ApplicantScreen = ({
  contentWrapper = true
}: Props) => {
  ...
}
```
https://twitter.com/dan_abramov/status/1133878326358171650?lang=en
https://github.com/reactjs/rfcs/pull/107

src/components/ui/JobsPageLayout.js
```diff
@@ -0,0 +1,129 @@
...
+          ) : (
+            <div className={cx.content}>{content}</div>
+          )}
+        </div>
+      </div>
+    </div>
+  )
+}
+
+JobsPageLayout.propTypes = {
```
rayepps: Just a general question, why do we use both flow and prop-types?

src/components/ui/ApplicantScreen.js
bernardodiasc: That's very nice to know, thanks for sharing references @rayepps. Indeed much simpler syntax, will make the changes. I wonder how this affect the `class` declarations.

src/components/ui/JobsPageLayout.js
bernardodiasc: The Flow have a limitation that only check types on dev instance. The XP-UI have components that is intended to use on other repositories so Flow can't guarantee checks in those consumer. The PropTypes on other hand expose problems with implementation in the browser. So basically they aren't doing the same thing. On XP, for instance, the PropTypes would not be required in favour of Flow because XP components will never be imported by external consumers.

bernardodiasc:
Fix a bug in the JobsPageLayout responsiveness in https://github.com/x-team/xp-ui/pull/453/commits/83fe1b1bc004a6d21f81361c59b83492cb48f3f6

This is how it looks now:



/cc @nicksp

src/components/ui/ApplicantScreen.js
nicksp: Yeah, good to keep that in mind when setting default values next time. I can see we started preferring ES6 approach over defaultValues 😄

PR #460, opened 2019-10-15
Reorganize storybook

rayepps:
**Release Type:** *Dev Improvements*

Fixes https://x-team-internal.atlassian.net/browse/XP-3408

## Description

The changes effect the way we interact with Storybook. They were done in an effort to make our Storybook more engaging with other people (non-devs specifically). So, in this change I've

* updated all the debug stories to actually be inside a `Debug` folder so they don't pollute the global list.
* grouped the components into sub headings using `|` character
* ordered top level story groups
* organized stories inside groups

## Checklist

## Steps to Test or Reproduce

Run Storybook locally, observe the new file structure in the menu

Storybook

## Screenshots

storybook-screenshot


.storybook/config.js
```diff
@@ -83,9 +83,17 @@ addParameters({
...
 
 // Load stories dynamically
-const req = require.context('../src/components', true, /\.stories\.js$/)
+const reqStyleguides = require.context('../src/components/styleguides', true, /\.stories\.js$/)
+const reqUI = require.context('../src/components/ui', true, /\.stories\.js$/)
+const reqForm = require.context('../src/components/forms', true, /\.stories\.js$/)
+
 function loadStories () {
-  req.keys().forEach(filename => req(filename))
+  // Order matters here - the order components are displayed
```
nicksp: That's a good point and I agree with that. Seems like we're getting several top-level menus that we want to have been displayed in a particular order. Previously it was A-Z type of sort.

src/components/styleguides/Styleguides.stories.js
```diff
@@ -28,7 +28,7 @@ const textOptions = {
   'Paragraphs': 'paragraphs'
 }
 
-storiesOf('Styleguides', module)
+storiesOf('Styleguide|XP', module)
```
bernardodiasc: I wonder if we need this `XP` group here. Because this is the XP-UI, so all the Styleguides are available for all components. I got the idea, but maybe too early for this move.

src/components/ui/GenericTabs/GenericTabs.stories.js
```diff
@@ -13,27 +13,27 @@ const ExampleButton = ({ isActive, text, ...props }) => {
       style={{ padding: '0 10px 10px 0' }}
```
bernardodiasc: I believe we should consider split the changes like this to a different task. It would be better to have this PR focused mainly on navigation rather than how components are displayed. Wdyt?

src/components/ui/Modal.stories.js
```diff
@@ -96,3 +91,10 @@ storiesOf('UI Components/Modal/Use cases', module)
       </StoryModal>
     </Body>
   ))
+
+storiesOf('Core|Modal/Debug', module)
+  .add('missing props (does component explode?)', () => (
+    <Body>
```
bernardodiasc: The `'missing props (does component explode?)'` should not be wrapped by any component, they should be the minimal render only, in this case just `<Modal />`

src/components/styleguides/Styleguides.stories.js
rayepps: This wasn't really by choice. If we want `Styleguides` to show up first we **have** to have something after the `|`. `XP` was the most general thing I could think of. I tried `Styleguide|Default` but that wasn't a good fit logically either.

If you put `storiesOf('Styleguides', module)` then `Styleguides is created as a group - not a category. And no matter what order you import categories are always displayed first.

Anyway, all that to say, in order to meet the AC of showing `Styleguide` first we have to use a format like `Styleguide|{some_text}`

src/components/ui/GenericTabs/GenericTabs.stories.js
rayepps: I think the purpose of the task is to make it easier for non devs to understand whats in storybook and this component in particular made my head spin - so I made it better. I'm a fan of making things better.... seems to get me in trouble here but I'm for this.

src/components/ui/Modal.stories.js
rayepps: 👍 updating

nicksp: Agree on that.

src/components/forms/InputField.stories.js
```diff
@@ -6,7 +6,7 @@ import State from '../../utils/State'
 
 import InputField from './InputField'
 
-storiesOf('Form Components/InputField', module)
+storiesOf('Form Components|/InputField', module)
```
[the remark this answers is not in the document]
rayepps: I missed these before updating but I wrapped them in last commit [aaee80b](https://github.com/x-team/xp-ui/pull/460/commits/aaee80b8477afc07c631d019b3956f2ef26ae2bc)

src/components/forms/InputField.stories.js
```diff
@@ -154,4 +154,6 @@ storiesOf('Form Components/InputField', module)
       placeholder='yyyy-mm-dd'
     />
   ))
+
+storiesOf('Form Components|/InputField/Debug', module)
```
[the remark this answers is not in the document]
rayepps: I missed these before updating but I wrapped them in last commit [aaee80b](https://github.com/x-team/xp-ui/pull/460/commits/aaee80b8477afc07c631d019b3956f2ef26ae2bc)

src/components/ui/Button.stories.js
```diff
@@ -50,102 +50,100 @@ const StoryButton = (props) => (
...
+  .add('openSans', () => (
     <StoryButton contentStyle='openSans'>openSans content style</StoryButton>
   ))
-  .add('sourceSansPro content style', () => (
+  .add('sourceSansPro', () => (
     <StoryButton contentStyle='sourceSansPro'>sourceSansPro content style</StoryButton>
   ))
 
-storiesOf('UI Components/Button/Extra States', module)
+storiesOf('UI Components|Button/States/Outlines', module)
```
bernardodiasc:
```suggestion
storiesOf('Core Components|Button/States/Outlines', module)
```

PR #467, opened 2019-10-30
Create confirmation box UI component

WagnerMoreira:
Fixes https://x-team-internal.atlassian.net/browse/XP-3374

## Description

The purpose of this component is to serve as a generic confirmation box, to be rendered inside a modal. 

ref: https://x-team-internal.atlassian.net/browse/XP-3374

## Checklist

## Steps to Test or Reproduce

This component is not in use yet but it can be visualized on [StoryBook](https://deploy-preview-467--xpui-develop.netlify.com/?path=/story/core-components-confirmationbox-use-cases--with-modal).

New core component created.


src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+    width: 100%
+  `),
+
+  actionButton: cmz(`
+    background-color: ${theme.baseRed}
+    margin: 0
+  `),
+
+  dismissButton: cmz(`
+    color: #8E8D93
```
nicksp: We'll want to reuse existing color as much as possible, or at least the closest ones. We can also always introduce new colors to our `styles/theme` is it's needed and gonna be reused elsewhere.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+    typeface.text, `
+    font-weight: normal
+    font-size: 20px
+    line-height: 140%
+    margin: 24px 0
+  `),
+
+  ctaContainer: cmz(`
+    display: flex
+    justify-content: space-between;
```
nicksp: I suggest we either use `;` or not, so we're consistent with other components.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+    font-size: 14px
+    justify-content: center
+    padding: ${wrapperPadding}
+    width: calc(500px - ${wrapperPadding})
+  `),
+
+  title: cmz(
+    typeface.semiHeading, `
+    color: #000
+    font-size: 24px
```
nicksp: I think we agreed to use `rem`'s for `font-size` at some point in the past to be able to scale the app when needed.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+
+type Props = {
+  title?: string,
+  content?: string,
+  action?: () => void,
+  actionLabel?: string,
+  dismissAction?: () => void
+}
+
+const wrapperPadding = '24px'
```
nicksp: It's intended to be used as a "constant", so let's use `ALL_CAPS` while naming it.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+    flex-direction: column
+    font-size: 14px
+    justify-content: center
+    padding: ${wrapperPadding}
+    width: calc(500px - ${wrapperPadding})
+  `),
+
+  title: cmz(
+    typeface.semiHeading, `
+    color: #000
```
nicksp: We'll want to reuse existing color as much as possible, or at least the closest ones. We can also always introduce new colors to our `styles/theme` is it's needed and gonna be reused elsewhere.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+    {dismissAction && action && (
+      <div className={cx.ctaContainer}>
+        <Button
+          className={cx.actionButton}
+          onClick={action}
+        >
+          {actionLabel}
+        </Button>
+
+        {dismissAction && (
```
nicksp: Is this check really needed if already confirmed both `dismissAction` and `action` exist? Seems like we can omit this check here 🤔

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+    margin-right: auto
+    margin-left: 10px
+  `)
+}
+
+const ConfirmationBox = ({ title = 'Are you sure?', content, action, actionLabel, dismissAction }: Props) => (
+  <div className={cx.wrapper}>
+    <h2 className={cx.title}>{title}</h2>
+    <p className={cx.content}>{content}</p>
+    {dismissAction && action && (
```
nicksp: Instead of this combined check here, I'd check separately for Action and Dismiss button, so in case of at least one missing prop we render one button. Wdyt?

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,97 @@
...
+          {actionLabel}
+        </Button>
+
+        {dismissAction && (
+          <Button
+            pseudolink
+            className={cx.dismissButton}
+            onClick={dismissAction}
+          >
+            DISMISS
```
nicksp: Please use CSS to style the text to be ALL_CAPS. This is important for accessibility and text readers.

src/components/ui/ConfirmationBox.stories.js
```diff
@@ -0,0 +1,92 @@
...
+  .add('random data', () => (
+    <ConfirmationBox
+      title={text('Title', faker.lorem.sentence())}
+      content={text('Content', faker.lorem.paragraph())}
+      action={action('CTA Action: WIDTHDRAW APPLICATION')}
+      actionLabel={text('Action', faker.lorem.sentence())}
+      dismissAction={action('Dismiss action')}
+    />
+  ))
+  .add('missing props (does this component explode?)', () => (
```
nicksp: we started using shorter `missing props` name.

src/components/ui/ConfirmationBox.js
WagnerMoreira: good call!

src/components/ui/ConfirmationBox.js
WagnerMoreira: good call again! haha, done.

src/components/ui/ConfirmationBox.js
WagnerMoreira: my bad on that logic mistake, I didn't notice that, thank you!

WagnerMoreira: and for sure, I've created two more stories for the buttons 👍

src/components/ui/ConfirmationBox.js
WagnerMoreira: done!

src/components/ui/ConfirmationBox.stories.js
WagnerMoreira: done!

src/components/ui/ConfirmationBox.js
WagnerMoreira: good call, changed!

nicksp:
FYI: I shipped all the changes and now waiting for the build to pass on Netlify and Travis CI...

nicksp:
AND IT PASSED 🎉

WagnerMoreira:
> I tested this PR with [x-team/xp-registration#95](https://github.com/x-team/xp-registration/pull/95) and works well. I had some feedback that need attention before approving this PR:
> 
> * It's missing an updated bundle (make sure to run `npm run build`)
> * Move `with Modal and random data` to under Debug group http://localhost:9001/?path=/story/core-components-confirmationbox-use-cases--with-modal-and-random-data
> * The heading of the component should not have fixed `height`, this causes problems when it has more than 1 line of text
> * The class `cx.ctaContainer` probably don't need fixed `height` as well
> 
> Also make sure to check Nick's notes to change request.
> 
> Other than that this is very promising, great stuff @WagnerMoreira
Good call on these changes and thank you!

WagnerMoreira:
> @WagnerMoreira Code looks good to me now, but found some UI related things to improve:
> 
> 1. When `Dismiss` is rendered alone, its left-hand side (while hovering) isn't aligned with text box above. This can be fixed by using some pseudo selectors when the child is the only one so we remove left padding in that case
>    image
> 2. According to designs, we should use `tuna` color for the main content textL
>    image
>    image

Thank you so much for taking the time to review that! I pushed a commit addressing those 2 points

bernardodiasc:
> I'm leaving line-height there to match the designs, I don't think it can cause any issues but please let me know if anything comes to mind.

Yeah, `line-height` is valid to keep it, just need to remove `height` 👍

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+    font-size: 24px
+    font-weight: 500
+    margin: 0
+    width: 100%
+  `),
+
+  content: cmz(
+    typeface.text,
+    `
+    color: ${theme.baseTuna}
```
bernardodiasc: Please use appropriate `typo` color, check instructions on https://deploy-preview-467--xpui-develop.netlify.com/?path=/info/styleguide-general--colorpalette

in this case the correct should be:

```suggestion
    color: ${theme.typoParagraph}
```

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+
+  ctaContainer: cmz(`
+    display: flex
+    justify-content: space-between
+    width: 100%
+  `),
+
+  actionButton: cmz(`
+    background-color: ${theme.baseRed}
+    margin-right: auto
```
bernardodiasc: There's an issue in the layout, the buttons should not be centralised

Screen Shot 2019-11-01 at 23 01 58

If you do the properties removal that I mentioned, **none of the styles of `actionButton` are needed at all**. The `background-color` is already the default of the Button component.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+  `),
+
+  dismissButton: cmz(`
+    & {
+      color: ${theme.baeBrightGray}
```
bernardodiasc: There's a typo here. Also, the `base` colors are not meant to be used by typography properties like `color`. Please take a look at the instructions https://deploy-preview-467--xpui-develop.netlify.com/?path=/info/styleguide-general--colorpalette

Those https://github.com/x-team/xp-ui/blob/develop/src/styles/theme.js#L73-L85 are the typography colors. If the one required to be used here is not available, you have 2 options: check with there is any existing shade that is very similar and check with Pawel if it would work reusing, or add a new typography color to the set.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+    flex-direction: column
+    font-size: 14px
+    justify-content: center
+    padding: ${WRAPPER_PADDING}
+    width: calc(500px - ${WRAPPER_PADDING})
+  `),
+
+  title: cmz(
+    typeface.semiHeading, `
+    color: ${theme.baseDark}
```
bernardodiasc: Please check other similar comments about theme color usage. The `base` colors are intended for different CSS properties, for this case you should use a `typo` color https://github.com/x-team/xp-ui/blob/develop/src/styles/theme.js#L73-L85

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+    font-size: 24px
+    font-weight: 500
+    margin: 0
+    width: 100%
```
bernardodiasc: The `width: 100%` is probably not needed because `h2` is display block by default, so it will fill the entire horizontal space.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+  title?: string,
+  content?: string,
+  action?: () => void,
+  actionLabel?: string,
+  dismissAction?: () => void
+}
+
+const WRAPPER_PADDING = '24px'
+const cx = {
+  wrapper: cmz(`
```
bernardodiasc: I made a test here, it seems that all those properties has no effect in the layout and can be safely removed:

```
    display: flex
    align-items: center
    flex-direction: column
    font-size: 14px
    justify-content: center
```

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+    typeface.text,
+    `
+    color: ${theme.baseTuna}
+    font-weight: normal
+    font-size: 20px
+    line-height: 140%
+    margin: 24px 0
+  `),
+
+  ctaContainer: cmz(`
```
bernardodiasc: I made a test and it seems that all these properties can be removed

```
display: flex
justify-content: space-between
width: 100%
```

after all, the `ctaContainer` class itself is not needed too

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+
+  actionButton: cmz(`
+    background-color: ${theme.baseRed}
+    margin-right: auto
+  `),
+
+  dismissButton: cmz(`
+    & {
+      color: ${theme.baeBrightGray}
+      margin-right: auto
```
bernardodiasc: This `margin-right: auto` is not applying any effect, then can be safely removed.

src/components/ui/ConfirmationBox.js
```diff
@@ -0,0 +1,108 @@
...
+      )}
+
+      {dismissAction && (
+        <Button
+          pseudolink
+          className={cx.dismissButton}
+          onClick={dismissAction}
+        >
+          <span className={cx.dismissButtonLabel}>
+              Dismiss
```
bernardodiasc: Looks like there's an extra indentation here.

WagnerMoreira:
@bernardodiasc thank you so much for reviewing! I just pushed the changes addressing all the comments 👍

PR #470, opened 2019-11-05
Create ClosedJobs component

WagnerMoreira:
Fixes https://x-team-internal.atlassian.net/browse/XP-3377

## Description

Creates the Closed Jobs UI component

## Checklist

## Steps to Test or Reproduce

This component is ready for use but not integrated yet.

My Applications page

## Screenshots

Screen Shot 2019-11-05 at 8 39 25 AM


src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,95 @@
...
+const ClosedJobs = ({ jobs }: Props) => (
+  <Fragment>
+    <h2 className={cx.title}>CLOSED</h2>
+    <ul className={cx.closedJobsContainer}>
+      {
+        jobs && jobs.map(item => (
+          <li key={item.title}>
+            <h3 className={cx.jobTitle}>{item.title}</h3>
+            <p className={description}>
+              Closed on {item.closedDate}, applied on {item.appliedDate}
```
bernardodiasc: I would suggest to use only one string prop here and compose the entire message on XP-Reg side.

src/styles/theme.js
```diff
@@ -28,6 +28,7 @@ export const palette = {
   alto: '#D8D8D8',
   nobel: '#B3B3B3',
   tuna: '#34323B',
+  lightTuna: '#34323BBF',
```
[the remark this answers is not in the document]
WagnerMoreira: actually, now I remember why, if we do ${theme.typoParagraph} for example, it is by default adding `rgb(color)`. so I don't know how to set the alpha to that color on the styles, I take there must be some sort of configuration but I wasn't able to find that on the cmz Github repo. Could you please advise me on how to proceed with that?

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
+// @flow
+/* global React$StatelessFunctionalComponent */
+
+import React, { Fragment } from 'react'
```
nicksp: Let's separate 3rd party imports from local ones with an empty line to keep it consistent with the surrounding files. Separation by logical groups is also welcomed. An example:

```js
import react ...

import LocalComponent from '../LocalComponent'

import style1 from ...

//...and so on
```

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+import typo from '../../styles/typo'
+import theme from '../../styles/theme'
+
+const cmz = require('cmz')
+
+const description = cmz(
```
nicksp: Is there a reason this isn't inside `const cx = {}` declaration?

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+  typo.baseText,
+  `
+    color: #34323B
+    font-style: normal
+    font-weight: 300
+    font-size: 16px
```
nicksp: let's use `rem`'s for font-size

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+import React, { Fragment } from 'react'
+import typo from '../../styles/typo'
+import theme from '../../styles/theme'
+
+const cmz = require('cmz')
+
+const description = cmz(
+  typo.baseText,
+  `
+    color: #34323B
```
nicksp: Could we please not hardcode color?

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    font-style: normal
+    font-weight: 300
+    font-size: 16px
+    line-height: 140%
```
nicksp: line-height is better to be in unitless values: https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+  `),
+
+  title: cmz(
+    typo.sectionHeading,
+    `
+    color: ${theme.typoParagraph.lighten(0.7)}
+    font-style: normal
+    font-weight: bold
+    font-size: 24px
+    line-height: 33px
```
nicksp: same here, to make it unitless

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    list-style: none
+    padding: 0
+  `),
+
+  title: cmz(
+    typo.sectionHeading,
+    `
+    color: ${theme.typoParagraph.lighten(0.7)}
+    font-style: normal
+    font-weight: bold
```
nicksp: let's use numeric values as they're shorter.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    font-size: 24px
```
nicksp: `rem`'s

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    flex-direction: column
+    list-style: none
+    padding: 0
+  `),
+
+  title: cmz(
+    typo.sectionHeading,
+    `
+    color: ${theme.typoParagraph.lighten(0.7)}
+    font-style: normal
```
nicksp: isn't it normal by default?

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    font-size: 24px
+    line-height: 33px
+    letter-spacing: -0.02em
+    text-transform: uppercase
+  `),
+
+  jobTitle: cmz(
+    typo.badgeHeading,
+    `
+    color: #34323B
```
nicksp: do not hardcode colors.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    font-style: italic
+    font-weight: normal
+    font-size: 18px
```
nicksp: `rem`

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    line-height: 25px
```
nicksp: make it unitless

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    text-transform: none
+  `),
+
+  divider: cmz(`
+    height: 1px
+    background: #D6D6D8
```
nicksp: hardcoded color

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+
+  archivedMessage: cmz(
+    description,
+    `
+      margin-top: 24px
+  `)
+}
+
+type Job = {
+  title?: string,
```
nicksp: Is there a reason it's an optional value? Are there any use cases when it can be optional?
As I can see from the [definition](https://github.com/x-team/xp-registration/blob/3ba8f6c7287f5d9ec346f5b51877e9d712dae926/cms/api/job/models/Job.settings.json#L15-L18), Job `name` is always expected to be provided.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+  <Fragment>
+    <h2 className={cx.title}>Closed</h2>
+    <ul className={cx.closedJobsContainer}>
+      {
+        jobs.map(job => (
+          <li key={job.title}>
+            <h3 className={cx.jobTitle}>
+              {AppLink && <AppLink>{job.title}</AppLink>}
+            </h3>
+            <p className={description}>
```
nicksp: could also be `cx.description` is we move declaration inside `const cx = {}`

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+              {job.description}
+            </p>
+          </li>
+        ))
+      }
+    </ul>
+    <div className={cx.divider} />
```
nicksp: Could we use a more semantic `<br />` element here to indicate the divider?

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    <p className={cx.archivedMessage}>Apllications older than 3 months are archived</p>
```
nicksp: typo here
```suggestion
    <p className={cx.archivedMessage}>Applications older than 3 months are archived</p>
```

nicksp: I'd suggest making this text a prop since we're hardcoding business logic here and this component is tightly coupled to the way it's gonna be used. We aren't sure that the logic will always be like `3 months..` before we archive, so it's smth we want to prevent to update should the logic change.

src/components/ui/ClosedJobs.stories.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+]
+
+storiesOf('UI Components|ClosedJobs', module)
+  .add('basic usage', () => (
+    <ClosedJobs jobs={jobsList} link={JobsLink} />
+  ))
+
+storiesOf('UI Components|ClosedJobs/Debug', module)
+  .add('without data', () => (
```
nicksp: remember to also add `missing props` debug story and several more to show the custom text when we make it a prop.

src/components/ui/ClosedJobs.js
bernardodiasc: I think you meant `<hr />`?

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+
+const ClosedJobs = ({ jobs = [], link: AppLink }: Props) => (
+  <Fragment>
+    <h2 className={cx.title}>Closed</h2>
+    <ul className={cx.closedJobsContainer}>
+      {
+        jobs.map(job => (
+          <li key={job.title}>
+            <h3 className={cx.jobTitle}>
+              {AppLink && <AppLink>{job.title}</AppLink>}
```
bernardodiasc: This won't work as expected unless you get the `job.link` too and pass down to this `<AppLink to={job.link}>`

nicksp: So basically, you're missing an address where to go when clicking the link

src/components/ui/ClosedJobs.js
nicksp: Good catch, that's what I meant 😆 Thanks. Just fixed that.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+    letter-spacing: -0.02em
+    text-transform: uppercase
+  `),
+
+  jobTitle: cmz(
+    typo.badgeHeading,
+    `
+    color: #34323B
+    font-style: italic
+    font-weight: normal
```
nicksp: let's use numeric value instead and let's also double check whether we need to override a default value of `normal`. I mean it might be `normal` by default.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,98 @@
...
+import typo from '../../styles/typo'
+import theme from '../../styles/theme'
+
+const cmz = require('cmz')
+
+const description = cmz(
+  typo.baseText,
+  `
+    color: #34323B
+    font-style: normal
```
nicksp: isn't it normal by default or was it overridden so we have to restore that back?

src/components/ui/ClosedJobs.js
WagnerMoreira: as we discussed before, to use rem here I would have to do https://github.com/x-team/xp/issues/2013 first, which I think needs takes time to test the UI. I'm not sure if I would have time to do it this week

nicksp: Hmm... @bernardodiasc aren't you using `rem`'s already on components that are meant to be used on `xp-reg`. Any concerns in regards to above?

src/components/ui/ClosedJobs.js
nicksp: And it's good to add stories that will show how the component will look when we pass a correct prop value to it.

src/components/ui/ClosedJobs.js
WagnerMoreira: I am doing `const description =` outside of `cx` context to be able to reuse code
so I can reuse like this https://github.com/x-team/xp-ui/pull/470/files?file-filters%5B%5D=.js#diff-b3935dc602ea8db09af6675984495f3aR59

```
  archivedMessage: cmz(
    description,
    `
      margin-top: 24px
  `)
```

src/components/ui/ClosedJobs.js
WagnerMoreira: please check the answer here https://github.com/x-team/xp-ui/pull/470#discussion_r343169917

src/components/ui/ClosedJobs.js
WagnerMoreira: good call thanks for the suggestion!

src/components/ui/ClosedJobs.js
nicksp: Makes sense. Is it possible to reference it like that? have you tried?

```js
archivedMessage: cmz(
    cx.description,
```

src/components/ui/ClosedJobs.js
WagnerMoreira: ok, in this case, I had to set it to `normal` because I extended `typo.sectionHeading` which is `font-weight: 600`, changed to `400`.

nicksp: ALl good here. Just double check if any of props is needed :)

src/components/ui/ClosedJobs.js
WagnerMoreira: this one is not needed, removed!  thank you

src/components/ui/ClosedJobs.js
WagnerMoreira: I tried but it throws an error `error 'cx' was used before it was defined`, I would have to use something like `this.descritpion` which also doesn't work.

src/components/ui/ClosedJobs.js
WagnerMoreira: this one is, removed.

src/components/ui/ClosedJobs.js
WagnerMoreira: This is because of debug stories, tests fail if something is required and I don't pass.
Would you know of another way of handling that?

nicksp: That doesn't mean we should set *every* param as optional. That is wrong. The logic should rely on storybook. Only options that **should be** optional, should be optional u know what I'm saying. Please check. Not every component we have have all optional fields. One of the ways to handle. Some props just have to be required in order to operate properly. And only props that are *optional* should be optional.

src/components/ui/ClosedJobs.js
WagnerMoreira: I thought it would come from the consumer, based on `JobsPageBreadcrumbs` and `JobsCard` which also has links but the component does not have a `to={}`, so using `to={}` in the component definition is a best practice?

src/components/ui/ClosedJobs.js
nicksp: @WagnerMoreira Please, don't get me wrong though. I'm not requiring to make this prop required. The idea I'm trying to convey is that sometimes some props *have to be* required in case it's crucial to be provided when the component doesn't have any point to be rendered if that prop isn't passed. Sometimes, it might be a good idea to provide some sane default where appropriate. So each case is unique and should be moderated sticking to common sense. Probably it makes sense to make most of the props optional, provide sane defaults where appropriate, and only in some rare cases make a prop required and expected it's passed. It would be nice to see how 3rd party UI components libs handle that case. We might do some quick research. But Storybook itself shouldn't be a driving factor to make *every* prop optional. It should follow a common sense 😄 
I hope my point is clear.

src/components/ui/ClosedJobs.js
nicksp: Don't add styles just because it's safe to do that. Tend to only add styles where it's really needed.

src/components/ui/ClosedJobs.js
bernardodiasc: I used, yes, but the font-size system is very weak at the moment and using relative units right now will not produce the benefits we desire. I would not push this approach without having a solid foundation for it, which I believe we can achieve in https://x-team-internal.atlassian.net/browse/XP-3436

nicksp: Sounds good 👍

src/components/ui/ClosedJobs.js
bernardodiasc: That's the case @WagnerMoreira, but in here each Job has a different Link. So you can either pass a different `link` prop on each job item, or pass only the url for each job and one global `link`.

src/components/ui/ClosedJobs.js
bernardodiasc: @WagnerMoreira, the

```
import typo from '../../styles/typo'

import theme from '../../styles/theme'
```

are meant to stay together, without extra line. Because these are same kind (local lib).

src/components/ui/ClosedJobs.js
bernardodiasc: We fall into this discussion quite frequently. For a matter of fact, it's a good thing to have almost every prop as optional in UI components. This is the only way we can ensure the UI render is bullet-proof, because relying on a Type checking system for a lib that's consumed on different repos doesn't enforce requirements. The mindset for UI components are not the same of regular components. Anyway, in this particular case it probably doesn't matter if it's optional or required, as long `jobs?: Array<Job>,` is optional.

src/components/ui/ClosedJobs.stories.js
```diff
@@ -0,0 +1,68 @@
...
+  .add('with jobs only', () => (
+    <ClosedJobs jobs={jobsList} />
+  ))
+  .add('with archived message only', () => (
+    <ClosedJobs arquivedMessage={'Applications older than 3 months are archived'} />
+  ))
+  .add('with link only', () => (
+    <ClosedJobs link={JobsLink} />
+  ))
+  .add('missing props (does component explode?)', () => (
```
bernardodiasc: this story's name was simplified to:

```suggestion
  .add('missing props', () => (
```

src/components/ui/ClosedJobs.js
nicksp: Well said here, thanks Bernardo.

src/components/ui/ClosedJobs.js
nicksp: Well, sometimes we don't need to ensure that. Let's say we have documentation that tells that some props are required. So you didn't pass anything, it's your fault 😛 At the same time, as developers, we should provide ways to remind consumers that smth didn't go well. I gave it another thought given the nature of the project. We can use `throw` from inside components as if we were doing runtime props validation. Here's an example:

```js
if (typeof requiredProp === 'undefined) {
    throw new Error('Required "requiredProp" isn't provided. Make sure you blah-blah...')
}
```

Most UI libs like React does same thing to ensure the lifecycle methods are used correctly by a developer. A nice benefit, that we can finally mark components according to their **real** purpose and intent: have both optional props (with sane defaults) and required props that are vital for the component to behave. That will change the way we treat are components and make their usage more predictable and logical.

We can give it a go at least if that sounds good. I would like to hope that can help us solve the endless "discussions" we've been having so far 😆 Let's at least discuss.

cc @bernardodiasc @WagnerMoreira

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,100 @@
+// @flow
+/* global React$StatelessFunctionalComponent */
+
+import React, { Fragment } from 'react'
+
+import typo from '../../styles/typo'
+
```
nicksp: It's preferred we separate files by logical group only 😄

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,100 @@
...
+import theme from '../../styles/theme'
+
+const cmz = require('cmz')
+
+const description = cmz(
+  typo.baseText,
+  `
+    color: ${theme.typoParagraph}
+    font-weight: 300
+    font-size: 1rem
```
nicksp: Let's revert back, sorry. As discussed lately, it's best we handle a blocking ticket first before we can use `rem`'s for UI components intended to be used from inside `xp-reg` repo.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,100 @@
...
+    list-style: none
+    padding: 0
+  `),
+
+  title: cmz(
+    typo.sectionHeading,
+    `
+    color: ${theme.typoParagraph.lighten(0.7)}
+    font-weight: 700
+    font-size: 1.5rem
```
nicksp: same here, let's revert back to `px`.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,100 @@
...
+    text-transform: uppercase
+  `),
+
+  jobTitle: cmz(
+    typo.badgeHeading,
+    `
+    color: ${theme.typoParagraph}
+    font-style: italic
+    font-weight: 400
+    font-size: 1.125rem
```
nicksp: `px`

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,100 @@
...
+type Job = {
+  title?: string,
+  description?: string,
+  link?: string
+}
+
+type Props = {
+  jobs?: Array<Job>,
+  link?: React$StatelessFunctionalComponent<*>,
+  arquivedMessage?: string
```
nicksp: First off, there's a typo, but secondly: don't you think we're better just name it in a generic way `message`? What if business logic changes and we instead render a message for smth different?! In this case, we'd need to update the UI component. But we can prevent that if we try to keep prop names generic.

src/components/ui/ClosedJobs.js
nicksp: Got it 👍 Let's keep as is.

src/components/ui/ClosedJobs.js
bernardodiasc: Isn't that the purpose of the React PropTypes we agreed on using it?

bernardodiasc: By the way, it's great and important to discuss the conventions. 👍

nicksp: `propTypes` is only checked in development mode, and has no effect in Prod. But the problem with our UI component that we don't know who is gonna use them.

src/components/ui/ClosedJobs.js
nicksp: cc @WagnerMoreira

src/components/ui/ClosedJobs.js
WagnerMoreira: Interesting discussion, I understand the point of having all props optional because as of now Flow is not enforcing required props in the consumer app. I also understand it is important to make optional only what is really optional otherwise things could fail and we would spend more time than usual to understand why in certain cases. The perfect solution would be to use `Typescript` (or make Flow do that trick) as it exports typings for everything you do, so the properties would still be required when you are using them, and Typescript would yell at you in the consumer app saying "prop is required".
Maybe Flow could do the same, I am researching if we can achieve these results with Flow.
If you guys already know if the answer is no and this is a limitation of the tech please let me know. 👍 
This seems like a discovery ticket, what you guys think? cc: @nicksp  @bernardodiasc

src/components/ui/ClosedJobs.js
WagnerMoreira: @bernardodiasc @nicksp  got it! and changed

src/components/ui/ClosedJobs.js
WagnerMoreira: done, reverted all font-sizes to px.

src/components/ui/ClosedJobs.js
WagnerMoreira: good call @nicksp  thanks for pointing that!

WagnerMoreira:
@nicksp @bernardodiasc I addressed all recent review comments and pushed 👍

src/components/ui/ClosedJobs.js
nicksp: @bernardodiasc @WagnerMoreira Here's the Discovery to look into deeply: https://x-team-internal.atlassian.net/browse/XP-3451

WagnerMoreira:
> @WagnerMoreira is there a reason you introduced changes to `package-lock.json` in this commit [1dd265e](https://github.com/x-team/xp-ui/commit/1dd265e67d6dfc4a728fb2812fe518af0175b6bf)?

that was my mistake, reverted in my last push

src/components/ui/ClosedJobs.stories.js
```diff
@@ -0,0 +1,68 @@
+// @flow
+/* global React$Node */
+
+import React from 'react'
+import { storiesOf } from '@storybook/react'
```
nicksp: we should also separate external deps from local ones to distinguish visually. Gonna fix as part of my review.

src/components/ui/ClosedJobs.js
```diff
@@ -0,0 +1,99 @@
...
+const cmz = require('cmz')
+
+const description = cmz(
+  typo.baseText,
+  `
+    color: ${theme.typoParagraph}
+    font-weight: 300
+    font-size: 16px
+    line-height: 1.4
+    margin: 4px 0 24px 0
```
nicksp: Instead of manipulating spacing inside container components (`li`), it is more flexible and reliable to manipulate `li` element itself. Just image there' no inside comment, thus margin won't apply and UI looks will be messed. `li` is ways there, so it's better to apply outer spacing to that component.

PR #475, opened 2019-11-11
Remove Links from ClosedJobs

bernardodiasc:
**Release Type:** *Breaking changes*

Fixes https://x-team-internal.atlassian.net/browse/XP-3454

## Description

- Remove `Links` from `ClosedJobs`

Extra:

- Rename `ClosedJobs` to `ClosedJobApplications` and changed terminologies from `job` to `application` given the clarifications in the sprint planning (this is a list of closed applications, not closed jobs)
- Several adjustments in the UI component and on its stories
- Improve `JobsPageLayout` stories by providing more complete showcases
- Standardised `basic usage` story name everywhere
- Include separation margin on `JobsGrid` when the component is not the only child

## Checklist

### `ClosedJobApplications`

http://localhost:9001/?path=/story/ui-components-closedjobapplications--basic-usage

Also check other debug stories for this component.

Open question: we seem to no handle the case where none closed application is provided, it's possible that a message to the user can be useful on this case (or maybe not, just something to ask about!). See http://localhost:9001/?path=/story/ui-components-closedjobapplications-debug--without-applications

### `JobsPageLayout` 

http://localhost:9001/?path=/story/screens-and-layouts-jobspagelayout-use-cases--my-applications-page

The "Use Cases" stories was provided and some others got updated given latest component additions. In particular this "My Applications page" story helps to see how the spacing between components looks like (`JobsGrid` and `ClosedJobApplications` now have spacing in between without hacks).

Mostly the previously called `ClosedJobs`.

## Screenshots

_The existing presentation should have no changes._


src/components/ui/JobsGrid.stories.js
```diff
@@ -9,7 +9,7 @@ import JobsGrid from './JobsGrid'
 import JobCard from './JobCard'
 import { JobCardLink } from './JobCard.stories'
 
-export const jobCards = Array(15)
+export const jobCards = (amount: number = 15) => Array(amount)
```
nicksp: Are you sure this represents what you think it represents? The current way will create an array consisting of `15` undefined values. Not sure it has any purpose in our case. If you want to create a proper array, I suggest doing smth like that:

```js
[...Array(15).keys()] // > [0, 1, 2, 3, 4, 5, ..., 14]
```

src/components/ui/JobDetails.stories.js
```diff
@@ -7,9 +7,9 @@ import faker from 'faker'
 
 import JobDetails from './JobDetails'
 
-const nameSample = faker.name.jobTitle()
-const skillsSample = faker.random.words(faker.random.number(10)).split(' ').join(',')
-const descriptionSample = `${faker.lorem.paragraph()}
+export const jobDetailsName = faker.name.jobTitle()
+export const jobDetailsSkills = faker.random.words(faker.random.number(10)).split(' ').join(',')
+export const jobDetailsDescription = `${faker.lorem.paragraph()}
```
nicksp: What is the purpose to export it? Isn't it used **just** internally by this story?

src/components/ui/ClosedJobApplications.js
```diff
@@ -0,0 +1,95 @@
...
+  message?: string
+}
+
+const ClosedJobApplications = ({ applications = [], message }: Props) => (
+  <div className={cx.wrapper}>
+    <h2 className={cx.heading}>Closed</h2>
+    <ul className={cx.list}>
+      {applications.map(application => (
+        <li key={application.title} className={cx.item}>
+          <h3 className={cx.title}>
```
nicksp: How do we accomplish this: `Do not show any entry if there is no job title.`?

src/components/ui/JobDetails.stories.js
bernardodiasc: This is re-used on other stories, in this case for demonstrating Use Cases compositions on `JobsPageLayout.stories.js`

nicksp: in this case, i'd suggest to move that to a shared file so that different stories don't depend on each other and are self-defined.

src/components/ui/JobsGrid.stories.js
bernardodiasc: Alright, `Array(15).fill('')` or `[...Array(15).keys()]`, the outcome will be the same, I don't mind changing it. It doesn't really matter if the initial value is undefined or the key, because the composition happens in the `map` after creating the array with the desired amount.

bernardodiasc: Out of curiosity :)
Screen Shot 2019-11-12 at 10 04 51

nicksp: I don't trust benchmarks :) But anyways, leave it as it is then. I don't mind either.

nicksp: Anyways, it's a story, not a real code, so don't matter much

src/components/ui/ClosedJobApplications.js
bernardodiasc: I added a new story to showcase "broken data" http://localhost:9001/?path=/story/ui-components-closedjobapplications-debug--broken-data

nicksp: so this AC was missing initially, right?

src/components/ui/JobDetails.stories.js
bernardodiasc: I believe this is the best place for this things to be. It's not something global, this file is the origin of the data, the stories are dependent on each other as much as their respective components are dependent of each other.

In the other story the import looks like:

```
import JobDetails from './JobDetails'
import { jobDetailsName, jobDetailsSkills, jobDetailsDescription } from './JobDetails.stories'
```

It's straight forward and in case of deprecation of components the connected parts are easy to reason.

src/components/ui/ClosedJobApplications.js
bernardodiasc: yeah 👍

PR #478, opened 2019-11-13
Create GenericTooltip

bernardodiasc:
**Release Type:** *non-breaking changes*

Fixes https://x-team-internal.atlassian.net/browse/XP-3456

### Acceptance Criteria

1. Create a generic Tooltip component (with minimal API requirements at the moment)

> Done! The `GenericTooltip` is an interface of https://github.com/wwayne/react-tooltip. I pick this lib because it's very similar to our needs and very flexible. Styling is not really good and current stage have very little support for different cases beyond defaults, but importing this lib leveraged a huge amount of work.

2. Create showcases in Storybook

> Done!

3. Implement the Tooltip in the HeaderBar component, in the "Update your profile" link (see https://xpui-develop.netlify.com/?path=/story/ui-components-headerbar-use-cases--with-links)

> Done!

4. Consider moving the business information to the consumer app (XP-Reg), but be aware of HeaderBar implementation under the scope of  https://x-team-internal.atlassian.net/browse/XP-3310

> During discussions with @WagnerMoreira, we realised that the HeaderBar menu links required an amount of work that's specific and could be scoped into own task, therefore I suggested to raise a new dev task focused on it: https://x-team-internal.atlassian.net/browse/XP-3459, I think we can consider this AC to be addressed there, in the scope of this PR I just used what already has

5. Make sure that the new generic Tooltip is also compatible and can be re-used in the new WelcomeHero UI component that's going to be created in the scope of  https://x-team-internal.atlassian.net/browse/XP-3367

> It's compatible because it's a generic and flexible Tooltip 👍 

## Checklist

## Steps to Test or Reproduce

1. Make sure to run `npm i`
1. Check new `GenericTooltip` component at http://localhost:9001/?path=/story/core-components-generictooltip--basic-usage
1. Check implementation in `HeaderBar` at http://localhost:9001/?path=/story/ui-components-headerbar-use-cases--with-links

New `GenericTooltip` added. Also extended `HeaderBar`.


src/components/ui/ApplicantScreen.js
```diff
@@ -73,7 +73,9 @@ const ApplicantScreen = ({
 
 ApplicantScreen.propTypes = {
   children: PropTypes.node,
-  contentWrapper: PropTypes.bool
+  noWrapper: PropTypes.bool,
+  wrapper: PropTypes.string,
```
nicksp: Having both `noWrapper` and `wrapper` is confusing as they kinda overlap with each other. It would be nicer if we had just one prop `wrapper` which if passed would mean the wrapper is needed and if it's omitted then it's clear wrapper isn't required. Simple and easy to reason about. If you meant different wrappers, then please update names to reflect their actual meaning. But if it's possible to live with less props - even nicer 🚀

src/components/ui/HeaderBar.js
```diff
@@ -175,7 +177,11 @@ class HeaderBar extends PureComponent<Props> {
               <ul className={cx.menu}>
                 {this.renderLinks()}
               </ul>
-              <a className={cx.profile} href='https://www.linkedin.com/in/' target='_blank'>Update your profile</a>
+              <GenericTooltip
+                message={<span>We currently heavily rely on LinkedIn<br />Profiles to match candiates with<br />potential roles.</span>}
```
nicksp: typo
```suggestion
                message={<span>We currently heavily rely on LinkedIn<br />Profiles to match candidates with<br />potential roles.</span>}
```

src/components/ui/HeaderBar.js
```diff
@@ -175,7 +177,11 @@ class HeaderBar extends PureComponent<Props> {
...
+              >
+                <a className={cx.profile} href='https://www.linkedin.com/in/' target='_blank'>Update your profile</a>
```
nicksp: When using HTML anchor elements, remember to also specify `rel="noopener noreferrer"` attr to prevent possible vulnarabilities.

src/components/ui/ApplicantScreen.js
bernardodiasc: We have a ticket for that already 😉

nicksp: Which one? Why don't do it right here since you're introducing this extra prop in current changes and not change it later? 🤔 It doesn't seem to be smth we had before. I'd prefer not introducing confusing prop until there's a strong reason to have it.

nicksp:
@bernardodiasc Let's get merge conflicts resolved 💪

bernardodiasc:
@nicksp  this PR got rebase, feel free to give it another try 👍

src/components/ui/ApplicantScreen.js
bernardodiasc: As discussed in DM, nothing was introduced in this component, just fixed the proptypes definitions to match what the component already have. There a ticket with a long description about how we want to refactor this component https://x-team-internal.atlassian.net/browse/XP-3430. This is indeed not so good, but the change requirements are bigger than it looks like and quire unrelated to this ticket. This fix was something I noticed while checking the tooltip implementation under the ApplicantScreen + HeaderBar composition... If you think it should not be here I can remove it.

nicksp: Yep, got it. sorry might thave missed that. XP-3430 is a great place to get things unified and improved 👍

PR #479, opened 2019-11-14
Create JobPageProTipCard

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3368

## Description

Create `JobPageProTipCard` (I pick this name because it seems something for "JobPages" only and to keep related components next to each other in the navigation... but I don't have strong opinion about whether it should be just `ProTipCard`).

ALso, I added some missing components at the `index.js`. As far as I got the index.js should export all available components, right? Please let me know if that's not correct then I'll revert changes.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-3368-create-prop-tip-card` | [XP Registration](https://github.com/x-team/xp-registration/pull/109)

## Steps to Test or Reproduce

1. Check the component at http://localhost:9001/?path=/story/ui-components-jobspageprotipcard--basic-usage
1. Check JobsPageLayout compositions with the new component included http://localhost:9001/?path=/story/screens-and-layouts-jobspagelayout--basic-usage

New component added `JobPageProTipCard`

## Screenshots

Screen Shot 2019-11-14 at 11 46 32


src/components/ui/JobsPageLayout.stories.js
```diff
@@ -49,7 +50,13 @@ storiesOf('Screens and Layouts|JobsPageLayout', module)
           hero={<SampleHero />}
           heading='Jobs'
           content={<JobsGrid jobCards={jobCards(5)} />}
-          sidebar={<SampleSidebar />}
+          sidebar={
+            <JobsPageProTipCard heading='How to Stand Out'>
+              <span>
+                To be selected for an interview among thousands of applicants, <a href='#'>ensure your LinkedIn profile is up-to-date</a>. It should clearly show your years of experience working on large scale projects relevant to the job you’re applying for. Remember to showcase the impact of your role on each team!
```
nicksp: LinedIn link should point to a linkedin where applicant can logic to his profile.
Also, remember about security for regular links: `rel='noreferrer noopener'`

I know it's a story only, but we should provide accurate examples that can be copy-pasted, so it might worth doing that 😄 But, we might also consider skipping this recommendation. Up to you 😛 It's just my personal thought that Storybook stories should be as close to real usage as possible. Otherwise, developers might say: "I took it from the stories so assumed it was written the recommended way" 😆

src/components/ui/JobsPageProTipCard.js
```diff
@@ -0,0 +1,65 @@
...
+const cmz = require('cmz')
+
+const cx = {
+  card: cmz(
+    typo.regularText,
+    `
+      & {
+        background: ${theme.baseYellow}
+        border: 1px solid ${theme.lineYellow}
+        box-shadow: 4px 4px 0px ${theme.baseYellowDark.fade(0.85)}
```
nicksp:
```suggestion
        box-shadow: 4px 4px 0 ${theme.baseYellowDark.fade(0.85)}
```

src/components/ui/JobsPageProTipCard.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+  `)
+}
+
+type Props = {
+  children?: React$Node,
+  heading?: string
+}
+
+const JobsPageProTipCard = ({ children, heading }: Props) => children || heading ? (
```
nicksp: I think it should be safe to exclude `children` from the condition check, since if it' falsy it won't be rendered anyways. I can check that yourself.
```suggestion
const JobsPageProTipCard = ({ children, heading }: Props) => heading ? (
```

src/components/ui/JobsPageProTipCard.js
```diff
@@ -0,0 +1,65 @@
...
+  <div className={cx.card}>
+    {heading && (
```
nicksp: If my prev comment is applied this check isn't needed at all so we can shorten the component rendering code.
```suggestion
    <h2 className={cx.heading}>{heading}</h2>
```

src/components/ui/JobsPageProTipCard.js
nicksp: Since we pass both the title and message content, this component is a good candidate to be used anywhere else on the app. But its name implicitly restricts its usage and binds it to JobsPage only. This can easily be included in other pages if needed. So I recommend renaming that to simply: `ProTipCard`. Nice and clean name 👍

bernardodiasc: I'm not sure about this change. Because there's no point in showing an empty box:

Screen Shot 2019-11-15 at 11 31 40

src/components/ui/JobsPageProTipCard.js
bernardodiasc: I think this will not behave as expected. If I made this change the component will not be rendered if `heading` is not available.

This is what is expected with current code:

Screen Shot 2019-11-15 at 11 33 13

src/components/ui/JobsPageProTipCard.js
nicksp: Got it, yeah let's leave as is. BTW, since it's a functional component we can omit `: null` part and just render like that: 

```js
const Component = () => condition && (<div></div>)
```

bernardodiasc: In the past I thought that would work, but it doesn't, here's the output:

Screen Shot 2019-11-15 at 11 53 17

nicksp: Ah OK. It works in CodeSandbox: https://codesandbox.io/s/romantic-liskov-84vms

src/components/ui/ProTipCard.js
```diff
@@ -0,0 +1,65 @@
...
+        line-height: 1.4
+        font-weight: 400
+      }
+
+      & a {
+        color: ${theme.typoAnchor}
+      }
+
+      & a:hover {
+        color: ${theme.typoAnchorHover}
```
nicksp: Just a note: `typoAnchorHover` is very slightly different from `typoAnchor` ;-) It's very hard to see the difference 😆

src/components/ui/JobsPageLayout.stories.js
```diff
@@ -41,6 +41,24 @@ const SampleSidebar = () => (
   <div style={{ border: '2px dashed green' }}>Sample Sidebar component</div>
 )
 
+const JobsPageProTipCard = () => (
```
nicksp: The code is shorter and easier to maintain 💪 Thanks for the improvement.

PR #477, opened 2019-11-13
Change ClosedJobsApplication component prop names

WagnerMoreira:
## Description

This change was necessary to match real data property names. On `xp-registration` we have `name`
 containing job title, and `description` is for job description and not closed and applied job application dates so I changed it to `info`.

## Checklist

- [x] set yourself as an assignee
- [x] set appropriate labels for a PR (`In Review` or `In Progress` depending on its status)
- [x] make sure your code lints (`npm run lint`)
- [x] Flow typechecks passed (`npm run flow`)
- [x] Snapshots tests passed (`npm run jest`)
- [x] if any snapshots have been changed, verify that component still works and looks as expected and update the changed snapshot
- [x] once finished with all the changes, you run `npm run build` command and committed changed files to the branch
- [x] **manually tested the app** by running it in several different browsers (Firefox, Chrome, Opera, Safari, MS IE/Edge, etc.) and checked nothing is broken and operates as expected!

## Related PRs

branch | PR
------ | ------
xp_branch_name | [XP](https://github.com/x-team/xp/pull/X)
xp_registration_branch_name | [XP Registration](https://github.com/x-team/xp-registration/pull/X)
xp_api_branch_name | [XP API](https://github.com/x-team/xp-api/pull/X)

## Steps to Test or Reproduce

Check ClosedJobsApplication component on storybook or in the upcoming `My Applications` page.

## Impacted Areas in Application

the upcoming My Applications page

PR #481, opened 2019-11-14
Create Tagslist Component to be used on JobCards and JobDetailsPage

WagnerMoreira:
Fixes https://x-team-internal.atlassian.net/browse/XP-3366

## Description

Creates TagsList component, which is being implement into JobCard and JobDetails Page, replacing code that was rendering the tags in the latter.

## Checklist

## Related PRs

branch | PR
------ | ------

xp_registration_branch_name | [XP Registration](https://github.com/x-team/xp-registration/pull/X)

## Steps to Test or Reproduce

Check My Applications page once this PR and the PR open on xp-registration are merged

My Applications Page, Job Details page

## Screenshots

Screen Shot 2019-11-14 at 10 34 07 AM


src/components/ui/TagsList.js
```diff
@@ -0,0 +1,52 @@
...
+    `
+  )
+}
+
+type Props = {
+  skills?: string,
+}
+
+const renderTags = (skills: string) => skills.split(',').filter(Boolean).map((skill, i) => (
+  <div key={`jobskill-${i}`} className={cx.skill}>{skill}</div>
```
nicksp: If skills names won't be repeated (which seems like a very likely case, but not sure actually), it would be more preferred from a perf perspective to use skills name as a `key` prop.

src/components/ui/TagsList.js
```diff
@@ -0,0 +1,52 @@
...
+))
+
+const TagsList = ({ skills = '' }: Props) => skills ? (
```
nicksp: BTW, when you're inside functional component, it's totally legit to simplify conditional rendering so the final code reads like that:

```js
const Component = () => skill && (<div>...</div>)
```

src/components/ui/TagsList.stories.js
```diff
@@ -0,0 +1,17 @@
...
+import faker from 'faker'
+
+import TagsList from './TagsList'
+
+const tagsList = faker.random.words(faker.random.number(10)).split(' ').join(',')
+
+storiesOf('Core Components|TagsList', module)
+  .add('basic usage', () => (
+    <TagsList
+      skills={text('Skills', tagsList)}
```
nicksp: Should we also add our "famous" `no props` story 😏 ?

src/components/ui/TagsList.stories.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { text } from '@storybook/addon-knobs'
+import faker from 'faker'
+
+import TagsList from './TagsList'
+
+const tagsList = faker.random.words(faker.random.number(10)).split(' ').join(',')
```
nicksp: For testing purposes, I'd also recommend specifying skills as a list that is separated by different number of spaces as it's smth that can happen in Strapi as we don't force any specific format for specifying skills, so that:

```js
skills = 'skills 1, skill2,   skill3,  skill4'
```

it can be possible if we render `,` or `, ` based on the faker.boolean check.

src/components/ui/ClosedJobApplications.js
```diff
@@ -62,8 +62,8 @@ const cx = {
 }
 
 type JobApplication = {
-  title?: string,
-  description?: string,
+  name?: string,
```
nicksp: out of curiosity, where did we decide to change these props? So we're on the same page on this decision.

bernardodiasc:
hey @WagnerMoreira, can we not use `TagsList` name? :) We had similar name for a component in the past, using "tags", it doesn't ended very well, it's misleading. This is not really generic, it's reusable indeed but very specific for Jobs Skills, I would suggest to use `JobSkills` name or something like that.

src/components/ui/ClosedJobApplications.js
WagnerMoreira: This is because there are no title and description props on a job application object. The object has `name`, `closeDate` and `ApplicationDate`, the last two compose the `info` prop that was `description`. When I implemented `ClosedJobApplications` I noticed that the props didn't match with the real data.

WagnerMoreira: ref: https://github.com/x-team/xp-ui/pull/477

src/components/ui/JobCard.stories.js
```diff
@@ -33,27 +35,44 @@ storiesOf('UI Components|JobCard/Debug', module)
   .add('without message', () => (
     <JobCard
       name={text('Name', nameSample)}
+      hasMarginLeft
```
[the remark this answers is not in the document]
WagnerMoreira: Initially, I thought it would make sense to showcase that, but I changed it so all stories have the prop now.

src/components/ui/JobCard.stories.js
```diff
@@ -33,27 +35,44 @@ storiesOf('UI Components|JobCard/Debug', module)
...
   .add('without summary', () => (
     <JobCard
       name={text('Name', nameSample)}
+      hasMarginLeft
+      skills={skills}
       link={JobCardLink}
       message={text('Message', messageSample)}
     />
   ))
+  .add('without hasMarginLeft', () => (
```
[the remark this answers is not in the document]
WagnerMoreira: removed

src/components/ui/TagsList.stories.js
WagnerMoreira: I understand the idea but I don't understand what you mean by using faker.boolean, can you post an example.

PR #482, opened 2019-11-19
Add new WelcomeHero component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3367

## Description

- Added new WelcomeHero component that reuses Text and VideoPlayer
- Upgraded VideoPlayer component to allow overlay cover
- Upgraded Text component to accept style classes for heading and content

## Checklist

### WelcomeHero

1. Go to http://localhost:9001/?path=/story/ui-components-welcomehero--basic-usage
1. Observe how the component behaves
1. Check Debug group for more bullet-proofing presentation

### VideoPlayer

1. Go to http://localhost:9001/?path=/story/core-components-videoplayer-use-cases--embedded-video-with-clickable-overlay
1. Click to play and observe that the Youtube iframe is loaded and played automatically (if the embedded iframe video doesn't play automatically, make sure to disable adblocker for the page - https://youtubevideoembed.com/youtube-video-not-autoplay/ - it should work)

- Added new WelcomeHero component
- Upgraded VideoPlayer component
- Upgraded Text component

## Screenshots

> Note, for the full responsive embedded iframe I would like to suggest a dedicated task, it has some tricks to accomplish that and would be good to make on a separated scope.


src/components/ui/VideoPlayer.js
```diff
@@ -35,7 +69,12 @@ class VideoPlayer extends PureComponent<Props> {
     loop: false,
     muted: false,
     poster: '',
-    preload: 'auto'
+    preload: 'auto',
+    overlay: false
+  }
+
+  state = {
+    displayOverlay: !!this.props.overlay
```
nicksp: A component should not sync its state with its props, the 2 must be separate. This is an anti-pattern in React world. Using props to initialize state in `state = {}` often leads to duplication of “source of truth”, i.e. where the real data is. This is because `state = {}` is only invoked when the component is first created, so any potential `overlay` props change aren't synched as expected which leads to bugs and confusion.

In other words, the danger of doing this is that if the props on the component are changed without the component being ‘refreshed’, the new prop value will never be displayed because the constructor function (or `state = {}`) will never update the current state of the component. The initialization of state from props only runs when the component is first created. That's not what you expect to have I assume, so in this case, the recommendation is to make it clear that the `overlay` prop is only seed data for the component’s internally-controlled `displayOverlay ` state. That is, if your component doesn’t need to stay aware of potential changes in the prop value it’s fine to initialize state this way. And in this case, to make the intent even clearer, I like to name such "one-time" seed props as `initialSmth`. In your case it could be `initialOverlay`. And this means, you no longer use `this.props.initialOverlay` directly from the component. It'll now be used by referencing the state so we don't introduce 2 source of truth.

For the future: If you’re thinking about whether or not to seed state data into a React component via props, ask yourself these:
1. What’s the lifecycle for this component? Will it hang around very long without being refreshed?
1. Can the values that come in through **props** be changed elsewhere? and if they are, should this component display the updated value?
1. If I mutate these seed values in **state**, will they be needed to re-seed this component on a ‘save’ event (or similar)?

Hope that was helpful and clear 😛

src/components/ui/VideoPlayer.js
```diff
@@ -65,6 +105,54 @@ class VideoPlayer extends PureComponent<Props> {
...
+    const regExp = /(?:youtube\.com\/(?:[^/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)([^"&?/ ]{11})/i
+    const match = src.match(regExp)
+
+    if (!match) {
+      return null
+    }
+
+    const videoID = match[1]
+
+    return displayOverlay ? (
```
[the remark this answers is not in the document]
bernardodiasc: Technically the overlay is an option of embedded...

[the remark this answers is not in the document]
nicksp: Then it should be like that: having both overlay and embedded.

bernardodiasc:
Hey @nicksp, I made few updates based on your feedback. I don't have much more ideas on things to change, unless you have any explicit recommendation to change the logic. The way I see it makes sense, but of course I can be wrong. On other hand, it looks like some of the functionality is not entirely clear for you, so please let me know any further question.

I also added some more debug stories for the VideoPlayer. So, well, I'm moving this back to Review, feel free to move it back to Progress is not satisfactory yet and let me know what else need to be updated. Thanks for the feedbacks!

nicksp:
**FYI:** Reabsed with `develop` to get rid of merge conflicts/

src/components/ui/VideoPlayer.js
```diff
@@ -65,6 +107,49 @@ class VideoPlayer extends PureComponent<Props> {
...
+      isOverlayHidden: true
+    })
+  }
+
+  renderEmbeddedVideo = () => {
+    const { src, width, height, overlay } = this.props
+    const { isOverlayHidden } = this.state
+
+    const youtubeUrlRegExp = /(?:youtube\.com\/(?:[^/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)([^"&?/ ]{11})/i
+    const matchYoutubeUrl = src.match(youtubeUrlRegExp) || []
```
nicksp: Just a tiny suggestion that will help improve the semantics of code intent:

```js
const matchYoutubeUrl = src.match(youtubeUrlRegExp)
const videoID = matchYoutubeUrl || matchYoutubeUrl[1]
```

This tiny change will help when reading the definition for `matchYoutubeUrl `. When you read it this way it's clear it's gonna be `undefined` or the _actual_ match result. In your case it's unclear it can be undefined and you need to check the definition for another variable to understand that. I hope it's clear ;-) As I said it's tiny but nice improvement. It also doesn't matter much in this particular case, but imagine those vars are spread across: you'll need to scroll to get the intent.

src/components/ui/VideoPlayer.js
```diff
@@ -65,6 +107,49 @@ class VideoPlayer extends PureComponent<Props> {
...
+    const videoID = matchYoutubeUrl[1]
+
+    return overlay && !isOverlayHidden && videoID ? (
```
nicksp: Hmmm I still don't get it. My prev comment applies here. The methos is about rendering embedded video, but it renders overlay under a condition and **not** the video. It's really confusing. I'd prefer code rendering the video **all the time** and overlay **under a condition**, so it's really driven by the `overlay` prop. 

Also, I don't think I clrarly understand the meaning of this long condition `overlay && !isOverlayHidden && videoID`

src/components/ui/WelcomeHero.js
```diff
@@ -0,0 +1,187 @@
...
+import VideoPlayer from './VideoPlayer'
+import GenericTooltip from './GenericTooltip'
+
+import theme, { breakpoints } from '../../styles/theme'
+import typo from '../../styles/typo'
+
+const cmz = require('cmz')
+
+const GAP = '32px'
+const WRAPPER_WIDTH = '1100px'
```
nicksp: Seems like these 2 constants are being re-used across 3 files now which means it might worth adding to a shared file for ease of maintenance.

src/components/ui/WelcomeHero.js
```diff
@@ -0,0 +1,187 @@
...
+      background: none
+    }
+  `)
+}
+
+type Props = {
+  heading?: string,
+  videoUrl?: string,
+  onDismiss?: () => void,
+  children?: React$Element<*>|string
```
nicksp: I believe `Rect$Node` already covers all possible children types.

src/components/ui/WelcomeHero.js
```diff
@@ -0,0 +1,187 @@
...
+}
+
+const WelcomeHero = ({ heading, videoUrl, onDismiss, children }: Props) => heading || children ? (
```
nicksp: Just out of curiosity, I wonder if we need to check for `children` here 🤔 If we just do `heading && ()` it should be fine, as the `Text` component will render nothing which is maybe an accepted case....

src/components/ui/VideoPlayer.js
```diff
@@ -46,11 +85,14 @@ class VideoPlayer extends PureComponent<Props> {
...
-      preload
+      preload,
+      overlay
     } = this.props
 
+    const shouldAutoplay = autoPlay || overlay
+
     return Object.entries({
-      autoplay: autoPlay,
+      autoplay: shouldAutoplay,
```
nicksp: I still don't understand why would `autoplay` prop be driven by `overlay` property. I'd only expect `autoplay` be the case for the embedded video if I set `autoPlay` to `true`. Also, we can rename `autoPlay` to `autoplay`, one word

src/components/ui/VideoPlayer.js
nicksp: What I would expect as a more natual logic is this. Let's say we have this component:

```js
<VideoPlayer src={videoUrl} embedded overlay />
```

From the code, I expect that it renders the embedded video (since `embedded is passed`) and it **puts** an overlay on top of the video (due to `overlay` prop). Overlay button will have a `click` handler than once clicked will delegate `click` (via `ref`) to the underlying video component.

That would make much more sense to me. The current logic is a bit unclear to me. Of course, I understand how it works, I just mean it's confusing to follow along.

src/components/ui/WelcomeHero.js
bernardodiasc: I was expecting this to be better defined in https://x-team-internal.atlassian.net/browse/XP-3297, I can do a temporary move now but I'm worried that without doing it right it can potentially scale bad.

src/components/ui/WelcomeHero.js
bernardodiasc: If I use `React$Node` here I'll have to change Text component and everywhere that uses Text component too. I started with `React$Node` until I get Flow errors. I'm not sure we want all possible children types here or want to make abroad changes.

src/components/ui/WelcomeHero.js
bernardodiasc: The `heading || children` says that both are optional as longer it has at least one of these. So you can use `children` without `heading`. This is a safety condition, it unlikely to have a use case, but helps to keep the UI more bullet-proof. The `heading &&` would have a different result.

src/components/ui/VideoPlayer.js
bernardodiasc: This piece of code is not related to the user expectations. The `autoplay` here is a param in the iframe URL, not the actual flag from prop. The overlay expects the iframe embed to autoplay, because once the iframe is loaded it means the user already triggered the playhead.

src/components/ui/VideoPlayer.js
bernardodiasc: What you expect is what you get from the user point of view, but for the code point of view **it has** to be different. You can search around how to implement such feature, this is the way. The iframe embed only renders after the use hit play in the overlay, then it renders with autoplay. The overlay is a simulation of a literal overlay. Having a literal overlay on top of the iframe will not produce what's expected by design.

src/components/ui/WelcomeHero.js
nicksp: Sure, let's get it defined in the appropriate XP-3297.

src/components/ui/WelcomeHero.js
nicksp: Hmm.. sounds like we don't know what we expect to receive as a `children` prop 🤔

src/components/ui/WelcomeHero.js
nicksp: Gotcha, `heading` is also optional for this particular rendering. Makes sense to have it like it is then 👍

src/components/ui/VideoPlayer.js
```diff
@@ -26,31 +38,63 @@ const videoStyles = cmz(`
...
-      preload
+      preload,
+      overlay
     } = this.props
 
+    const shouldPlayed = autoplay || overlay
+
     return Object.entries({
-      autoplay: autoPlay,
+      autoplay: shouldPlayed,
```
nicksp: typo
```suggestion
      autoplay: shouldBePlayed,
```

src/components/ui/VideoPlayer.js
```diff
@@ -65,45 +109,88 @@ class VideoPlayer extends PureComponent<Props> {
...
-          muted={muted}
-          poster={poster}
-          preload={preload}
-          src={src}
-        >
-          Video cannot be played in this browser.
-        </video>
-      )
+    if (embedded) {
+      const parsedVideoSrc = parseVideoUrl(src)
```
nicksp: Good approach. Nice alternative to what we discussed 😄

src/utils/helpers.js
```diff
@@ -85,3 +85,37 @@ export function timeSince (date: Date | string | number | void | null, addSpaceA
 
   return 'just now'
 }
+
+export function parseVideoUrl (url) {
```
nicksp: I extended signature t have Flow types defined. Other than that it's 👍 
I believe the regexp is solid hard to test without unit tests. Where did you get it?

src/components/ui/WelcomeHero.stories.js
```diff
@@ -0,0 +1,85 @@
...
+    <WelcomeHero
+      heading='WELCOME!'
+      videoUrl={validVideoUrl}
+      onDismiss={action('Handle dismiss')}
+    >
+      Watch X-Team's CEO explain how X-Team makes working from anywhere the most energizing and rewarding experience imaginable.
+    </WelcomeHero>
+  ))
+
+storiesOf('UI Components|WelcomeHero/Debug', module)
```
nicksp: I really like these debug stories as they reveal lots of potential use cases 👍

src/utils/helpers.js
bernardodiasc: Indeed, unit tests here would be great. I'll create a cleanup issue to add that.

bernardodiasc: oh... we don't have any unit tests in XP-UI 😱

bernardodiasc: I created a cleanup issue so we don't lose track of it https://github.com/x-team/xp/issues/2017

nicksp: Great 👍

nicksp: I'll add Jest support next week.

nicksp: Where did you get it regexp from 😄 ? Can we trust this source?

bernardodiasc: XP-UI already have Jest, I think it should be ready to start adding unit tests.

bernardodiasc: > Where did you get it regexp from 😄 ? Can we trust this source?

https://gist.github.com/yangshun/9892961 looks well tested by the community, but for sure we should add unit tests

nicksp: Looks solid, thanks for sharing. Seems like people are using and forking it 🤞

nicksp: > XP-UI already have Jest, I think it should be ready to start adding unit tests.

Great then 🚀  There might probably some tweaking required, but it's not a big deal and we figure that out once starting working.

PR #490, opened 2020-01-13
Add mobile version for HeaderBar

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3494

## Description

Add mobile version for HeaderBar

## Checklist

### in Storybook

1. Go to http://localhost:9001/?path=/story/ui-components-headerbar-use-cases--with-links
1. Resize the viewport to less than `768px` (precisely 767px or less)
1. Also go to http://localhost:9001/?path=/story/screens-and-layouts-applicantscreen-states--with-menu-links

### in Jobs Dashboard

1. Go to http://localhost:8001/applicant/9d92856f67e4c0ff90c265f25b4b7434/jobs (use valid token)
1. Observe that the mobile menu expands and collapses while the content below doesn't change position

- `HeaderBar`


src/components/ui/SvgIcon.js
```diff
@@ -726,6 +726,12 @@ export const getIcons = ({ color, hover, ...rest }: { [x: string]: any } = {}) =
...
         <path className={fillClassName} d='M1.6422 5.39588H2.15833C2.34499 5.39588 2.524 5.32173 2.65599 5.18975C2.78798 5.05776 2.86213 4.87874 2.86213 4.69208C2.86213 4.50542 2.78798 4.32641 2.65599 4.19442C2.524 4.06243 2.34499 3.98828 2.15833 3.98828H1.6422C1.20666 3.98828 0.788963 4.1613 0.48099 4.46927C0.173018 4.77724 0 5.19495 0 5.63049V14.3576C0 14.7932 0.173018 15.2109 0.48099 15.5188C0.788963 15.8268 1.20666 15.9998 1.6422 15.9998H10.3693C10.8049 15.9998 11.2226 15.8268 11.5306 15.5188C11.8385 15.2109 12.0115 14.7932 12.0115 14.3576V13.7946C12.0115 13.6079 11.9374 13.4289 11.8054 13.2969C11.6734 13.1649 11.4944 13.0908 11.3077 13.0908C11.1211 13.0908 10.9421 13.1649 10.8101 13.2969C10.6781 13.4289 10.6039 13.6079 10.6039 13.7946V14.3576C10.6039 14.4198 10.5792 14.4795 10.5352 14.5235C10.4912 14.5675 10.4316 14.5922 10.3693 14.5922H1.6422C1.57998 14.5922 1.52031 14.5675 1.47632 14.5235C1.43232 14.4795 1.4076 14.4198 1.4076 14.3576V5.63049C1.4076 5.56827 1.43232 5.50859 1.47632 5.4646C1.52031 5.4206 1.57998 5.39588 1.6422 5.39588Z' />
       </svg>
+    ),
+
+    menu: (
```
nicksp: Have you seen we have another similar looking (but not yet) of the hamburger menu called `hamburger`? What do you think if we use only one icon (that you added), rename it to `hamburger` and thus replace the existing one 🤔 ? Do you think this will bring lots of breakings to the existing app? To find out, let's check where the existing hamburger icon is used and whether it's fine to use a new one instead. As I can see it's used in one place only - `XHeader` - which seem to not be used in the admin area only, so it might be quite safe to have one hamburger icon only. WDYT? It feels wrong to have almost identical icons representing same thing IMO....

src/components/ui/HeaderBar.js
```diff
@@ -61,16 +94,74 @@ const cx = {
     }
   `),
 
+  hamburger: cmz(`
```
nicksp: In your demo and by checking manually on my machine, I noticed that when you click on the hamburger icon, there's a weird flash of selected text/graphics happen. We can easily prevent that if we add `user-select: none` to it. Quit a neat trick to use 😏

bernardodiasc: This is interesting, but that flash is on a mobile preview (on Chrome debug), usually is something expected and said an anti-pattern to disable. Do you think is preferable to remove?

bernardodiasc: Hmm, this probably should be fine then, will apply this solution.

nicksp: I imagine you could reproduce text selection flashing on a desktop?

bernardodiasc: No no, this should be fine. I think what's anti-pattern to disable is the `outline`, but we are not using the outline here. My confusion.

nicksp: Ah, all righty then 😏

src/components/ui/SvgIcon.js
bernardodiasc: > What do you think if we use only one icon (that you added), rename it to hamburger and thus replace the existing one 🤔 ? Do you think this will bring lots of breakings to the existing app?

I like this idea. This is the only difference in the app:

Current:

Screen Shot 2020-01-16 at 08 52 00

After:

Screen Shot 2020-01-16 at 09 03 42

bernardodiasc:
@nicksp I applied the changes, you can see how the choice of using only the new hamburger icon impacts the app in this comment https://github.com/x-team/xp-ui/pull/490#discussion_r367381543

I think it's alright, but please let me know anything. I pushed changes but I can revert is doesn't looks good.

bernardodiasc:
Also, good point with the observations you put in the latest comment in the ticket. To be honest, that was a mistake of my side 😅 I forgot to prevent the responsive changes for the header bar without links, but actually, I think that's the correct way to go. As you noticed, the `HeaderBar` and `ApplicantScreen` is being used exclusively for applicant-facing screens. Let's see what Paul think about that. :)

src/components/ui/SvgIcon.js
nicksp: Yep, looks totally fine :) No need to keep duplicated icons.

PR #503, opened 2020-03-06
Make TwoColumnsLayout sidebar collapsible

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3571

## Description

Make TwoColumnsLayout sidebar collapsible.

## Checklist

## Steps to Test or Reproduce

Try this changes in XP (copy XP-UI lib files over XP node_modules).

1. Open http://localhost:8000/admin/search on mobile viewport and observe it loads the page with sidebar collapsed
1. Toggle the sidebar by clicking in the splitter
1. Also try out https://deploy-preview-503--xpui-develop.netlify.com/?path=/story/screens-and-layouts-twocolumnslayout-use-cases--composed-in-adminscreen-with-filters-and-sidebarheading-element

To see all details of the implementation, please take a look at this demo https://youtu.be/ZNYKln-_GTo, also make sure to read my comment in the ticket https://x-team-internal.atlassian.net/browse/XP-3571?focusedCommentId=22560&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22560 explaining why AC 2 and 5 are not covered. (at this moment I proposed to remove these items, still waiting feedback)

`TwoColumnsLayout`


src/components/ui/TwoColumnsLayout.js
```diff
@@ -149,8 +217,56 @@ class TwoColumnsLayout extends PureComponent<Props, void> {
...
 
+  state: State = {
+    initial: true,
+    mobile: false,
+    expanded: true,
+    sidebarWidth: this.props.sidebarWidth
+  }
+
+  componentDidUpdate (prevProps: Props, prevState: State) {
+    if (!isEqual(prevState, this.state.expanded)) {
```
jaspax: Shouldn't this just be `if (prevState.expanded != this.state.expanded)` ? Otherwise you're comparing the entire state object against the `expanded` property, which will never be equal.

src/components/ui/TwoColumnsLayout.js
```diff
@@ -149,8 +217,56 @@ class TwoColumnsLayout extends PureComponent<Props, void> {
...
+      // This is required for recalculation of react-data-grid width
+      window.dispatchEvent(new Event('resize')) // eslint-disable-line
```
jaspax: What specifically needs to be disabled here? Can you update this to disable ONLY that check?

src/components/ui/TwoColumnsLayout.js
```diff
@@ -149,8 +217,56 @@ class TwoColumnsLayout extends PureComponent<Props, void> {
...
+
+  componentWillUnmount () {
+    window.removeEventListener('resize', this.updateDimensions)
+  }
+
+  updateDimensions = () => {
+    const { sidebarWidth } = this.props
+    const { initial, expanded } = this.state
+    const { innerWidth } = window
+    const mobileBreakpoint = Number(breakpoints.sm.replace('px', ''))
```
jaspax: On a recent changelist I used `breakpoints.md` for enabling mobile behavior. Let's start a discussion to standardize where we want mobile behavior to start, and then create an entry `breakpoints.mobile` to use throughout the product.

src/components/ui/TwoColumnsLayout.js
```diff
@@ -149,8 +217,56 @@ class TwoColumnsLayout extends PureComponent<Props, void> {
...
+    if (innerWidth <= mobileBreakpoint) {
+      this.setState({
+        mobile: true,
+        expanded: initial ? false : expanded,
+        sidebarWidth: maxSidebarWidth
+      })
+    } else {
+      this.setState({
+        mobile: false,
+        expanded: initial ? true : expanded,
```
jaspax: Since `expanded` initializes to true, you can make this just be `expanded`

src/components/ui/TwoColumnsLayout.js
```diff
@@ -213,10 +337,26 @@ class TwoColumnsLayout extends PureComponent<Props, void> {
     )
   }
 
+  handleLayout = () => {
```
jaspax: I'd prefer a more descriptive name, such as "toggleExpanded"

src/components/ui/TwoColumnsLayout.js
nicksp: valid point ➕

nicksp:
> the target for toggling the sidebar is pretty small.

One of the options could be increasing the padding area to give fingers enough room to make a click. cc @bernardodiasc 

Out of interest @jaspax, what kind of device are you using for testing?

src/components/ui/TwoColumnsLayout.js
bernardodiasc: oops, it should be `!isEqual(prevState, this.state)`.

src/components/ui/TwoColumnsLayout.js
bernardodiasc: This is a good point for discussion, I grab the code style from there indeed. But the fact is that changes here relates much more to available viewport for the sidebar, so `768px` is the breakpoint desired for the feature here. In your case, I would say `1024px` is perfectly fine. Both are mobile nowadays, but sidebar should be fine to appear expanded in viewport above 768px.

bernardodiasc: Actually, let me try this effect for `breakpoints.md` instead. Could be better...

src/components/ui/TwoColumnsLayout.js
bernardodiasc: This will not result the same UX. The `expanded` can be false while `initial` still true.

bernardodiasc:
Bigger button:

src/components/ui/TwoColumnsLayout.js
nicksp: As suggested on Jesse's PR, I suggest we rewrite it like that to make more concise: `Number.parseInt(breakpoints.sm, 10)` @bernardodiasc

src/components/ui/TwoColumnsLayout.js
```diff
@@ -149,8 +217,56 @@ class TwoColumnsLayout extends PureComponent<Props, void> {
...
+
+  componentDidUpdate (prevProps: Props, prevState: State) {
+    if (!isEqual(prevState, this.state)) {
+      // This is required for recalculation of react-data-grid width
+      window.dispatchEvent(new Event('resize'))
+    }
+  }
+
+  componentWillMount () {
+    this.updateDimensions()
```
nicksp: [componentWillMount](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) is a legacy lifecycle hook that isn't recommended to use anymore. Avoid introducing any side-effects or subscriptions in this method. For those use cases, use `componentDidMount()` instead (as you already do).

bernardodiasc:
@nicksp about this:

> Not sure I'm missing anything, but it's doesn't seem that the sidebar auto-collapses (when switching to mobile), even though I can see a code that does subscribe to resize event, but doesn't seem to hook up 🤔

Can you demonstrate the problem you are having, it works for me:





But observe that once you interact with the sidebar by collapsing or expanding, your, as user,  preferences will be persisted.

nicksp:
Hey @bernardodiasc Why isn't this working for me?

PR #505, opened 2020-03-18
Add AccountConfirmation component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3596

## Description

Add `AccountConfirmation` component and upgrade `Modal` with white theme.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-3596-allow-users-to-logout-on-reg-form-and-jobs-pages` | [XP Registration](https://github.com/x-team/xp-registration/pull/159)

## Steps to Test or Reproduce

- Check `AccountConfirmation` at http://localhost:9001/?path=/story/ui-components-accountconfirmation--basic-usage
- Check Modal with white theme at http://localhost:9001/?path=/story/core-components-modal-states--white-theme

- `src/components/ui/AccountConfirmation.js`
- `src/components/ui/Modal.js`

## Screenshots

Screen Shot 2020-03-18 at 12 51 24


src/components/ui/AccountConfirmation.js
```diff
@@ -0,0 +1,136 @@
...
+
+  render () {
+    const { avatar = '', fullName = '', email = '' } = this.props
+    return (
+      <div className={cx.wrapper}>
+        <h1 className={cx.title}>Ready to continue<br />your sign up?</h1>
+        {(avatar || fullName) && (
+          <span className={cx.avatar}>
+            <Avatar
+              src={avatar || fullName}
```
nicksp: what happens if we pass a string (`fullName` in this case) to the `Avatar` component? What is it gonna render in this case?

src/components/ui/AccountConfirmation.js
```diff
@@ -0,0 +1,136 @@
...
+              alt={fullName}
+              size={80}
+            />
+          </span>
+        )}
+        {fullName && (
+          <h2 className={cx.fullName}>{fullName}</h2>
+        )}
+        {email && (
+          <a className={cx.email} href='mailto:'>{email}</a>
```
nicksp: missing `mailto:` link?

src/components/ui/AccountConfirmation.js
```diff
@@ -0,0 +1,136 @@
...
+  handleSignOut = () => {
+    const { onSignOut } = this.props
+    onSignOut && onSignOut()
+  }
+
+  render () {
+    const { avatar = '', fullName = '', email = '' } = this.props
+    return (
+      <div className={cx.wrapper}>
+        <h1 className={cx.title}>Ready to continue<br />your sign up?</h1>
```
nicksp: I wonder if this is the version of text we want to use 'cause in Figma I see a different representation? Was it discussed with Paul we'll want to change that? 

image

src/components/ui/AccountConfirmation.js
bernardodiasc: It renders the initial letters uppercased on top of a round and coloured background. It's the default behaviour of the Avatar component.

src/components/ui/Modal.js
```diff
@@ -71,19 +93,26 @@ class Modal extends PureComponent<Props, void> {
...
       >
-        <section className={cx.frame} onClick={this.noClick}>
-          <a className={cx.close} onClick={this.handleClose}>
-            <SvgIcon icon='x' color='grayscale' />
-          </a>
+        <section
+          className={[cx.frame, cx.theme[theme].frame].join(' ')}
+          onClick={this.noClick}
+        >
+          {onClose && (
```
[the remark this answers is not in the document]
bernardodiasc: It does. The AccountConfirmation uses the Modal without the X button, and it's also not expected to be closable when clicking on modal's overlay, so `onClose` can be omitted and this button should not appear.

src/components/ui/AccountConfirmation.js
nicksp: by looking at changes, I presume you used a `png` as a reference, right? So, it wasn't mentioned by Paul that we should refer to the png instead of the Figma mocks, was it?

src/components/ui/Modal.js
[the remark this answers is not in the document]
nicksp: Gotcha, that makes it clear 👍

nicksp:
@bernardodiasc  Please get merge conflicts resolved

src/components/ui/AccountConfirmation.js
```diff
@@ -0,0 +1,138 @@
...
+
+type Props = {
+  avatar?: string,
+  fullName?: string,
+  email?: string,
+  onContinue?: () => void,
+  onSignOut?: () => void
+}
+
+class AccountConfirmation extends PureComponent<Props, void> {
```
nicksp, line 99: As a quick note, using `PureComponent` will have no benefits here as we're passing functions that are not memozied and thus represents different objects that will still cause the component to re-render.

bernardodiasc:
Maybe we should stick to default button instead add customs

Screen Shot 2020-04-08 at 11 38 59

Talking with @mccrodp he confirmed that the Figma design should not be the source of truth for UI development, the source of truth should be existing patterns instead.

nicksp:
> Maybe we should stick to default button instead add customs

The fix was quite easy to proceed with so, I'd resolved that already. No probs 😛

bernardodiasc:
> I believe his message was in relation to cases when it's much more effort to match designs

That's not the case, you can confirm with Paul. The goal is to use more UI patterns and less custom code. You can think of if we need to change the font-size of all buttons at once, if we have custom code a move like this is much more complicated. Without customs we would just need to update 1 place. This concern exists because there's a rebranding coming soon, which might make this design details discarded anyway, better to keep it simple for our future selves.

PR #506, opened 2020-03-24
Make the header bar appear above all elements on the profile page

jaspax:
**Release Type:** *Bug Fix* 

Fixes https://x-team-internal.atlassian.net/browse/XP-3657

## Description

Set the z-index to the HeaderBar component to 1000. This is greater than
every UI component that appears on that page.

Remove the z-index from the element that contains the HeaderBar on the
ApplicantInfo page. Setting a z-index on that element creates a new
"stacking context" in CSS which makes the z-index for contained
elements do the wrong thing.

## Checklist

## Steps to Test or Reproduce

Visit `http://localhost:8001/applicant/:token/info` in your browser. Note that as you scroll down input elements appear above the bar at the top of the streen.

Build these changes, and copy the generated files in `lib/` into `xp-registration/site/node_modules/xp-ui/lib`.

Now refresh the `applicant/:token/info` page. This time you can scroll up and down and input elements never go above the top bar.

Applicant profile ("info") page.

## Screenshots

scroll-z-index


src/components/ui/HeaderBar.js
```diff
@@ -48,6 +48,7 @@ const cx = {
         background: ${theme.baseBrighter}
         position: fixed
         top: 0
+        z-index: 1000
```
nicksp: Is this really needed? Seems like if we change it inside `ApplicantScreen` from `z-index; 2` to `1000` it does the trick 
image

HeaderBar is being called only from `AplicantScreen` so it should be safe to increase existing value `2` to a higher number. Wdyt?

nicksp:
@jaspax we should not create PRs against `master` as it's only happens when we release. This bug is reproducible on `develop` branch as well (I just confirmed that) and hence should be against `develop`.

PR #514, opened 2020-04-07
Add SkillsSelector component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3570

## Description

Add SkillsSelector component

## Checklist

## Steps to Test or Reproduce

- Check the component at http://localhost:9001/?path=/story/core-components-form-components-skillsselector--basic-usage

Please confirm expected behaviour from https://x-team-internal.atlassian.net/browse/XP-3570 AC.

New `SkillsSelector` component added.


nicksp:
Hey @bernardodiasc Please resolve merge conflicts

bernardodiasc:
@nicksp this branch is rebased and conflicts fixed now

nicksp:
@bernardodiasc From CI:

image

bernardodiasc:
Yeah, it was a package-lock conflict, it should be fixed now. npm install is now working.

bernardodiasc:
Adjustments has been made:

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+  componentDidUpdate (prevProps: Props) {
+    const { values } = this.props
+    const { values: prevValues } = prevProps
+    if (values !== prevValues) {
+      this.setState(() => ({ values }))
+    }
+  }
+
+  handleChange = (values: Array<Option>) => {
+    const diff = differenceBy(values, this.state.values, 'value')
```
nicksp, line 193: Could we please rename this variable as well as the other one in `handleRemove` to reflect what diff it contains in it? That would simplify code reading a lot, thanks/

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+  static defaultProps = {
+    options: [],
+    values: []
+  }
+
+  state: State = {
+    values: this.props.values || []
+  }
+
+  componentDidUpdate (prevProps: Props) {
```
nicksp, line 184:
```suggestion
  componentDidUpdate ({ values: prevValues }: Props) {
```

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+}
+
+type State = {
+  values: Array<Option>
+}
+
+class SkillsSelector extends Component<Props, State> {
+  static defaultProps = {
+    options: [],
+    values: []
```
nicksp, line 177: What are `values` here? Skills? Let's make it clear. I think seeing meaningful props is beneficial to understand what values each component expects. Just a thought. Without seeing the code it's hard to tell what values are.

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  optionComponent = (component: Object) => {
+    const handleMouseDown = (event) => {
+      event.preventDefault()
+      event.stopPropagation()
+      component.onSelect(component.option, event)
+    }
+
+    const handleMouseEnter = (event) => {
```
nicksp, line 220: Do you think we could [type enhance](https://flow.org/en/docs/react/events/) `event` param here and there to be clear about expectations?

flow-typed/npm/lodash.some_vx.x.x.js
```diff
@@ -0,0 +1,33 @@
+// flow-typed signature: 629358ea5117d0c9f2e90e75edbbb7ad
```
nicksp, line 1: Seems like this file isn't needed anymore, is it?

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+}
+
+class SkillsSelector extends Component<Props, State> {
+  static defaultProps = {
+    options: [],
+    values: []
+  }
+
+  state: State = {
+    values: this.props.values || []
```
nicksp, line 181: Do you think it may be redundant given we have `defaultProps` that will catch not passed props? 🤔 Did you face any problem with not specifying it again with this line of code?

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  componentDidUpdate (prevProps: Props) {
+    const { values } = this.props
+    const { values: prevValues } = prevProps
+    if (values !== prevValues) {
+      this.setState(() => ({ values }))
```
nicksp, line 188: This whole `cdu` lifecycle event seems like it might not be needed. What is the use case you're solving with having it?

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+
+  handleChange = (values: Array<Option>) => {
```
nicksp, line 192: I find this name too generic. Is it easier to read when named smth like `handleSkillSelection`?

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+    const diff = differenceBy(values, this.state.values, 'value')
+    if (diff.length > 0) {
+      const newValues = [ ...this.state.values, ...diff ]
+      this.setState(() => ({ values: newValues }), () => {
+        const { onChange } = this.props
+        onChange && onChange(newValues)
+      })
+    } else {
+      this.handleRemove(values)
```
nicksp, line 201: I wonder if this is needed at all?! It was the case when clicking same item from the selectbox could add or remove item, but now it's only one operation.

src/components/ui/SkillsSelector.js
nicksp, line 193: Do we even need to calculate diff now? The logic should be like that. Once you select a skill, local state should be updated to filter that out from the original list of skills we passed in. Re-render will happen to reflect that, so select box will always show up to date list of skills.

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+
+  handleRemove = (values: Array<Option>) => {
+    const diff = differenceBy(this.state.values, values, 'value')
```
nicksp, line 206: same here. Once we remove a selected skill tag by clicking `X`, we update the state and that's all we need to do I guess. Do we really need to calculate anything here?

src/components/ui/SkillsSelector.js
bernardodiasc: It might be needed because this is not a regular react-select case. I removed atm because I can't tell any actual use case now, but we might experience issues if consumer attempt to update selected values via props.

src/components/ui/SkillsSelector.js
bernardodiasc: This is safe to be removed, but just to clarify, the `values` here is not exactly the entire list of selected skills, because the react-select in this component is not holding selected values. That's because, by design, the selected values are not present in the select component, so they are stored in the state instead.

src/components/ui/SkillsSelector.js
nicksp: We seem to not expect that component to be updated in any way (in terms of skills), so it seems legit to not have that method. Skills will be passed upon mounting and that's where they seem to be fulfilled.

src/components/ui/SkillsSelector.js
nicksp: Do you mean that we do not have a list of original skills passed to the component and a list of filtered skills we keep in local state is also not around? We should simply be updating the local state of skills (when adding skill tags or removing them) and always depend on it. This way we don't need any extra calculations like `differenceBy`. Am I missing anything?

bernardodiasc: sorry, I don't get what's currently different from what you said...

nicksp: Why do you need to use `differenceBy` at all? This seems to be extraneous. What you should do instead is update local state when you select a skill (by filtering out selected skill) and update local state again when you remove selected skill (by adding that to the local state skills list). Then, you simply operate on a local state. Don't see the need for calculating that diff.

bernardodiasc: This was needed in previous implementation, there's no `differenceBy` here anymore...

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,269 @@
...
+type State = {
+  selectedSkills: Array<Option>
+}
+
+class SkillsSelector extends Component<Props, State> {
+  state: State = {
+    selectedSkills: this.props.applicantSkills || []
+  }
+
+  handleSkillSelection = (selectedSkills: Array<Option>) => {
```
nicksp, line 179: I wonder if it's meant to be one skill (as the name and logic imply) or an array of skills? Please adjust the code accordingly.

bernardodiasc: This is how react-select works, it will be always passing an array of objects.

nicksp: This is odd.. but whatever. Can we in this case rename the variable at least tp denote it's one skills, not skills?

nicksp: notice, that in reality we always get **one** skill back.

nicksp: Also, check whether we need to pass `multi` prop to Selector UI control or not. I believe that is the thing that controls whether is passes an array or a single element back.

src/components/ui/SkillsSelector.js
```diff
@@ -0,0 +1,268 @@
...
+
+  handleSkillRemove = (removedSkill: Option) => {
+    const selectedSkills = this.state.selectedSkills.filter(skill => skill.value !== removedSkill.value)
+    this.setState(() => ({ selectedSkills }), () => {
+      const { onChange } = this.props
+      onChange && onChange(selectedSkills)
+    })
+  }
+
+  optionComponent = (component: Object) => {
```
nicksp, line 195: This seems like a candidate to be moved to a separate file cause it's self-contained component with encapsulated actions inside. Once removed, the code for `SkillsSelector` will be much concise.

bernardodiasc: Is there a real benefit from doing this?

nicksp: Code is easier to reason about + more opportunities for optimization if needed. Why don't you think it's beneficial?

PR #523, opened 2020-04-22
Move CustomSelect from XP-Reg to XP-UI

kararade:

Fixes https://x-team-internal.atlassian.net/browse/XP-3702

## Description

- Create `CustomSelector` component
- Rename to a more general name the `SkillsSelectorOption` component to just `SelectorOption`

## Checklist

**Before submitting a Pull Request,** please make sure the following is done:

- [ ] if you've fixed a bug, make sure to also include a new story that will expose declarations with problematic data that caused the bug in the first place, so that we can assure that no regressions will pop up on future interactions

## Related PRs

branch | PR
------ | ------
xp_registration_branch_name | [XP Registration](https://github.com/x-team/xp-registration/pull/186)

## Steps to Test or Reproduce

Test the component in http://localhost:9001/?path=/story/core-components-form-components-customselector--basic-usage

- `xp-ui/src/components/ui/CustomSelector.js` > Created
- `xp-ui/src/components/ui/CustomSelector.stories.js` > Created
- `xp-ui/src/components/ui/SelectorOption.js`> Renamed
- `xp-ui/src/components/ui/SelectorOption.stories.js`> Renamed
- `xp-ui/src/components/ui/SkillsSelector.js` > Updated `SelectorOption` component name 

## Screenshots

* Basic usage

Screen Shot on 2020-04-22 at 04:39:48

Screen Shot on 2020-04-22 at 04:40:04

Screen Shot on 2020-04-22 at 04:40:18

* States > Clearable

Screen Shot on 2020-04-22 at 04:40:42

* States > Searchable

Screen Shot on 2020-04-22 at 04:41:06

* States > Disabled

Screen Shot on 2020-04-22 at 04:41:33


kararade:
> There are only 1 thing, that I believe it's not a blocker but something we should clarify with @mccrodp. The visual effect for mouse hover in the element. When I built the SkillsSelector I followed what exists in https://www.figma.com/file/HdK6csNOupqLMyLxOykrPN/XP-Registration-app?node-id=269%3A193
> 
> So we got this effect (I'm really sure if this is the correct expectation!):
> 
> 
> This is what currently have in CustomSelector:
> 

I've removed that hover effect actually, but, I will add it just to keep the consistency meanwhile we discuss about to improve the styles in the ticket https://x-team-internal.atlassian.net/browse/XP-3705 😉

> And strangely, this is how CustomSelector hover effect is implemented in XP-Reg (observe that it's different from what happens in Storybook):
> 

Maybe in XP-Reg are some styles overriding the component's styles. I will double check

Thanks!

nicksp:
@carlosalmonter @bernardodiasc Seems like we'll want to update state of input here, right? So we don't apply red border on focus as it's an indication of an error. More details are in #xp-team-collab:

nicksp:
@kararade  @bernardodiasc Seems like we'll want to update state of input here, right? So we don't apply red border on focus as it's an indication of an error. More details are in #xp-team-collab Please move back to `in progress` in this case to reflect it.

bernardodiasc:
> Seems like we'll want to update state of input here, right?

@nicksp that's not the case, the effect is not applied in this component, the suggestion was to not implement at this stage. This PR should be in accordance with expectations. 👍

bernardodiasc:
@nicksp to be clearer, there are few details in the styles that are not perfect. That condition applies to this component but also to other input. The suggestion is to not hold progress because of that given that we was already expecting inconsistencies like that. Instead blocking this I suggested to raise concerns in the scope of https://x-team-internal.atlassian.net/browse/XP-3705 and with appropriate design direction we could fix all input components to have consistent styles.

nicksp:
> Instead blocking this I suggested to raise concerns in the scope of https://x-team-internal.atlassian.net/browse/XP-3705

@bernardodiasc Totally agree. Please update the ticket accordingly to take into consideration those design changes.

PR #525, opened 2020-04-24
Create Timeframe component

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3701

## Description

Create Timeframe component.

## Checklist

## Steps to Test or Reproduce

- Check the showcase in Storybook http://localhost:9001/?path=/story/core-components-form-components-timeframe--basic-usage

Please take a look at the other Debug showcases, it has some edge cases.

For instance, the start date should never be bigger than an end date. So: 

- If the **start year** is the same of the **end year**, then only the available months to select in the **end month** will be bigger or equal than the **start month**
- If the **start year** is the smaller of the **end year**, all **end month** with be available
- The **end year** options will only have years equal or bigger than **start year**
- If the `Still working there` is checked, then the value returned will ignore the end date (as end date will be undefined)

New `Timeframe` component was added.


jaspax:
I tried to use this component and noticed a few things that were off:

* The year selector is sorted in inverse order with the highest (ie most recent) years at the top. I can see why you might think this is a good idea, but it's contrary to my expectations and caused confusion. Can we get a double-check whether this is a good idea?
* When clicking, "Still working there", can we clear out the values for the field with blanks for dashes?

src/components/ui/Timeframe.js
```diff
@@ -0,0 +1,283 @@
...
+  `)
+}
+
+type Option = {
+  label: string,
+  value: string | number
+}
+
+type Props = {
+  startDate?: string,
```
jaspax, line 98: Shouldn't `startDate` be required? We never want startDate to be null.

src/components/ui/Timeframe.js
```diff
@@ -0,0 +1,283 @@
...
+  endDate?: string,
```
jaspax, line 99: More importantly, why are these props STRINGS?! They really ought to be `Date` unless there's some really good reason that I'm not seeing.

src/components/ui/Timeframe.js
```diff
@@ -0,0 +1,283 @@
...
+
+  getOptionValue = (options: Array<Option> = [], value: ?string | ?number) => options.find(option => option.value === value)
+
+  updateDateValues = () => {
+    const { startDate = '', endDate = '', noEndDate = false } = this.props
+
+    const startMonth = Number(startDate.slice(5, 7))
+    const startYear = Number(startDate.slice(0, 4))
+    let endMonth = Number(endDate.slice(5, 7))
+    let endYear = Number(endDate.slice(0, 4))
```
jaspax, line 157: Slicing date strings to get values seems extremely fragile. Please use a `Date`.

src/components/ui/Timeframe.js
```diff
@@ -0,0 +1,283 @@
...
+type Option = {
+  label: string,
+  value: string | number
+}
+
+type Props = {
+  startDate?: string,
+  endDate?: string,
+  noEndDate?: boolean,
+  onChange?: ({ startDate: string, endDate: ?string }) => void
```
jaspax, line 101: The `onChange` event should likewise take `Date` arguments rather than strings

src/components/ui/Timeframe.js
```diff
@@ -0,0 +1,283 @@
...
+
+type Option = {
+  label: string,
+  value: string | number
+}
+
+type Props = {
+  startDate?: string,
+  endDate?: string,
+  noEndDate?: boolean,
```
jaspax, line 100: In the `onChange` handler we're handling the "no end date" case simply by passing a null `endDate`. Why not do this here? Why do we need this property at all?

src/components/ui/Timeframe.js
```diff
@@ -0,0 +1,283 @@
...
+    }, this.validateValues)
+  }
+
+  updateFieldValue = (field: string, options: Array<Option> = []) => ({ value }: { value: string }) => {
+    this.setState({
+      [field]: this.getOptionValue(options, value)
+    }, this.validateValues)
+  }
+
+  validateValues = () => {
```
jaspax, line 174: Why is this called `validateValues`? I don't see any validation going on here. Since method seems to do a simple call to `setState` and then set the callback, and I'm not sure why it couldn't just be unified with `handleOnChange`.

src/components/ui/Timeframe.js
bernardodiasc: It's null when the user haven't added any start date yet.

src/components/ui/Timeframe.js
bernardodiasc: The end date can be undefined before the user add a value and that doesn't mean the checkbox that disable the field should be checked.

src/components/ui/Timeframe.js
bernardodiasc: This method handles cases where end dates are previous than start dates. Because there are 4 fields and the user can simply set these in any order, we need to validate those situations. I provided some explanation regarding these situations in the demo https://youtu.be/I6qvFBPE2K4, please let me know if you think something should be different from that, thanks!

jaspax: Okay, I had to re-read the function to see how this method actually accomplishes that, because it's not clear from the names how `this.getEndMonthsList` is actually validating anything. Maybe add something to the name to make the semantics clearer.

src/components/ui/Timeframe.js
```diff
@@ -0,0 +1,283 @@
...
+    const { startMonth, startYear, endMonth, endYear } = this.state
+    this.setState({
+      startMonth: startMonth && this.getOptionValue(MONTHS, startMonth.value),
+      startYear: startYear && this.getOptionValue(YEARS, startYear.value),
```
jaspax, line 178: The checks for `startMonth` and `startYear` are completely redundant as far as I can tell.

src/components/ui/Timeframe.js
jaspax: I mean the functions `getEndMonthsList`. Maybe something like `getValidEndMonths`

bernardodiasc:
hey @jaspax, thanks for the feedback. I addressed your suggestions. Given the discussion in the dev-design collab channel, here's how the component looks like now:

PR #538, opened 2020-05-13
Admin work experience styles

jaspax:

Part of https://x-team-internal.atlassian.net/browse/XP-3737

## Description

Made a number of style improvements to make the WorkExperience components integrate better into the XP Admin stylesheet.

- remove typographical styles so that text inherits its styles from the container
- move the pencil icon which switches to edit mode onto the right side of the header (requires changes to `WorkExperiencePresenter`)
- change header styles to match other headers used in XP
- remove unnecessary bottom borders
- minor tweaks to margins around several elements

## Checklist

## Related PRs

branch | PR
------ | ------
XP-3401-admin-work-experience | [XP](https://github.com/x-team/xp/pull/2072)

## Steps to Test or Reproduce

View the steps on the XP pull request above to integrate into XP and view the styles in their intended context.

XP admin interface.

## Screenshots

image

Note changes:
- font size in the Work Experience section is the same as in the previous profile section
- the bolded role name has the same style as the profile question preceding it
- pencil icon is on the right of the header instead of the left
- no more extra margin on the left side of the work experience entries

### DIFF
--- a/src/components/ui/Timeframe.js
+++ b/src/components/ui/Timeframe.js
@@ -92,7 +92,7 @@ type Option = {
 
 type Props = {
   startDate?: Date,
-  endDate?: Date,
+  endDate?: ?Date,
   noEndDate?: boolean,
   disabled?: boolean,
   onChange?: ({ startDate: Date, endDate: ?Date }) => void
--- a/src/components/ui/WorkExperience/index.js
+++ b/src/components/ui/WorkExperience/index.js
@@ -61,8 +61,12 @@ const renderEditor = ({ value: { experience, skills }, onValueChange }: EditorPr
 )
 
 const renderPresenter = ({ value: { experience }, activateEditingMode, isHover }: PresenterProps) => (
-  <WorkExperienceEntryPresenter experience={experience}
-    headerDecoration={isHover && <PencilButton color='monochrome' onClick={activateEditingMode} />} />
+ <WorkExperienceEntryPresenter
+    experience={experience}
+    headerDecoration={isHover && (
+      <PencilButton color='monochrome' onClick={activateEditingMode} />
+    )}
+  />
 )
 
 class WorkExperience extends Component<Props, State> {
--- a/src/components/ui/WorkExperience/index.js
+++ b/src/components/ui/WorkExperience/index.js
@@ -17,12 +17,14 @@ import typo from '../../../styles/typo'
 
 export type Props = {
   experiences: ?Array<Experience>,
+  initialVisibleExperiences: number,
   skills: Array<SkillOption>,
   onSave: ({ experience: Experience, isNew: boolean }) => Promise<any>
 }
 
 type State = {
-  showNewExperience: boolean
+  showNewExperience: boolean,
+  visibleExperiences: number
 }
 
 // This stub experience is used as a property when creating new experience
@@ -43,6 +45,12 @@ const cx = {
     margin-bottom: 1em
     padding-top: 0.4em
     padding-bottom: 0.6em
+  `),
+
+  buttons: cmz(`
+    & * {
+      margin-top: 0.4em
+    }
   `)
 }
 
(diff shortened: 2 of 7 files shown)


src/components/ui/WorkExperience/index.js
```diff
@@ -66,17 +61,18 @@ const renderEditor = ({ value: { experience, skills }, onValueChange }: EditorPr
...
 
 const renderPresenter = ({ value: { experience }, activateEditingMode, isHover }: PresenterProps) => (
-  <div className={cx.presenterWrapper}>
-    <div className={cx.editGutter}>
-      {isHover && <PencilButton color='monochrome' onClick={activateEditingMode} />}
-    </div>
-    <WorkExperienceEntryPresenter experience={experience} />
-  </div>
+  <WorkExperienceEntryPresenter experience={experience}
+    headerDecoration={isHover && <PencilButton color='monochrome' onClick={activateEditingMode} />} />
```
bernardodiasc, line 65: It's easier to read and edit when JSX syntax is better presented.

```suggestion
 <WorkExperienceEntryPresenter
    experience={experience}
    headerDecoration={isHover && (
      <PencilButton color='monochrome' onClick={activateEditingMode} />
    )}
  />
```

bernardodiasc: Oops, sorry, looks like the suggestion block had wrong white-spaces for the indentation 😓 
```
  64:1  error  Expected indentation of 2 spaces but found 1
  64:2  error  Expected indentation of 2 space characters but found 1
  65:5  error  Expected indentation of 3 space characters but found 4
  66:5  error  Expected indentation of 3 space characters but found 4
```

jaspax:
@nicksp Caused by accepting bernardo's suggestion through the GH UI rather than using the command-line. Fixed.

nicksp:
@jaspax Where the PR for the parent `XP-3401-admin-work-experience` branch? Am I missing anything?

src/components/ui/Timeframe.js
```diff
@@ -92,7 +92,7 @@ type Option = {
 
 type Props = {
   startDate?: Date,
-  endDate?: Date,
+  endDate?: ?Date,
```
nicksp, line 95: please choose either Optional or Maybe type, both is confusing

jaspax: Eliminating either `?` causes a bunch of Flow errors. It turns out that we want to support both ways of specifying "no end date": both "don't pass the `endDate` prop" and "pass the `endDate` prop with `null` value", hence the type definition given above.

jaspax:
@nicksp The `XP-3401-admin-work-experience` branch doesn't have a pull request because so far it's empty. Hopefully it will have something once its children get merged in.

jaspax:
Closing this and continuing it in the proper task branch here: https://github.com/x-team/xp-ui/pull/545

src/components/ui/Timeframe.js
nicksp: @jaspax  How is that possible to support both? I mean, from what I can see, we either provide end date or not. So smth like that: `endDate?: Date`. Can you explain how it can be possible in our flow to provide a bull value? A date should be a valid date in case applicant decide to provide it 🤔

jaspax: Sometimes we want to write `<Timespan startDate={val} />` without providing the `endDate` prop at all. This requires that the `endDate` prop itself be optional, ie `endDate?: Date`.

Sometimes we want to write `<Timespan startDate={startDate} endDate={endDate} />`, where the value passed to `endDate` is potentially null (because it came from a server response). This requires that the value for the `endDate` prop be `?Date`.

Put them together, and we need `endDate?: ?Date`.

nicksp: The only thing I do not understand is the possibility for the `endDate` to be `null`. That's our task I believe to not allow for that actually. In case we provide value, it should be set. We do not allow for an empty end date. Can you show me in code how come it can be true to have `endDate` to be `null`? This code is new so it's up to us to specify its logic and behavior hence I'm trying to get to the root.

jaspax: The "Still working there" experience entry is indicated by `endDate=null`. This is normal and expected behavior.

nicksp: Hmmm, OK. I thought the opposite. Anyways, this is confusing to have both specified, but please keep it.

jaspax: Within the app itself, we pretty much always pass `endDate`, so the cleaner recommendation of `endDate: ?Date` would be possible. However, there are a bunch of stories which don't pass the `endDate` prop at all. I would have to update all of those if we change the definition, which is out of scope for this changelist, which doesn't directly concern the Timeframe component at all. You can file a cleanup task if you'd like.

PR #545, opened 2020-05-25
Admin work experience styles

jaspax:

Part of https://x-team-internal.atlassian.net/browse/XP-3737

## Description

Made a number of style improvements to make the WorkExperience components integrate better into the XP Admin stylesheet.

- remove typographical styles so that text inherits its styles from the container
- move the pencil icon which switches to edit mode onto the right side of the header (requires changes to `WorkExperiencePresenter`)
- change header styles to match other headers used in XP
- remove unnecessary bottom borders
- minor tweaks to margins around several elements

## Checklist

## Related PRs

branch | PR
------ | ------
XP-3401-admin-work-experience | [XP](https://github.com/x-team/xp/pull/2072)

## Steps to Test or Reproduce

View the steps on the XP pull request above to integrate into XP and view the styles in their intended context.

XP admin interface.

## Screenshots

image

Note changes:
- font size in the Work Experience section is the same as in the previous profile section
- the bolded role name has the same style as the profile question preceding it
- pencil icon is on the right of the header instead of the left
- no more extra margin on the left side of the work experience entries

### DIFF
--- a/src/components/ui/CustomSelector.js
+++ b/src/components/ui/CustomSelector.js
@@ -71,7 +71,7 @@ const cx = {
 
       & .Select-input > input {
         height: 100%
-        padding: 0px !important
+        padding: 0 !important
       }
 
       & .Select-input:focus {
--- a/src/components/ui/CustomSelector.js
+++ b/src/components/ui/CustomSelector.js
@@ -49,7 +49,7 @@ const cx = {
         overflow-y: hidden
         display: flex
         align-items: center
-        height: 18px
+        height: 100%
       }
 
       & .Select-placeholder,
@@ -59,12 +59,19 @@ const cx = {
         white-space: nowrap !important
         overflow: hidden !important
         text-overflow: ellipsis !important
+        height: 100%
       }
 
       & .Select-input {
-        padding: 0 3px !important
-        position: absolute !important
-        top: 11px !important
+        padding: 0 !important
+        display: flex !important
+        align-items: center
+        height: 100%
+      }
+
+      & .Select-input > input {
+        height: 100%
+        padding: 0px !important
       }
 
       & .Select-input:focus {
--- a/src/components/ui/Timeframe.js
+++ b/src/components/ui/Timeframe.js
@@ -92,7 +92,7 @@ type Option = {
 
 type Props = {
   startDate?: Date,
-  endDate?: Date,
+  endDate?: ?Date,
   noEndDate?: boolean,
   disabled?: boolean,
   onChange?: ({ startDate: Date, endDate: ?Date }) => void
--- a/src/components/ui/WorkExperience/WorkExperienceEntryPresenter.js
+++ b/src/components/ui/WorkExperience/WorkExperienceEntryPresenter.js
@@ -6,40 +6,35 @@ import cmz from 'cmz'
 
 import JobSkills from '../JobSkills'
 import theme from '../../../styles/theme'
-import typo from '../../../styles/typo'
 
 import type { Experience } from '../../../utils/types'
-type Props = { experience: Experience }
+type Props = {
+  experience: Experience,
+  headerDecoration: any
+}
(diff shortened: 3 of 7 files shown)


src/components/ui/CustomSelector.js
```diff
@@ -59,12 +59,19 @@ const cx = {
...
-        top: 11px !important
+        padding: 0 !important
+        display: flex !important
+        align-items: center
+        height: 100%
+      }
+
+      & .Select-input > input {
+        height: 100%
+        padding: 0px !important
```
nicksp, line 74:
```suggestion
        padding: 0 !important
```

src/components/ui/WorkExperience/WorkExperienceEntryPresenter.js
```diff
@@ -6,40 +6,35 @@ import cmz from 'cmz'
 
 import JobSkills from '../JobSkills'
 import theme from '../../../styles/theme'
-import typo from '../../../styles/typo'
 
 import type { Experience } from '../../../utils/types'
-type Props = { experience: Experience }
+type Props = {
+  experience: Experience,
+  headerDecoration: any
```
nicksp, line 13: I believe it's safer to use `*` instead of `any`. When Flow sees `any` that means you can use any type. The program is indifferent as to the parameter's type, and it will not try to infer result type. So result type will be `any` also. **Using `any` is completely unsafe, and should be avoided whenever possible.**

The asterisk `*` in Flow tells it to infer the type, rather than making you write it out explicitly.

**NOTE:** Flow treats missing type parameters as `any`, so you in your example you could even leave it out and it would be implicit.

src/components/ui/WorkExperience/index.js
```diff
@@ -13,17 +13,18 @@ import type { Experience, SkillOption } from '../../../utils/types'
 import WorkExperienceEntryPresenter from './WorkExperienceEntryPresenter'
 import WorkExperienceEntryEditor from './WorkExperienceEntryEditor'
 
-import theme from '../../../styles/theme'
 import typo from '../../../styles/typo'
 
 export type Props = {
   experiences: ?Array<Experience>,
+  initialVisibleExperiences: number,
```
nicksp, line 20: Wouldn't it be clearer in terms of intent to use a name like `initialNumberOfVisibleExperiences`? This will make it clear what the purpose is even without checking the code and type. Wdyt?

src/components/ui/Timeframe.js
```diff
@@ -92,7 +92,7 @@ type Option = {
 
 type Props = {
   startDate?: Date,
-  endDate?: Date,
+  endDate?: ?Date,
```
nicksp, line 95: Following up on the discussion we had before. I've made an extensive testing on this PR and here's what I discovered: 

image

We do not seem to set `endDate` to null when deselected. Is there still a case this is can be set to `null`? I'd like the code to currently reflect the possible behavior of the app so that we do not run into confusing parts. Please double-check and raise a follow up ticket if it's smth that can't be done in this PR.

nicksp:
@jaspax All of the requests on that PR are kinda important in a way the impact the app and being clear on decisions here is crucial/ Things like `0px` can be ignored of course and what left for the sake of completeness of the review. I spent time doing that and find it important to share anything I could find and things you can improve next time. It's a knowledge sharing process. It's then up to you to ignore any of those or discuss if needed.

src/components/ui/WorkExperience/index.js
jaspax: I believe that the `number` type annotation makes the intended usage sufficiently clear.

src/components/ui/Timeframe.js
jaspax: This is bug XP-3764 and won't be handled in this changelist. In any case, the component behavior here is correct; the server has a bug.

PR #548, opened 2020-05-27
Rework TextareaEditor component

venanciorodrigo:

Fixes https://x-team-internal.atlassian.net/browse/XP-3759

## Description

The existing TextareaEditor with text formatter is deprecated and doesn't provide the desired UX. The new TextareaEditor should handle plain text and have a character limit.

## Checklist

**Before submitting a Pull Request,** please make sure the following is done:

## Screenshots

image


src/components/ui/TextareaEditor/TextareaEditor.js
```diff
@@ -106,77 +63,33 @@ class TextareaEditor extends PureComponent<Props, State> {
...
-        </div>
+        <InputField
+          id={id}
+          name={name}
+          type='textarea'
+          placeholder={placeholder}
+          value={value}
+          linesLimit={linesLimit}
+          onChange={this.handleChangeValue}
+        />
```
bernardodiasc, line 90: You can use:

```js
const { charLimit, ...rest } = this.props
```
and then declare this way:
```suggestion
        <InputField
          {...rest}        
          type='textarea'
          onChange={this.handleChangeValue}
        />
```

That would eliminate the redundancy on the type settings and props definitions of this component.

src/components/ui/TextareaEditor/TextareaEditor.stories.js
```diff
@@ -1,76 +1,19 @@
...
-      linesLimit={5}
       placeholder='A magical placeholder.'
-      text='Magical text #1.\nMagical text #2.\nMagical text #3.\nMagical text #4.\nMagical text #5.\nMagical text #6.\nMagical text #7.\nMagical text #8.\nMagical text #9.\nMagical text #10.'
-      html='<p>Magical text #1.</p><p>Magical text #2.</p><p>Magical text #3.</p><p>Magical text #4.</p><p>Magical text #5.</p><p>Magical text #6.</p><p>Magical text #7.</p><p>Magical text #8.</p><p>Magical text #9.</p><p>Magical text #10.</p>'
-      onChange={action('Fired when textarea value changes')}
-      onFocus={action('Fired when textarea is focused')}
-      onBlur={action('Fired when textarea is blured')}
-    />
-  ))
-  .add('initial value bigger than character limit', () => (
```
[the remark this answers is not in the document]
bernardodiasc: Just to be more clear, this is an edge case that can happen, this story would expose how the component behaves in such case. Important to say, the data should never be cropped.

venanciorodrigo:
@bernardodiasc Thanks for the golden advises here in this PR, made me understand better the new standard on how we structure new components. I hope things looks better now, let me know if I'm missing something.

src/components/ui/TextareaEditor.stories.js
```diff
@@ -0,0 +1,49 @@
...
+  .add('invalid with value', () => (
+    <TextareaEditor
+      isInvalid
+      value='Textarea invalid.'
+    />
+  ))
+
+storiesOf('Core Components|Form Components/TextareaEditor/Debug', module)
+  .add('missing props', () => <TextareaEditor />)
+  .add('initial value bigger than character limit', () => (
```
bernardodiasc, line 44: This is a good edge case to have exposed. In terms of UX there's major flaw here.

When this field is loaded with value bigger than the limit, the user can't edit the data. This is kinda a bug, we need to create conditions for the user to be able to still edit data here. As said, it's an edge case, but we need to consider. As an example of real situation: imagine the field had some limit in one day, then the data is stored, then in the future this limit is changed, then later if the user tries to edit this field, it won't be allowed.

To keep it simple, my recommendation is to disable the limit prevention (keeping the counter/limit display in red) if the initial data is bigger than the limit. But once the user clean up the data down to a number below the limit, the validation becomes enabled. 

This is unlikely to happen so easily, but it's a real edge case scenario that need to be considered. :)

venanciorodrigo:
@bernardodiasc changes applied this is how things looks like now: image

Feel free to suggest anything that may be missing here.

src/components/ui/TextareaEditor.js
```diff
@@ -0,0 +1,104 @@
...
+
+  state = {
+    value: this.props.value || '',
+    initialValueLength: this.props.value ? this.props.value.length : 0
+  }
+
+  handleChangeValue = (input: Object): void => {
+    const inputValue = input.target.value
+    const charLimit = this.props.charLimit
+    const isValueLimitExceeded = this.state.initialValueLength >= inputValue.length
```
jaspax, line 67: I find this name very misleading. The actual logic here is checking whether the value length has gotten shorter, so let's use that to name the variable `isValueShorter`.

venanciorodrigo:
@jaspax Thanks for the suggestions here 👍 changes applied to the PR

src/components/ui/TextareaEditor.js
```diff
@@ -0,0 +1,104 @@
...
+    charLimit: 1000,
+    linesLimit: 0
+  }
+
+  state = {
+    value: this.props.value || '',
+    initialValueLength: this.props.value ? this.props.value.length : 0
+  }
+
+  handleChangeValue = (event: Object): void => {
```
nicksp, line 64: That is a weak typing. The more accurate option would be to use smth like `SyntheticKeyboardEvent< HTMLInputElement >`. More details: https://flow.org/en/docs/react/events/

It can also be found throughout the app as well.

src/components/ui/TextareaEditor.js
```diff
@@ -0,0 +1,104 @@
...
+    const inputValue = event.target.value
+    const charLimit = this.props.charLimit
+    const isValueShorter = this.state.initialValueLength >= inputValue.length
```
nicksp, line 67: nitpicking but it's unclear _what_ it's shorter than?

src/components/ui/TextareaEditor.js
```diff
@@ -0,0 +1,104 @@
...
+        <InputField
+          {...rest}
+          value={value}
+          type='textarea'
+          onChange={this.handleChangeValue}
+        />
+        <div className={cx.textCountStyles}>
+          <span className={value.length >= charLimit ? cx.textLimitExceeded : ''}>
+            {value.length}/{charLimit}
+          </span>
```
nicksp, line 97: Just a note, not a change request. This would better be off done as a separate component so it can be placed declaratively where we want. It's in one position today, but use cases can be different to be able to render that component in other places. It just adds more flexibility.

src/components/ui/TextareaEditor.js
venanciorodrigo: Sorry I ran out of creativity in this one from `isValueLimitExceeded` to `isValueShorter` perhaps `isValueShorterThanInitialValue`. Any other suggestion are welcome @nicksp @jaspax

nicksp: @venanciorodrigo If i got it right, what you want here is simply saying that we're eligible to perform updates on the value provided, that this value still fits the limit. If so, then smth like that might look good:

```js
const inputValueLength = inputValue.length
const  =  < this.state.initialValueLength
const hasAllowedLength = inputValue.length <= charLimit

<...>

if (hasAllowedLength || hasAllowedInitialLength) {
....
}

```

But to be honest, I do not fully understand the purpose of this condition: `isValueValidLength || isValueShorter` What is the use case? Shouldn't it be like we check whether the current value has valid length (doesn't exceed the limit) OR we check for the initial value if **only** no new text has been provided? 🤔

venanciorodrigo: The first condition `hasAllowedValueLength` consist to check if the new value inserted by user doesn't overlap the char limit for that field. In this case, we update the value state. 

The second condition `hasAllowedInitialValueLength` consist to check if the initial value after the component render is bigger than the limit provided to the field. 
eg: image
In this case, the user will be able to update the input value so we set the state here `this.setState({ value: inputValue }...` but when the user hit the char limit from the field (20 according to the example above) we will set the initialValue state as the field limit (20). The goal here is to give the possibility to the user edit the field even if the initial value is bigger than the limit set on it. Also, we don't allow the user to write more than 20 characters once he updates the field to be less or equal than 20. 
eg: image I hope this a bit clearer now.

nicksp: It is, thanks.

PR #550, opened 2020-06-02
Xp 3401 admin work experience

jaspax:

Fixes https://x-team-internal.atlassian.net/browse/XP-3401

## Description

Feature branch for epic XP-3401.

## Checklist


src/components/ui/SkillsSelector.stories.js
```diff
@@ -55,5 +56,6 @@ storiesOf('Core Components|Form Components/SkillsSelector/Debug', module)
     </Body>
   ))
   .add('missing props', () => (
+    // $FlowFixMe
```
bernardodiasc, line 59: This is not ideal fix. UI components are not regular JS code. By allowing to declare a UI component without any prop we expose can expose critical problems that happens with undefined data. It's important that UI components doesn't cause WSOD just because the server crashed or some data source the feeds these props gets corrupted somehow. Render appropriately when no data is provided is one of the points to have this debug story.

jaspax: @bernardodiasc , I agree that we should have tests for when props are missing, but Flow does not allow you to pass tests on such a component (unless you declare all props as optional, which is even worse). Using `$FlowFixMe` inside the stories files is the only way to allow Flow checks to pass and still test the missing-props case.

bernardodiasc: > unless you declare all props as optional, which is even worse

Technically all UI props are optional, that's an assumption that allow us to address the render for invalid data passed on props. :)

jaspax: All UI props are optional from React's POV. But Flow considers declared prop types to be mandatory unless marked otherwise, and it's actually *very convenient* to be able to get Flow warnings if you miss a prop.

jaspax: Anyway, the task is already closed. If you want to talk more we should take it to Slack.

PR #551, opened 2020-06-02
Updates for components used at About page  

bernardodiasc:

Fixes https://x-team-internal.atlassian.net/browse/XP-3679

## Description

Added some updates for components used at About page.

## Checklist

## Related PRs

branch | PR
------ | ------
`XP-3679-create-about-page` | [XP API](https://github.com/x-team/xp-api/pull/833)
`XP-3679-create-about-page` | [XP Registration](https://github.com/x-team/xp-registration/pull/207)

### CustomSelector

Added invalid state.

http://localhost:9001/?path=/story/core-components-form-components-customselector-states--invalid

Screen Shot 2020-06-02 at 14 14 05

### SkillsSelector

Added invalid state.

http://localhost:9001/?path=/story/core-components-form-components-skillsselector-states--invalid

Screen Shot 2020-06-02 at 14 15 07

### TextareaEditor

Removed bold on counter text.

http://localhost:9001/?path=/story/core-components-form-components-textareaeditor--basic-usage

Screen Shot 2020-06-02 at 14 15 41

### Timeframe

Added invalid and disabled state.

http://localhost:9001/?path=/story/core-components-form-components-timeframe-states--invalid

Screen Shot 2020-06-02 at 14 16 35
Screen Shot 2020-06-02 at 14 16 12


src/components/ui/CustomSelector.js
```diff
@@ -101,15 +101,28 @@ const cx = {
...
+  ),
+
+  invalid: cmz(`
+    & .Select-control {
+      border: 1px solid ${theme.lineRed}
+    }
+
+    &.is-focused .Select-control {
+      border: 1px solid ${theme.lineRed} !important
+      box-shadow: 0 0 3px ${theme.lineRed} !important
```
jaspax, line 113: Didn't we agree NOT to use the red for focus any longer? Or am I misremembering?

bernardodiasc: In this case in the `invalid` state, which is expected to be the only case in red borders.

src/components/ui/TextareaEditor.js
```diff
@@ -80,6 +82,12 @@ class TextareaEditor extends PureComponent<Props, State> {
     }
   }
 
+  componentWillReceiveProps ({ value: nextValue }: Props) {
```
venanciorodrigo, line 85: I'm wondering if we should use this hook since from version 17 the support for this will change not sure also if we should care about it and use `componentDidUpdate` instead for instance.

nicksp: I agree we should not use unsafe and about to deprecate method: https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops Better doing it sooner rather than later so it's more future-proof. By the link above are several recommendations of how to substitute that depending on the use case. cc @bernardodiasc

bernardodiasc: I made changes in the XP-Reg to deal with this problem and worked fine: https://github.com/x-team/xp-registration/pull/207/commits/88daa73302d4cdb1e69c11950ec80d9e3df9a414

But I tried the alternatives suggested in the link for this component and didn't worked as expected. Is this repo ready to make use of functional components and hooks (for such things like useState)? That would be a good solution.

bernardodiasc:
Alright, one more. 3 approvals means ready for merge, right? 😄

